PR #273, opened 2025-12-23
Use $LISTENARR_LOG_LEVEL to override default log level

lhanson:
When debugging potential issues, it would be useful to be able to override the default log level at runtime. This PR looks for the `LISTENARR_LOG_LEVEL` environment variable to override the default of `Information`.


therobbiedavis:
Closing this as it was cherrypicked into #284. Thank you!

PR #340, opened 2026-02-18
Fix download import pipeline for single-file torrents

zombopanda:
## Summary

Fixes multiple issues causing single-file torrent downloads (e.g. from MyAnonamouse) to fail to import or to catastrophically import the entire downloads directory instead of the target file.

- **DownloadMonitorService cookie auth**: The factory `HttpClient` "DownloadClient" has `UseCookies=false`, dropping the qBittorrent SID cookie. Now creates a local `HttpClient` with its own `CookieContainer`.
- **Single-file torrent path resolution** (`QbittorrentAdapter.GetImportItemAsync`): For single-file torrents the subfolder was empty, so `outputPath` resolved to `save_path` (the entire downloads directory). Now uses `Path.Combine(savePath, fileName)`.
- **Race condition fix** (`DownloadMonitorService`): Was setting status to `Completed` before the stability window expired, allowing `DownloadProcessingBackgroundService` to pick up the download too early. Now only updates progress without changing status until finalization.
- **Path fallback fix** (`DownloadProcessingBackgroundService`): Was using `dl.FinalPath ?? dl.DownloadPath` (= `save_path`). Now prefers `ClientContentPath` from metadata (`content_path` = actual file).
- **Purge safety check** (`DownloadService`): Skips orphan purge when the client returns 0 queue items but active downloads exist, preventing deletion during transient connectivity issues.
- **EF Core version wildcards**: Uses `8.*` for all EF Core packages to avoid NU1605 version mismatch warnings.

## Test plan

- [x] Built and deployed to homelab Docker container
- [x] Triggered download of single-file M4B torrent from MyAnonamouse via Listenarr
- [x] Verified DownloadMonitorService successfully detects completion (cookie fix working)
- [x] Verified stability window prevents premature processing (race condition fix working)
- [x] Verified correct file path resolution (single-file fix: `/data/downloads/file.m4b` not `/data/downloads/`)
- [x] Verified successful import to audiobook library (only 1 file moved, not entire directory)
- [x] Verified purge safety check prevents accidental deletion

### DIFF
--- a/listenarr.api/Listenarr.Api.csproj
+++ b/listenarr.api/Listenarr.Api.csproj
@@ -17,19 +17,19 @@
   
   <ItemGroup>
     <PackageReference Include="AsyncKeyedLock" Version="8.0.0" />
-    <PackageReference Include="Microsoft.Data.Sqlite.Core" Version="8.0.23" />
-    <PackageReference Include="Microsoft.EntityFrameworkCore" Version="8.0.23" />
+    <PackageReference Include="Microsoft.Data.Sqlite.Core" Version="8.*" />
+    <PackageReference Include="Microsoft.EntityFrameworkCore" Version="8.*" />
     <PackageReference Include="SixLabors.ImageSharp" Version="3.1.12" />
     <PackageReference Include="Microsoft.Extensions.Http.Polly" Version="8.0.11" />
     <PackageReference Include="Polly" Version="8.5.0" />
     <PackageReference Include="SharpCompress" Version="0.36.0" />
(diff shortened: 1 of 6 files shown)


therobbiedavis:
Hey @zombopanda, thanks for the PR. I just released a new version that fixed imports so I closed this, but I will review this PR further to see if there any improvements I can make on my implementation and give you credit.

Thank you!

zombopanda:
@therobbiedavis hi, the issue is still present in the latest release, I have recreated the PR with the latest changes https://github.com/Listenarrs/Listenarr/pull/350

PR #389, opened 2026-03-01
Minor dict-based and semaphore-based optimizations

MarkCiliaVincenti:


therobbiedavis:
Hey @MarkCiliaVincenti Thanks for another PR! I am not a C# expert, obviously, hence this project. Can you help me understand some of the differences here and what we gain from these optimizations?

MarkCiliaVincenti:
There were a bunch of changes. Mainly avoiding double dictionary lookups. There were a number of places where we were checking if a key exists and if it does then get its value. That's doing double the work.

Also the use of AsyncKeyedLock which you're already depending on ensures that semaphores are both pooled as well as the dictionary cleaned up after use to avoid slow memory leaks.

Is there anything in particular that you don't understand well here?

therobbiedavis:
Hey @MarkCiliaVincenti, I'm testing this out now but I wanted to keep you informed that I had to resolve a merge conflict when bringing it up to date with 0.2.56. Also could you provide a PR description so if regression testing goes well I can just approve this PR? Thanks!

MC Context:
In DownloadService, the conflict was around the fallback lookup when removing a download. Your optimization changed the code to use TryGetValue for TorrentHash, which is fine mechanically, but current behavior in 0.2.56 also supports matching by ClientDownloadId and does it case-insensitively. I resolved that by keeping the broader existing behavior, but still using the TryGetValue pattern instead of the older dictionary access style.

In DownloadMonitorService, your changes were mostly the dictionary/collection cleanup pattern: ContainsKey to TryGetValue, and ContainsKey plus Remove to just Remove. We kept that approach. The conflict there overlapped with later qBittorrent monitoring changes, so I kept the richer current implementation that tracks ratio/seed-limit state and CanMoveFiles / CanBeRemoved, because later logic already depends on that behavior.

MarkCiliaVincenti:
Do you still need this PR?

PR #398, opened 2026-03-06
feat: Library Import page (Sonarr-style)

### LINKED ISSUE #182: Scanning for existing audiobooks
Does Listenarr scan the root folder for existing audiobooks?

tsolo4ever:
## Summary

This PR adds the much requested feature, Library Import. It also fixes multipart import behavior, and adds safer filesystem cleanup options when deleting audiobooks.

## Added

- Full-page **Library Import** view at `/library-import`
- Sidebar navigation link for **Library Import**
- Root folder selector with explicit **Scan** action for unmatched files
- Cached unmatched scan results with last-scanned timestamp and unmatched count
- Grouped multipart-file rows in Library Import instead of one row per file
- Expandable grouped-file list so users can inspect all files included in a row
- Destination root folder selector in the Library Import footer
- Live import-progress indicator on the **Import N Books** button
- Sortable and resizable Library Import table columns
- Mobile-optimized Library Import table/card layout
- Embedded metadata parsing for unmatched scan, including better ASIN discovery from more tag variants
- Support for non-standard embedded ASIN tags such as `ASIN:` and `CDEK`
- ASIN embedding on import
- User-configurable import extension/filetype blacklist in Settings
- Filesystem delete options for audiobooks, including folder deletion and empty author-folder cleanup when safe
- Regression test coverage for Library Import, unmatched scan grouping, multipart import behavior, delete behavior, and footer/import UX
- Library Import matching prefers:
  - `detectedAsin`
  - detected album/title plus author
  - filename/folder fallback
- Library Import search modal pre-fills and uses detected author more intelligently
- The Book column prefers detected album/title metadata and falls back to folder name only when metadata is unavailable
- Import actions treat grouped files as a single import unit
- Move/Hardlink-Copy imports bring companion files along unless blocked by the configured blacklist

## Fixed
- Multipart imports assigning colliding filenames and ending up out of order after rename
- Multi-file manual/library/download imports being associated to the wrong audiobook path or refused as `File not associated`
- Forewords/front-matter being split away from the rest of the book when album/title/author metadata indicates they belong together
- Audiobook delete flows that did not fully remove files/folders as expected

## Test plan

- [x] Navigate to `/library-import` from the sidebar
- [x] Select a root folder and load cached unmatched results
- [x] Run a new scan and verify unmatched items populate
- [x] Verify grouped multipart books appear as a single row
- [x] Expand grouped files and confirm all included source files are visible
- [x] Start auto-matching and verify rows search one at a time
- [x] Verify ASIN-tagged files prefer exact ASIN matches
- [x] Verify detected album/title + author are used before folder fallback
- [x] Verify mixed folders split into the correct books
- [x] Import selected rows and confirm grouped files are imported together and in order
- [x] Verify companion files are imported and source-folder cleanup works for move mode
- [x] Verify blacklist settings skip configured file extensions
- [x] Verify audiobook delete filesystem options remove files/folders safely
- [x] Frontend unit tests, type-check, build, and backend test coverage updated for the new flows

Closes #182


fe/src/stores/libraryImport.ts
```diff
@@ -0,0 +1,527 @@
...
+    subtitle: result.subtitle,
+    series,
+    seriesNumber,
+    seriesAsin,
+    description: result.description,
+    publisher: result.publisher,
+    language: result.language,
+    runtime: result.runtime ?? (result.lengthMinutes ? result.lengthMinutes * 60 : undefined),
+    imageUrl: result.imageUrl,
+    genres: result.genres,
```
tsolo4ever: sorry is there any other duplicates? its a lot of code to go through

listenarr.api/Services/UnmatchedScanQueueService.cs
```diff
@@ -0,0 +1,102 @@
...
+        Task<Guid> EnqueueAsync(string rootFolderPath);
+        bool TryGetJob(Guid id, out UnmatchedScanJob? job);
+        void UpdateJob(Guid id, string status, List<UnmatchedFileResult>? results = null, string? error = null);
+        bool TryGetLastJobForPath(string rootFolderPath, out UnmatchedScanJob? job);
+        ChannelReader<UnmatchedScanJob> Reader { get; }
+    }
+
+    public class UnmatchedScanQueueService : IUnmatchedScanQueueService
+    {
+        private readonly ConcurrentDictionary<Guid, UnmatchedScanJob> _jobs = new();
```
therobbiedavis, line 50: Can we add TTL or cleanup for completed jobs? Right now every scan result is kept in-memory forever, which makes repeated large scans an unbounded memory sink and keeps stale path data retrievable indefinitely.

fe/src/stores/libraryImport.ts
therobbiedavis: The search results now serialize genres as objects while /library/add expects metadata.genres to be string[]. We already sanitize searchResult.genres below, but not metadata.genres, so library import sends an invalid payload and model binding fails. This is also handled in the AddLibraryModal.vue with mapAudimetaToAudible() which flattens the genres, but we should probably create a reusable helper for this.

fe/src/views/library/LibraryImportView.vue
therobbiedavis, line 1: The UX needs some work here.

- It's not really usable on mobile.
- I would also break out the search into a modal instead of inline. Inline does look nice, but it creates limitations.
- Unify the UI sizing: fields and buttons are various heights
- Follow the existing UX for buttons: Non-destructive action buttons should be blue. Destructive action buttons should be red. Non-destructive non-action buttons should be grey. (At least this is how I'm trying to keep it)

listenarr.api/Services/UnmatchedScanBackgroundService.cs
```diff
@@ -0,0 +1,268 @@
...
+                        {
+                            if (AudioExtensions.Contains(Path.GetExtension(file), StringComparer.OrdinalIgnoreCase))
+                                candidates.Add(file);
+                        }
+                        catch (Exception ex) when (ex is not OperationCanceledException && ex is not OutOfMemoryException && ex is not StackOverflowException)
+                        {
+                            _logger.LogDebug(ex, "Skipped file {File} during unmatched scan", file);
+                        }
+                    }
+                    foreach (var sub in Directory.EnumerateDirectories(normalizedDir))
```
therobbiedavis, line 404: We should probably reject reparse points here, or at least verify the resolved child path still stays under the configured root before recursing. It would be safer from a security perspective because otherwise it could allow the scan to walk into unrelated host paths.

docker-compose.yml
therobbiedavis, line 1: Why is this needed?

listenarr.api/Controllers/RootFoldersController.cs
```diff
@@ -121,5 +128,76 @@ public async Task<IActionResult> Delete(int id, [FromQuery] int? reassignTo = nu
...
+            if (folder == null) return NotFound(new { message = "Root folder not found" });
+
+            var jobId = await _unmatchedQueue.EnqueueAsync(folder.Path);
+            return Ok(new { jobId = jobId.ToString() });
+        }
+
+        /// <summary>
+        /// Returns the status and results of a previously enqueued unmatched scan job.
+        /// </summary>
+        [HttpGet("unmatched-results/{jobId}")]
```
therobbiedavis, line 149: Just a note: this is fine for the current self-hosted single-admin model, but the job lookup is effectively global because completion is broadcast to all clients and results are keyed only by jobId. I want to move this over tosScoping jobs/events to the caller as that is more secure, and honestly just the right way to do it. This also happens in the ScanQueueService, ScanBackgroundService, MoveQueueService, and the DownloadProcessingJob, so there's a lot to change there, which is why I haven't yet.

therobbiedavis (changes requested):
Hey there, thanks for taking up the mantle on this massive undertaking! I've reviewed this and I think it's a great start and about 85% there. My concerns surround primarily the UX, especially for mobile, and the result.genres not getting flattened causing the import to fail. The others where I give actionables need to be addressed too, but until those primary concerns are addressed this isn't functional.

listenarr.api/Controllers/RootFoldersController.cs
tsolo4ever: msedge_VN6mM1Uyff
3Oye2t6ctw

therobbiedavis:
@tsolo4ever Sorry just noticed this behavior inconsistency, maybe you can help explain. My test audiobook "The Hobbit" is part of the "Lord of the Rings" series.
image
But when I search, it searches "Lord of the Rings" (the series) instead of "The Hobbit" the title.
image

tsolo4ever:
oops almost all of mine use the ASIN didnt really thing to test that its good now sorry

therobbiedavis:
> oops almost all of mine use the ASIN didnt really thing to test that its good now sorry

How are you searching via existing ASIN from the file?

therobbiedavis:
@tsolo4ever Mind if I make some updates to the UI/UX?

tsolo4ever:
feel free to change whatever you need this is a group effort right? i know i am not the best at this i just have a 2175 audiobook library... lol

as for the ASIN i use mp3tag and tag everything as i get it i have been using libation to pull my books from Audible then feeding them through everything to get tagged and renamed

therobbiedavis:
Hey @tsolo4ever I haven't forgotten about this, I've been working on getting a documentation website up. https://listenarrs.github.io/ 

Still not 100% with it, but I will review and tweak anything needed for this PR this week

therobbiedavis:
@tsolo4ever I had to make some updates here.
- Updated the scan to look for ASINs in the file metadata and use that for matching, otherwise fallback to title/author search.
- Updated the API to better handle when author searches to audimeta fail and fallback to Audible page scraping.
- Updated the mobile UX a bit so it should be a bit more user-friendly on smaller resolutions.
- Updated the table to allow for sorting and resizing.
- Added a `Path` to the table for user visibility

therobbiedavis:
@codex review

therobbiedavis (approved):
lgtm

tsolo4ever:
sorry i have been busy i will check this over in the morning i do have a question tho do you plan on using the calendar?

therobbiedavis:
I went ahead and merged it, but please do test it out. We can always iterate over it and added fixes. I have no doubt we'll need to. Overall it was there, just needed to test against some use cases and fix it up accordingly.

The calendar is already hooked up and should be functional. It uses the published date from Audimeta. If that date is in the future then it shows as upcoming on the calendar.

PR #465, opened 2026-04-01
Add Slskd support: Both for indexer and download client

### LINKED ISSUE #183: Requesting support for soulseek/slskd
Would be great to add soulseek/slskd as an indexer option

T4g1:
### Summary
Fix #183: Adds Slskd (https://github.com/slskd/slskd) support both for:
* Indexer: As a search provider
* Download client

I also tried to streamline the process of managing those with the objective of making it easier to add indexers and/or download clients

This PR is way too big, sorry about that :x

### Added
* Proper download client selection based on `IDownloadClientAdapter` and supported protocols to fetch list of `DownloadClientConfiguration` types to look for
* Slskd indexer support using the native API
* Slskd download client support using the native API
* Basic mock class and basic test class to reduce setups required to run those

### Changed
* Indexer model updated to support enumeration and enhanced attribute name
* Improved Protocol/Implementation attributes by adding enumeration: Less typo/error prone than having string constant all over the place
* Factorizing some methods into specific classes where they belonged to reduce backend complexity and improve separation of concern
* Improved search provider and download client adapter handling to make it easier to add or update them

### Test coverage
* Tests Add, Remove and Fetch from Slskd adapter
* Test basic Search from Slskd search provider

### Risks
* Model changed through migration: Having access to few data localy, it may be required to double check the migration code to make sure it works for every cases
* Backend for search/download and polling is modified to support third-party that associates one download with multiple transfers (previous data structure assumed one download was at most one transfer)

### Next steps
* It may be useful to make front end for the indexers and download clients form rely more on configuration rather than complex logic that may be inadvertly broken when adding new ones
* Properly supports download client selection based on priority settings: Right now, only the first one is considered, each of them should be tried until one succeeds by order of priority
* `DownloadConfigurationClient.Type` should probably be an enumeration or at least restricted by the `DownloadClientAdapter` configured
* There is a lot of code duplication in classes that make use of `IDownloadOrchestrator` that could be removed
* `IDownloadOrchestrator` and other interfaces in general should not expose or expect models specifity like technical object ID but rather pass objects except for API interfaces
* Direct download should be implemented by an adapter and not hardcoded as a special case
* TorrentHash and ClientDownloadId should be reconciliated, I believe removing the TorrentHash logic to make the ClientDownloadId generic instead is the way to go (it simplifies the downloads processing imho)
* It could make sense to create a base class for interfaces (defining user/browser agents, http methods helpers, ...) to use for Adapters and Providers


fe/src/components/domain/download/DownloadClientFormModal.vue
```diff
@@ -401,35 +400,23 @@ const normalizeHost = (value: string): string => {
 }
 
 const isUsenet = computed(() => {
-  return formData.value.type === 'sabnzbd' || formData.value.type === 'nzbget'
+  return DownloadProtocol.Usenet in DownloadClientConfigs[formData.value.type].protocols
```
therobbiedavis: This will always return false because the array checks for index keys (0, 1...), not values so "Usenet" in ["Usenet"] is false. The new form behaviour for SABnzbd and NZBGet is now broken.

therobbiedavis: You could do `return DownloadClientConfigs[formData.value.type].protocols.includes(DownloadProtocol.Usenet)`

listenarr.api/Services/Adapters/SlskdAdapter.cs
```diff
@@ -0,0 +1,313 @@
...
+        {
+            using var db = _dbFactory.CreateDbContext();
+            var transfers = await GetQueueAsync(client, ct);
+            downloads.ForEach(async download =>
+            {
+                var transfer = transfers.FirstOrDefault(transfer => transfer.Id == download.Id);
+                if (transfer == null)
+                {
+                    // Download no longer in the download client
+                    await service.HandleFailedDownloadAsync(download, client, db, settings, $"Slskd does not have the download {download.Id}", ct);
```
therobbiedavis, line 282: These three awaits are fire-and-forget, meaning the ForEach treats the async as void, the method returns downloads immediately without and poll work and any exceptions will crash the process silently.

listenarr.api/Services/Adapters/SlskdAdapter.cs
```diff
@@ -0,0 +1,313 @@
...
+                    return;
+                }
+
+                await service.UpdateDownloadProgressAsync(download, transfer.Progress, transfer.Size - transfer.Downloaded, "downloading", db, ct);
```
therobbiedavis, line 286: 2/3 fire-and-forget awaits

listenarr.api/Services/Adapters/SlskdAdapter.cs
```diff
@@ -0,0 +1,313 @@
...
+
+                if (transfer.Status == "completed")
+                {
+                    // Download completed
+                    await service.FinalizeDownloadAsync(download, download.DownloadPath, client, ct);
```
therobbiedavis: 3/3 fire-and-forget awaits.

listenarr.api/Services/DownloadMonitorService.cs
```diff
@@ -727,21 +730,33 @@ private async Task PollDownloadClientsAsync(
...
-                            await PollSABnzbdAsync(client, clientGroup.ToList(), dbContext, appSettings, cancellationToken);
-                            break;
-                        case "nzbget":
-                            await PollNZBGetAsync(client, clientGroup.ToList(), dbContext, appSettings, cancellationToken);
-                            break;
+                    using (var scope = _serviceScopeFactory.CreateScope())
+                    {
+                        var downloadClientAdapterFactory = scope.ServiceProvider.GetRequiredService<IDownloadClientAdapterFactory>();
+                        var adapter = downloadClientAdapterFactory.GetByType(client.Type);
+                        if(adapter == null)
```
therobbiedavis: GetByType throws so never returns null

listenarr.api/Services/DownloadMonitorService.cs
```diff
@@ -727,21 +730,33 @@ private async Task PollDownloadClientsAsync(
...
+                        {
+                            _logger.LogCritical($"No download client adapter exists for: {client.Type}");
+                            continue;
```
therobbiedavis, line 743: which means this is never reached

listenarr.api/Services/DownloadMonitorService.cs
```diff
@@ -727,21 +730,33 @@ private async Task PollDownloadClientsAsync(
...
+                        }
+                        
+                        var updatedDownloads = await adapter.PollAsync(this, client, clientGroup.ToList(), appSettings, cancellationToken);
+
+                        // Persists download updates
+                        updatedDownloads.ForEach(async download =>
```
therobbiedavis: fire-and-forget again, signalr updates are not sent.

fe/src/components/domain/search/ManualSearchModal.vue
```diff
@@ -341,6 +342,9 @@ const anyHasPeers = computed(() =>
 const anyHasLanguage = computed(() =>
   displayResults.value.some((r) => !!normalizeLanguage(r.language)),
 )
+const anyHasGrabs = computed(() =>
+  displayResults.value.some((r) => r.grabs !== undefined && r.grabs !== null),
```
therobbiedavis, line 346: This checks for undefined but grabs is always defined, defaults to 0.

fe/src/components/domain/search/ManualSearchModal.vue
```diff
@@ -179,7 +179,7 @@
                       <PhArrowDown /> {{ result.leechers }}
                     </span>
                   </td>
-                  <td class="col-grabs">
+                  <td v-if="anyHasGrabs" class="col-grabs">
```
therobbiedavis, line 182: this will always be defined as it currently is now. `grabs` defaults to 0, not null or undefined.

fe/src/components/settings/IndexerFormModal.vue
```diff
@@ -390,24 +397,21 @@ watch(
...
+    switch(implementation) {
+      case 'InternetArchive':
+      case 'Newznab':
+        formData.value.protocol = DownloadProtocol.Usenet
+        break;
+      case 'Slskd':
+        formData.value.protocol = DownloadProtocol.Soulseek
+        break;
+      case 'MyAnonamouse':
+        formData.value.url = 'https://www.myanonamouse.net'
```
therobbiedavis, line 409: missing break

listenarr.api/Services/Search/Providers/SlskdSearchProvider.cs
```diff
@@ -0,0 +1,193 @@
...
+    {
+        _httpClient = httpClient;
+        _logger = logger;
+    }
+
+    private async Task<HttpContent?> Call(Indexer indexer, HttpMethod method, String endpoint, Object? content = null)
+    {
+        _logger.LogDebug("Call on {endPoint}", endpoint);
+        var apiRequest = new HttpRequestMessage(method, endpoint);
+        apiRequest.Headers.Add("X-Api-Key", indexer.ApiKey);
```
therobbiedavis: This will always pass an api-key, we should add a check or null-guard.

listenarr.api/Services/Search/Providers/SlskdSearchProvider.cs
```diff
@@ -0,0 +1,193 @@
...
+            {
+                return [];
+            }
+            
+            // Wait for the search to be completed
+            var cts = new CancellationTokenSource(TimeSpan.FromSeconds(30));
+            bool searchCompleted = false;
+            do
+            {
+                await Task.Delay(5000, cts.Token);
```
therobbiedavis: Why the hardcoded 5sec? Is there a way to remove this delay, or change it to use the delay as needed?

listenarr.infrastructure/Persistence/Converters/DownloadProtocolConverter.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        public static DownloadProtocol ConvertStringToDownloadProtocol(string value)
+        {
+            if (Enum.TryParse<DownloadProtocol>(value, out DownloadProtocol result))
+            {
+                return result;
+            }
+
+            // Fallback value
+            return DownloadProtocol.Torrent;
```
therobbiedavis, line 23: Existing rows with Type = "Custom" or "DDL" become Torrent after migration with no warning. Enum.TryParse is also case-sensitive by default, so we should probably normalize the comparsion value for legacy handling and future proofing.

therobbiedavis (changes requested):
Looks great! There's some small logic issues we need to handle and I haven't done a real-world test of the implementation yet, but overall it looks good so far.

listenarr.api/Services/Search/Providers/SlskdSearchProvider.cs
T4g1: The search is asynchronous, Slskd receive the query and process it internaly, once status is completed (or something like that), we can retrieve the results
Afaik, there is no way to avoid that pool loop. On my setup, searches takes around 5-10 seconds to complete, hence the hardcoded 5 seconds polling and hard 30 seconds limits. We could add a way to customize it though. wdyt ?

Here is the API fetched directly from my Slskd instance: 
[slskd-api.json](https://github.com/user-attachments/files/26685820/slskd-api.json)
It's from Slskd version 0.24.5.0 (0.24.5.0+9c1c0203) (not v0 as the JSON says, nobody bother updating API versions for some reason...)

fe/src/components/domain/search/ManualSearchModal.vue
T4g1: I changed that behavior so Grab can be null now (see [SearchResult.cs](https://github.com/Listenarrs/Listenarr/pull/465/changes#diff-2bc81c590a809c7dc8d6f11a936400219768347e91c60c33aef861c95a578bd8L50) which is now in [SearchResult/SearchResult.cs](https://github.com/Listenarrs/Listenarr/pull/465/changes#diff-2bc81c590a809c7dc8d6f11a936400219768347e91c60c33aef861c95a578bd8R27)

Here in action (Grab column is not present): 
2026-04-13 21_05_25-Listenarr - Audiobook Management

fe/src/components/settings/IndexerFormModal.vue
T4g1: It's intended as switch fallthrough so protocol gets set to Torrent too but the URL is relevant only for MyAnonamouse, I discovered that it is disabled in C# though so I can remove it to stay consistent if you want (it seems to be considered bad nowadays but i find that elegant)

listenarr.infrastructure/Persistence/Converters/DownloadProtocolConverter.cs
T4g1: DDL case should be handled by the [migration](https://github.com/Listenarrs/Listenarr/pull/465/changes#diff-ffc86d38f5f9cfa2a903c8bcd4414c1f0709b9b39351683287f3d834aa7d7291R24) but what about the "Custom" case ? Is the type (protocol) "Custom" relevant ? Nothing handles that right ? Are you sure it's not implementation ?

We can do the comparison using normalized values but I'm not sure it's really required here, the only other part that uses this is the prowlarr indexers bit and we should probably sanitize the protocol we receive directly there if needed instead of adding interface specific knowledge in this converter (either in `IndexersController.cs` for the direct indexer fetch or `ProwlarrCompatController.cs` for the actual WIP of the API/interface)

fe/src/components/domain/download/DownloadClientFormModal.vue
T4g1: Good to know, I assumed the behavior was the same as Python regarding this, that's rather nasty

listenarr.infrastructure/Persistence/Converters/DownloadProtocolConverter.cs
therobbiedavis: I want to remove "Custom" from the codebase as that was an ambitious naive idea lol when I first started out.

fe/src/components/settings/IndexerFormModal.vue
therobbiedavis: Oh! Got it, clever. Hmm, would you mind making it explicit so it helps the readability and maintainability?

listenarr.infrastructure/Persistence/Converters/DownloadProtocolConverter.cs
T4g1: Well in that case, and assuming the legacy indexers with "custom" are not working. Any instalation with them will have no search available when using them. Silently changing the protocol from custom to torrent has no consequence then:
- Either it wont work, but they are already not working so not really an issue ?
- Either it will work (user probably selected somehow custom for a torrent indexer) and it will now work (very happy case :p)

T4g1:
@therobbiedavis I think this is ready for the second round of review 💯

listenarr.domain/Models/Indexer.cs
```diff
@@ -33,19 +41,43 @@ public class Indexer
...
+        /// Implementation type (e.g., "Newznab", "Torznab", ...)
         /// </summary>
-        public string Implementation { get; set; } = string.Empty;
+        public Implementation Implementation { get; set; }
 
         /// <summary>
         /// Base URL for the indexer API
         /// </summary>
-        public string Url { get; set; } = string.Empty;
+        public String _url = string.Empty;
```
therobbiedavis: I think this should be private.

fe/src/components/domain/search/ManualSearchModal.vue
therobbiedavis: wait, this is confusing because while grabs is null, line 496 will set null as 0. So im not really sure how this is working...

listenarr.api/Services/Adapters/SlskdAdapter.cs
therobbiedavis: @T4g1 i just noticed here that downloadpath may not actually be where slskd stores the files.. User could be using docker. We need to get the path from Slskd (Maybe`OutputPath` from `GetItemsAsync` and compare and translate it to the users remote mapping directory.

listenarr.api/Services/Adapters/SlskdAdapter.cs
```diff
@@ -0,0 +1,313 @@
...
+                }
+            }
+
+            return downloads;
+        }
+
+        public Task<List<(string Id, string Name)>> GetRecentHistoryAsync(DownloadClientConfiguration client, int limit = 100, CancellationToken ct = default)
```
therobbiedavis, line 299: The others below have a comment stating their not used by the backend, is this the same?

fe/src/components/domain/search/ManualSearchModal.vue
```diff
@@ -490,11 +494,11 @@ async function search() {
             seeders: typeof dto.seeders === 'string' ? Number(dto.seeders) || 0 : typeof dto.seeders === 'number' ? dto.seeders : undefined,
             leechers: typeof dto.leechers === 'string' ? Number(dto.leechers) || 0 : typeof dto.leechers === 'number' ? dto.leechers : undefined,
             grabs: typeof dto.grabs === 'string' ? Number(dto.grabs) || 0 : typeof dto.grabs === 'number' ? dto.grabs : 0,
```
therobbiedavis, line 496: Maybe we just update this to:
```grabs: typeof dto.grabs === 'string' ? Number(dto.grabs) || 0 : typeof dto.grabs === 'number' ? dto.grabs : undefined```

File: listenarr.api/Services/Search/Providers/SlskdSearchProvider.cs (hunk 12 above)
Comment (therobbiedavis) [reply listenarr.api/Services/Search/Providers/SlskdSearchProvider.cs]: I played with this a bit and yeah, this implementation seems to be the only method that works. Let's lower the delay to 1000, and decrease the cts to 20 since the default timeout for slskd is 15, we only need to be slightly more than theirs. Eventually we should make that configurable.

File: listenarr.api/Services/Adapters/SlskdAdapter.cs (hunk 4 above)
Comment (T4g1) [reply listenarr.api/Services/Adapters/SlskdAdapter.cs]: Good point, i'll investigate that

File: fe/src/components/domain/search/ManualSearchModal.vue (hunk 9 above)
Comment (T4g1) [reply fe/src/components/domain/search/ManualSearchModal.vue:182]: That's the legacy normalization though, the "if" from line 485 prevent it to be used when there is no guid on the received object (which there isn't because Slskd results are already using IndexerSearchResult format)

(but that whole legacy search result things is indeed confusing :p )

File: listenarr.api/Services/Adapters/SlskdAdapter.cs (hunk 15 above)
Comment (T4g1) [reply listenarr.api/Services/Adapters/SlskdAdapter.cs:299]: Kind of, for that one i was not even able to understand what it's supposed to do or be replaced by, I dont believe it is used anywhere exept in tests

File: fe/src/components/domain/search/ManualSearchModal.vue (hunk 16 above)
Comment (T4g1) [reply fe/src/components/domain/search/ManualSearchModal.vue:496]: We can but I believe you missed the fact that this is already legacy code and new search indexers should probably not trigger that part (cf my other comment)

File: listenarr.infrastructure/Persistence/Converters/DownloadProtocolConverter.cs (hunk 13 above)
Comment (therobbiedavis) [reply listenarr.infrastructure/Persistence/Converters/DownloadProtocolConverter.cs:23]: Sounds good, let's just remove "Custom" from the indexer implementation in the IndexerFormModal.vue select, and the DownloadClientConfig.ts, then let's add a TODO to clean up the custom case later.

File: fe/src/components/domain/search/ManualSearchModal.vue (hunk 16 above)
Comment (therobbiedavis) [reply fe/src/components/domain/search/ManualSearchModal.vue:496]: Ah, I did miss that. The code review window did me dirty and collapsed the code!

File: listenarr.api/Services/Adapters/SlskdAdapter.cs (hunk 15 above)
Comment (therobbiedavis) [reply listenarr.api/Services/Adapters/SlskdAdapter.cs:299]: It was initially meant to call SABnzbd's /history and NZBGet's history() RPC to use for failed download detection, but since slskd doesn't function similarly we can leave it like this, just add a comment similar to the others.

File: listenarr.infrastructure/Migrations/20260403104655_UpdateDownloadProtocol.cs [hunk 17]
```diff
@@ -0,0 +1,58 @@
...
+            @"
+                UPDATE Indexers
+                SET Type = CASE 
+                    WHEN Protocol = 'DirectDownload' THEN 'DDL'
+                    ELSE Protocol
+                END;
+            ");
+
+            migrationBuilder.DropColumn(
+                name: "DownloadType",
```
therobbiedavis: Hey @T4g1, was this migration manually edited? Curious as to why in the Down() we're not dropping `Protocol` but instead `DownloadType` which doesn't exist.

listenarr.api/Services/Adapters/SlskdAdapter.cs
```diff
@@ -0,0 +1,313 @@
...
+                if (transfer == null)
+                {
+                    // Download no longer in the download client
+                    await service.HandleFailedDownloadAsync(download, client, db, settings, $"Slskd does not have the download {download.Id}", ct);
+                    continue;
+                }
+
+                await service.UpdateDownloadProgressAsync(download, transfer.Progress, transfer.Size - transfer.Downloaded, "downloading", db, ct);
+
+                if (transfer.Status == "completed")
```
therobbiedavis: Are we sure that the `state.ToString()` on line 233 will return "completed" lowercase? Should we add a guard like .ToString().ToLower() to ensure this comparison always works?

therobbiedavis (changes requested):
Found a couple more things that I have questions about

listenarr.infrastructure/Migrations/20260403104655_UpdateDownloadProtocol.cs
T4g1: Yes, Its handmade, I'll fix that right away

listenarr.api/Services/Adapters/SlskdAdapter.cs
T4g1: So, after investigation (I figured that I completely forgot to actually test that part), I added several things:
* Configuration of Download path is now correctly used (simply ignored before)
* If not present, API call to get the configured one on Slskd side (so the tooltip in the frontend stays valid eh eh)
* Added actual test for that part
* Also added a block long comment explaining how Slskd handles the file/directory thing and collisions, I might do a PR on Slskd side to update the API and simplify things in that regards if i remember that issue long enough (might be worth an issue somewhere)

Lastly, I did not use path mapping in any way here, I assumed that if Slskd is dockerized, the configured path will be a docker specific one and hence not accessible by Listenarr but in that case, user can still override it in Download Client configuration and input a correct one as visible by Listenarr. UI/documentation about that behavior might need to be updated though

T4g1:
@therobbiedavis I believe I've processed all your comments, let me know if I missed anything

listenarr.api/Controllers/LibraryController.cs
therobbiedavis: Can we add a branch here for slskd?

listenarr.application/Search/AutomaticSearchService.cs
```diff
@@ -625,12 +590,12 @@ private async Task<string> GetAppropriateDownloadClientAsync(SearchResult search
             var clients = await configurationService.GetDownloadClientConfigurationsAsync();
             var enabledClients = clients.Where(c => c.IsEnabled).ToList();
 
-            _logger.LogInformation("Looking for {ClientType} client. Found {Count} enabled download clients: {Clients}",
```
therobbiedavis, line 633: Also need a branch for slskd here too

therobbiedavis (changes requested):
I think we need some branching logic for slskd otherwise I don't think automatic downloads will work.

listenarr.api/Services/Adapters/SlskdAdapter.cs
therobbiedavis: > Lastly, I did not use path mapping in any way here, I assumed that if Slskd is dockerized, the configured path will be a docker specific one and hence not accessible by Listenarr but in that case, user can still override it in Download Client configuration and input a correct one as visible by Listenarr.

Yes, so that remote path mapping option for the download clients is exactly the use-case you described here. It would take the path from slskd, compare against the remote path mapping, and translate it to the destination path the the user has setup.

For example, my listenarr and qbit run both in docker with different paths. Listenarr can't access `/downloads/listenarr` because that doesn't exist in the listenarr docker container, only in the qbit docker container, but `/server/downloads/complete/listenarr/` does exist in listenarr. Both paths map though to the same destination on the file system, it's just how they're mounted in the docker config.

image

listenarr.application/Search/AutomaticSearchService.cs
T4g1: Ahh yes, this can be re-written to avoid hardcoding clients in fact, i'll see what i can do

T4g1: Ok, latest commit contains a fix for that, here is where the magic happens: https://github.com/Listenarrs/Listenarr/pull/465/changes#diff-45b210453765f2e5050b2ab454d55b748a861d4597f790c4bbbfb3e4a1054341R135-R148

Let me know what you think of this, the two biggest advantages imho:
* Less troublesome to add DownloadClientAdapters: Minus one hardcoding to do
* Better separation of concerns: Right now both method were duplicates

listenarr.api/Controllers/ProwlarrCompatController.cs
```diff
@@ -555,20 +556,18 @@ public async Task<IActionResult> PutIndexer(int id, [FromBody] System.Text.Json.
                         indexer = new Listenarr.Domain.Models.Indexer
                         {
                             Name = string.IsNullOrEmpty(nameFromPayload) ? (string.IsNullOrEmpty(baseUrlFromPayload) ? "Prowlarr Indexer" : baseUrlFromPayload) : nameFromPayload,
-                            Implementation = string.IsNullOrEmpty(implementationFromPayload) ? "Custom" : implementationFromPayload,
+                            Implementation = Implementation.Custom, // TODO: Prowlarr use Cardigann
```
therobbiedavis: I think this will break prowlarr imported indexers. Prowlarr sends over Torznab or Newznab for the implementation in the payload.

listenarr.application/Search/AutomaticSearchService.cs
therobbiedavis: I think this looks good! Thank you!

therobbiedavis (changes requested):

listenarr.api/Services/Adapters/SlskdAdapter.cs
T4g1: After investigation and according to the codebase, as long as QueueItem content path is filled, path mapping is already applied (I do not understand why some specific adapters are still calling that service manually, imho, Adapters should not care about this), I'll add some test to make sure it's the case though

T4g1: Test added, indeed, the QueueItem contentPath is already processed if the job contains it as sourcePath.

I removed the custom normalization from remote path mapping too to normalize normalization methods

listenarr.api/Controllers/ProwlarrCompatController.cs
T4g1: It should be fixed but im not sure about the schema sent from Prowlarr so I based the logic on what was already there, feel free to check

T4g1:
PR rebased on canary

therobbiedavis:
Heads up: #599 changes the download-client and DDL/search handoff architecture substantially. Since this PR adds both an indexer and a download client, it should be rebased on #599 and split/refactored around the new boundaries before review.

For the download-client side, please follow the post-#599 structure: thin adapter facade, client-specific infrastructure workflows, full queue snapshot vs targeted monitor polling, gateway-owned remote path mapping, and client-specific import item resolution. For any direct-download behavior, use additive source policies/artifact planning rather than embedding generic fetch behavior in the provider.

therobbiedavis:
#717 replaces the old remote/source-path normalization contract: client-reported paths are external filesystem identities and must preserve legal whitespace/separator/case semantics until translated under the owning filesystem context. This older Slskd PR remains a separate feature, but it will need a substantial rebase; in particular, please do not restore the RemotePathMappingService NormalizeStoredPath handling shown in this branch because that would reintroduce the class of source-path bug fixed by #717/#528.

PR #466, opened 2026-04-01
Security hardening, bug fixes, and test improvements

therobbiedavis:
This PR is a focused security and quality pass. No new features — just making things safer, more reliable, and easier to maintain.

### Fixed

- **Log injection across the board** — Anything user-controlled (file paths, search titles, download IDs, ASINs, URLs) now goes through `LogRedaction` before hitting the logger. Prevents newline injection attacks that could corrupt or spoof structured log output.
- **SSRF in NZB and torrent downloaders** — `NzbgetAdapter` and `TorrentFileDownloader` now validate outbound URLs via `OutboundRequestSecurity.TryValidateExternalHttpUrl` before making any request. Every redirect hop is checked too, not just the initial URL. Private/LAN hosts are permitted since user-configured indexers typically run on Docker networks.
- **SQL injection in debug schema endpoints** — Column and table names in raw `PRAGMA`/`SELECT` calls in `LibraryController` are now double-quoted and validated against a whitelist. Low risk in practice (internal endpoints), but still worth fixing.
- **Race condition in `LoginRateLimiter`** — `BlockUntil` and the failure counter are now protected by a per-entry lock. Previously two concurrent login attempts could both slip through the rate limit window at the same time.
- **Raw session token in `ClaimsPrincipal`** — `SessionService` was emitting the actual session token as a claim, which could leak through middleware or claim-inspection APIs. Removed.
- **Column injection in `StartupDbNormalizer`** — Column names resolved at startup are now whitelisted + regex-validated before being dropped into `ALTER TABLE` SQL.
- **Bulk audiobook update wasn't transactional** — `BulkUpdateAudiobooks` looped over audiobooks and saved after each one, so a failure halfway through left a partial update in the database. It's now wrapped in a single transaction; everything commits or nothing does.
- **`FileMover` could delete the original before confirming the copy** — Hardlinks and copies now go to a temp file first, then get atomically renamed over the destination. The original is never removed until the replacement is confirmed on disk. Temp paths use `Path.Join` instead of `Path.Combine` so a rooted filename can never silently discard the destination directory.
- **`QbittorrentAdapter` auth bypass** — `AddAsync` now throws on any non-success, non-403 login response instead of logging a warning and falling through. Previously a qBittorrent API returning an unexpected status code could result in a torrent being added without authentication.
- **`QualityProfile.Equals` was comparing against itself** — The `Equals(object?)` override shadowed the variable name (`qd` instead of `other`), so it always delegated to a self-comparison. Renamed.
- **`AccountController.Me()` boolean precedence bug** — A missing parenthesis in the auth check meant an unauthenticated request could fall through to the session lookup path.
- **CS8620 nullability warning in `UnmatchedScanBackgroundService`** — Tuple elements passed to `MultiFileImportPlanner.BuildPlans` now use explicit named fields and a `string?` cast to match the expected parameter type.

### Changed

- **`FileMover` P/Invoke modernised to `LibraryImport`** — `[DllImport]` replaced with `[LibraryImport]` (source-generated, .NET 7+) for `CreateHardLink` on Windows and `link` on Linux/macOS. Eliminates runtime reflection overhead and satisfies CA1060/SYSLIB1054. `AllowUnsafeBlocks` enabled in the API project to support the source generator.
- **Prowlarr compat controller moved to `/api/v1/prowlarr`** — Was mounted at `/api/v1`, causing route ambiguity with `IndexersController`. All Prowlarr compat endpoints are now under `/api/v1/prowlarr/`. (Thanks @T4g1)
- **`update_version.py` now validates before it modifies** — The CI version bump script now checks that `NEW_VERSION` is set and matches `MAJOR.MINOR.PATCH[.REVISION]` before touching any files. Missing or malformed values exit early with a clear error message. File write failures are also caught and reported.
- **Duplicate entries removed from `AudiobookOnlyFilter`** — `suffixIndicators` had `"Paperback –"` and `"Hardcover –"` listed twice. Cleaned up.
- **Bulk update test migrated to SQLite in-memory** — The EF in-memory provider doesn't support transactions, so `LibraryController_BulkUpdateTests` now uses a real SQLite in-memory connection. Required FK rows are seeded in the constructor. The class implements `IDisposable` to clean up the connection.
- **Reflection tests now guard against renamed private methods** — Every `GetMethod()` call in `IndexersNewznabParsingTests` is followed by `Assert.NotNull`, and `Invoke()` calls use the null-forgiving operator. A renamed internal method now produces a clear assertion failure instead of a cryptic `NullReferenceException`.
- **Mock HTTP handlers in `DownloadService_ImportTests` no longer share response instances** — Returning the same `HttpResponseMessage` from a mock handler multiple times breaks on the second call because the content stream is already consumed. Handlers now create a fresh response per invocation via local factory methods, which also satisfies CodeQL's IDisposable-not-disposed analysis.


therobbiedavis:
@codex review

PR #468, opened 2026-04-02
feat: add session cookie for browser-native image authentication

lzinga:
## Summary

Adds an HttpOnly session cookie on login so browsers can authenticate image requests and resource loads without JavaScript workarounds.

## Problem

When auth is enabled, `` can't attach an `Authorization` header since browsers don't support custom headers on resource loads. The frontend works around this with a composable that fetches each image via JS, converts it to a blob URL, manages cache maps per component, and revokes blob URLs on unmount. It works but it adds a lot of complexity and a visible placeholder flash on every image.

### Added
- Session cookie (`listenarr_session`) set on login for browser-native image authentication (HttpOnly, SameSite=Strict, Secure on HTTPS)
- Cookie fallback in `SessionAuthenticationMiddleware` after Bearer, X-Session-Token, and SignalR query param
- Cookie deletion and token extraction from cookie on logout
- 7 integration tests for cookie auth, revocation, Bearer priority, API key regression, and auth-disabled mode
- Shared `TestHelpers.cs` for common test utilities

### Changed
- Session duration constants in `SessionService` consolidated as public static fields so cookie expiry stays in sync with DB session expiry
- `TestStartupConfigService` moved from `AuthenticationMiddlewareTests.cs` into shared `TestHelpers.cs`

Existing Bearer token and API key auth are completely unaffected. The cookie is only used when no header is present.

## Security

- HttpOnly, SameSite=Strict, Secure on HTTPS
- Same opaque token and DB hash validation as Bearer
- Server-side revocation works instantly
- Image endpoints are GET-only, no CSRF risk

## Tests

7 new integration tests covering cookie auth, invalid/revoked tokens, Bearer priority, API key regression, and auth-disabled mode. 602 total tests pass.

## Next Steps

With this in place a follow-up PR can simplify the frontend by replacing the blob URL fetching in `useProtectedImages` with direct image URLs since the browser now handles auth automatically via the cookie.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [0.2.69] - 2026-04-04
 
 ### Added
 - **Session cookie for browser-native image authentication ([#468](https://github.com/Listenarrs/Listenarr/pull/468)):** Login now sets an HttpOnly `listenarr_session` cookie (SameSite=Strict, Secure on HTTPS) alongside the existing JSON token response. `SessionAuthenticationMiddleware` checks this cookie as a fallback after Bearer header, X-Session-Token, and SignalR query param, enabling browsers to authenticate image requests and other resource loads without JavaScript workarounds. Existing Bearer token and API key auth are unaffected.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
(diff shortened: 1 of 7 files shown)


therobbiedavis:
@lzinga Can you rebase this, update the changelog.md, and the PR description into Added, Fixed, Changed, Removed format? You can take a look at any other PR for this. This is because the PR sent to Discord as a message.

lzinga:
Rebased onto the latest canary and added the CHANGELOG entry using the `[Unreleased]` section from the Keep a Changelog spec that the file references in its header. No conflicts in my files, though technically a rebase is only needed when there are conflicting changes between the PR and the target branch. If there aren't any, GitHub can merge it cleanly as-is. But no harm in keeping things up to date either way.

I've also updated the PR description to include the Added/Changed format:

### Added
- Session cookie (`listenarr_session`) set on login for browser-native image authentication (HttpOnly, SameSite=Strict, Secure on HTTPS)
- Cookie fallback in `SessionAuthenticationMiddleware` after Bearer, X-Session-Token, and SignalR query param
- Cookie deletion and token extraction from cookie on logout
- 7 integration tests for cookie auth, revocation, Bearer priority, API key regression, and auth-disabled mode
- Shared `TestHelpers.cs` for common test utilities

### Changed
- Session duration constants in `SessionService` consolidated as public static fields so cookie expiry stays in sync with DB session expiry
- `TestStartupConfigService` moved from `AuthenticationMiddlewareTests.cs` into shared `TestHelpers.cs`

Issue #233, opened 2025-11-24
[Feature Request] Audiobookshelf integration


kman3107:
Before anyone bites my head off, I know that Audiobookshelf integration is in the roadmap.
I was hoping to have this issue as a means of keeping track off when this is added, since I don't wish to start using Listenarr until Audiobookshelf integration is added.

Thanks!

m4bard:
> **Note added 2026-08-17.** The framing below understates how much already works with no code at all, and overstates how much the rescan endpoint matters. Nothing in it is wrong as written, so I have left it intact rather than striking anything. See [the follow-up comment](https://github.com/Listenarrs/Listenarr/issues/233#issuecomment-5318679519) for the measured version.

I went looking at what this would involve, mostly out of self-interest since I serve books through Audiobookshelf myself. I came away thinking the awkward part is not Audiobookshelf, and I would rather ask than assume.

The thing I did not expect: Listenarr has no downstream server integration of any kind. Not Plex, Emby, Jellyfin, Kavita, Subsonic, Komga, Navidrome or Kodi. I checked with a word-boundary search, because a plain grep for "plex" matches "complex properties" and "complexity: medium" and looks like a hit.

So the notification surface today is entirely human channels: Pushbullet, Telegram, Slack, Discord, Pushover, NTFY and Zapier, per `WebhookConfiguration.Type`. Each of those answers "tell a person something happened". None answers "tell a server to go and look at the disk again".

That reframed it for me. Adding Audiobookshelf is not adding one more provider next to an existing Plex one, it is opening a category that does not exist yet. Which might also explain why this issue has been open a while without a clear answer: the request reads small and the decision underneath it is not.

The *arr lineage does already have a shape for this. Readarr keeps media servers as notification providers, in the same folder as the chat ones, implementing the same interface, so Plex, Kavita, Subsonic and Synology sit alongside Discord and Ntfy. Kavita is the closest analogue to Audiobookshelf, being a self-hosted reading server rather than a chat target. One available answer, then, is that a media server is a notification provider whose action happens to be an HTTP call rather than a message.

On the Audiobookshelf side the surface is smaller than the word "integration" suggests. I looked at how another project in this space talks to it. Their abstraction has eight methods, but seven exist because that project does not own the files and has to ask Audiobookshelf what the user has. Listenarr already knows what the user has, because it put the files there. What would actually be needed is one call:

```
POST /api/libraries/{libraryId}/scan
```

[...]

Issue #494, opened 2026-04-06
[Feature Request] option for keep local files while library import from root folder


LoveBootCaptain:
currently I can't import my local files from my audiobook root folder whithout moving them ... they are already in "root folder/author/title" file structure, files named and properly tagged. so i don't need to move or copy the files into the root folder (which is my only folder (default folder) I configured in settings->root folder. I just want to add them to my collection not monitoring them until I see a possibility to upgrade my version of the audiobook.

**sonarr example local library import without moving files**

<img width="1448" height="778" alt="Image" src="https://github.com/user-attachments/assets/ebc6de8c-cba8-42c7-aa1e-4e0fe83f1b7c" />

LoveBootCaptain:
further improvement:

create the option to process like "file renaming" to local library files after importing them first like described above.
most *arr apps provide such a workflow for importing local libraries and file management afterwords.

**sonarr example renaming to configured pattern**

<img width="1051" height="360" alt="Image" src="https://github.com/user-attachments/assets/cec36a24-03ab-4278-8cfe-27c84ec6a62b" />

T4g1:
If I undestand correctly:
* Files should be matched to an Audiobook
* The audiobook should be added in Listenarr
* File location should not be impacted (add option to neither move nor hardlink/copy)
* Link the file(s) to the correct audiobook with the actual location
* Audiobook should not be monitored: Add option to not monitor them
* It should be possible to scan arbitrary folder and not only root folders

Is that correct ?

LoveBootCaptain:
> If I undestand correctly:
> 
> * Files should be matched to an Audiobook
> * The audiobook should be added in Listenarr
> * File location should not be impacted (add option to neither move nor hardlink/copy)
> * Link the file(s) to the correct audiobook with the actual location
> * Audiobook should not be monitored: Add option to not monitor them
> * It should be possible to scan arbitrary folder and not only root folders
> 
> Is that correct ?

perfectly summarized

LoveBootCaptain:
awesome feature ... just tested it ... works perfect for me.
thx :)

PR #499, opened 2026-04-07
AudiobookShelf Integration

redwiz666:
## Summary

Added AudiobookShelf Connection in settings. 

image
image

This gives us the option to scan after importing a new book or when a download completes. 

Once ABS libraries are loaded you can like them to root folders
image

This allows for different paths in ABS and Listenarr. This is mainly used for the import options. 
Had to modify the Library Import dropdown to keep RootFolders and ABS Libraries seperate
image

When importing from ABS it doesn't move the files it just Adds the books from their current location, will convert the path to once accessible to Listenarr if its mapped to a RootFolder. Once ABS books are imported they will show up in your Listenarr library. 

## Testing

Tested importing over 500 books from ABS version 2.33.1. The only issue I have found during my testing is that when a book in ABS doesn't have its metadata filled out (mainly ASIN) then it will import as missing. Since Listenarr requires ASIN or ISBN to use the scan feature. If later this gets updated to be able to scan using something else then it should work fine. I didn't want to rewrite your scan function just to make this work 100%. The scan feature would need to be adjusted similar to when/If you allow for manual import later on.


T4g1:
I dont understand the way this integration works. Why should Listenarr manage a third-party settings ? If I understand correctly, the goal of your PR is to import books managed by audiobookshelf in specific folders rather than the configured root folder(s) ?

The way I see it is that Listenarr should import media and metadata in the configured folders and it's the responsibility of third-party to scan those just like Jellyfin/Plex with Radarr/Sonarr/Lidarr.

Eventually, shouldn't Listenarr simply do API calls to Audiobookshelf on imports to trigger scan on it's side ? More like a plugin would do it

Also, I believe this PR closes https://github.com/Listenarrs/Listenarr/issues/233 and update the roadmap but if find both of those lacking in terms of feature description/discussion

redwiz666:
Maybe I didn't explain it well. It still does it within the root folder but I added a way for it to translate audiobookshelf path to listenarr path. This is similar to how sonarr/radarr does it for download clients. For example if in audiobookshelf your books are saved in /audiobooks but in listenarr the root folder is /nas/audiobooks. When importing it will convert the audobookshel path to the rootfolder before adding the book to the database

T4g1:
Isn't what you describe there the feature requested here: https://github.com/Listenarrs/Listenarr/issues/494 ? How does it relate to Audiobookshelf at all ?

Wouldn't this behavior solve the issue you have:
* Allow import of arbitrary folder
* Monitor added audio books to current path (maybe create a specific root path for those)
* Trigger events on import succeeded, download completed, and/or other events
* Write a plugin that interface those events to Audiobookshelf

This way, the code can stay agnostic of third-party, here it's really adding a lot of Audiobookshelf specificity into the backend, I know it's already what the code does for a lot of thing but low coupling is probably the direction we should follow

For context, im trying to understand what problem the PR solves so I can do a review of it, worst case you'll have to wait for Robbie's return

redwiz666:
The path translation is only part of the problem. The main part is adding in the integration to audiobookshelf to allow importing the books directly with metadata and also notifying audiobookshelf of new books and to scan the new book. Since a lot of people probably will use listenarr to download the books but audiobookshelf to listen to them.

T4g1:
I understand but I do not think it is required to have any Audiobookshelf knowledge in Listenarr to do that. We can do exactly the same thing that Jellyfin does with Sonarr/Radarr/... with the features i describe in my comment. Don't you agree ?

To integrate with audiobook shelf, you would have to first support #494 so we can import library without impacting the files themselves: 
* Add a root folder specific for Audiobookshelf (or only one that already contains the library)
* Scan and import audiobooks

Now, there should be all your audiobooks from Audiobookshelf in the Listenarr library.

And then, all you need is a plugin system with an event queue with subscriber/publisher to trigger API calls to Audiobookshelf so it can import new files/refresh library on some given events (that feature is not in issue form but I saw it somewhere on the roadmap)

lzinga:
I agree with @T4g1 here. There's an unneeded level of dependency on Audiobookshelf implemented throughout the core pipeline. Listenarr should stay media-player-agnostic, the same way Sonarr/Radarr don't have Jellyfin or Plex code in their download processor. A post-import event/notification system would serve ABS, Plex, Jellyfin, and any future integration without coupling them into the codebase.

Also worth noting there are no tests added for the 5,600+ lines of new code, and the existing test suite would fail to compile due to the new `IAudiobookshelfService` constructor parameter added to `CompletedDownloadProcessor`. There are at least 6 call sites in `CompletedDownloadProcessorTests.cs` that would need updating.

I'd recommend closing this and revisiting after a generic notification/event system is in place that ABS (and others) can hook into.

therobbiedavis:
Hey there @redwiz666, first off thanks for taking to the time to contribute! 

From my understanding of this PR it seems like this is pretty complex and solves for specifically one program without taking a more macro-view of the roadmap. I would like to build a foundation for multiple integrations like Audiobookshelf, Plex, Jellyfin, Emby, ReadMeABook, or any other audiobook consumer program that comes out in the future.

With that vision, I don't see a need for an ABS integration to add an ABS root folder. We should first solve #494, then let the user add their ABS root folder independently for the library import. This would decouple ABS from Listenarr. Then we should build out the post-import event integrations for ABS and others, where Listenarr just sends to those applications. This would make Listenarr more flexible while solving what you're accomplishing. The idea being that it would be one way communication: Listenarr -> ABS/Plex/etc. Those applications do not need to send any info back to Listenarr.

If you would mind kindly refactoring more towards that direction, I'll happily take a deeper review of your PR. Please also make sure to add relevant unit tests.


Thank you!

redwiz666:
@therobbiedavis Sorry this was my first PR for a project link this, only ever worked on personal projects. 

I will work on refactoring it towards that direction and update when I get it done.

therobbiedavis:
> @therobbiedavis Sorry this was my first PR for a project link this, only ever worked on personal projects. 
> 
> 
> 
> I will work on refactoring it towards that direction and update when I get it done. 

No worries, @redwiz666, this is actually my first real public project so we're in the same boat 😄

T4g1:
> @therobbiedavis Sorry this was my first PR for a project link this, only ever worked on personal projects.
> 
> I will work on refactoring it towards that direction and update when I get it done.

fyi, this PR is already taking care of the issue #494 we mentionned

kuldan5853:
Just to add to this - I have played around with listenarr a bit now and it wasn't able to automatically match any of my books that I copy/pasted for testing from audiobookshelf, even if I had embedded metadata - the search code is too specific to match even if there is minor differences in the title. 

This is especially apparent for German "Hörspiel" content where the author is irrelevant and only the series / title matter - I wasn't able to successfully import anything into listenarr in my (short) testing period. 

Since I would use listenarr the same as OP wanted to (integrating with Audiobookshelf) having a hook where Listenarr is able to simply import all library items and their paths (and metadata) directly from the server would obviously perfect - even though I understand that a tool agnostic option would be preferred. 

I'm just adding this comment to say that just pointing listenarr at the same root folder as an already existing library and then trying to rely on listenarrs matching code is not really a very viable option - especially if a perfectly maintained metadata source "of truth" would already be available. It might not matter if you have a handful of books, but in my case with >8000, 1:1 matching each file by hand would not be feasible. 


Please don't take this as criticism, as this is obviously only my individual opinion - just wanted to expand on why redwiz666 probably built what he built in the way he did it.

T4g1:
There's a couple question I still have with the use case this solves. Right now, I assume the use case is the following:
Given a current library managed in Audiobookshelf, the user wants to automaticaly import  the audiobooks and the files to Listenarr. I also assume this should be done only when adding new root folder or scanning a folder.

This is helps the user because he is not forced to manualy add every audiobook he already owns into Listenarr

And here is how I believe the current workflow is designed:
- For manual import: The frontend is responsible to build the list of files and matched audiobooks that should be imported
- For automatic import: It's still a TODO

The way that list of items/audiobook association is constructed works like this:
- In front end: Select a root folder and trigger a scan of the associated path
- IUnmatchedScanQueueService is used to create a scan job that is then processed eventualy whith UnmatchedScanBackgroundService
- Front end receive a list of files that are not associated with any Audiobook and some metadata associated with them

There's an issue with the current implementation:
- The UnmatchedScanQueue feature is very specific. We could extend it to return the list of files in the given path and, eventualy, the related audiobook when relevant
- Frontend would then only shows the ones that are not associated with audiook yet.

With that done, we could then extend the ScanJob object to support the addition of a scanner to use. Then, we could define a IDirectoryScanner interface (in a similar fashion to what is done for Download Clients):
- Default implementation: Returns all relevant files and audiobook associated to them
- Audiobookshelf implementation: This is where we could leverage the audiobookshelf API and produce a pre-association of files with audiobooks (those would not be persisted yet)
In the front end, we can ask the user which scanner he wants to use

We would also need the following:
- The Import controller request DTO should support non existing audiobooks (probably in a similar way to how it's already done for adding audiobooks somewhere else)
- The front end would need a way to make the distinction between:
   * Unmatched file
   * Unmatched file with suggested audiobook
   * Matched file with existing audiobook
- The Import controller needs to create audiobooks when they do not exist yet (delegating to appropriate services of course)
- A way to configure the scanner (so the configuration does not ends up directly on the general configuration but stays specific to the IDirectoryScanner or something like we have fo DownloadClientConfiguration)

With that kind of implementation, we can also hide the scanner choice when the user does not have a valid configuration for it, so by default, the interface stays the same for non-audiobookshelf users.  We can support any other third-party with minimal effort, assuming we keep configuration generic, we only need a new IDirectoryScanner implementation and it's done

Here is a very simplified diagram of that (ImportController should go through Infra/Application and UI does active polling to get the job result instead of the synchronous interaction shown here):
```mermaid
graph TD
    subgraph Frontend ["Frontend (UI)"]
        UI_ScannerSelector["Scanner Selector (Default vs ABS)"]
        UI_ImportTable["Import Grid: <br/>- Unmatched<br/>- Suggested (ABS)<br/>- Matched (Existing)"]
    end

    subgraph API ["Application Layer (API)"]
        ImportController["ImportController<br/>(Handles New/Existing Audiobooks)"]
        ScanQueue["RootFoldersController"]
    end

    subgraph Core ["Infrastructure / Core Logic"]
        BackgroundService["UnmatchedScanBackgroundService"]
        ScannerFactory["Directory Scanner Factory"]
        
        subgraph Scanners ["IDirectoryScanner Implementations"]
            DefaultScanner["DefaultScanner<br/>(Local Disk Scan)"]
            ABSScanner["AudiobookshelfScanner<br/>(ABS API + Metadata Mapping)"]
        end
    end

    subgraph External ["External / Storage"]
        Disk["Local File System"]
        ABS_API["Audiobookshelf API"]
        DB[(Listenarr DB)]
    end

    %% Workflow Connections
    UI_ScannerSelector -->|"1. Request Scan (Type + Path)"| ScanQueue
    ScanQueue -->|2. Process Job| BackgroundService
    BackgroundService -->|3. Get Implementation| ScannerFactory
    ScannerFactory --> DefaultScanner
    ScannerFactory --> ABSScanner

    DefaultScanner -->|Scan| Disk
    ABSScanner -->|Query| ABS_API

    Scanners -->|4. Return Enriched List| BackgroundService
    BackgroundService -->|5. End of job result| ScanQueue
    ScanQueue -->|6. List of files with/without audiobook| UI_ImportTable

    UI_ImportTable -->|"7. Confirm Import (DTO)"| ImportController
    ImportController -->|8. Create Missing & Link Files| DB
```

lamb0985:
How can I get this docker image of listenarr with abs integrated? I find it pretty annoying that listenarr only allows import when it matches metadata...even though abs has metadata.json already present next to the files. I'm pretty much only interested in listenarr for managing downloads and folder/file naming (once naming issue is fixed). The fact that it auto populates information in order to import that I then have to undo to match what abs has already is frustrating.

therobbiedavis:
> How can I get this docker image of listenarr with abs integrated? I find it pretty annoying that listenarr only allows import when it matches metadata...even though abs has metadata.json already present next to the files. I'm pretty much only interested in listenarr for managing downloads and folder/file naming (once naming issue is fixed). The fact that it auto populates information in order to import that I then have to undo to match what abs has already is frustrating.

@lamb0985 As it stands the contributor still needs to refactor and rebase this PR, and that's not even mentioning the lingering questions @T4g1 still has. At this juncture this is on the backburner in terms of priority. There are plenty of annoyances in Listenarr we are trying to solve, but a feature request is almost never going to be a higher priority than bug fixes.

PR #503, opened 2026-04-09
Add option to leave files where they are while importing library

### LINKED ISSUE #494: [Feature Request] option for keep local files while library import from root folder

T4g1:
## Summary
* Fix #494 

### Added
* Ability to add root folders while selecting the root folder from which we want to import files
* Ability to leave files in place when importing library
* Ability to set the desired monitoring for audiobooks added through library importation

## Testing
* Check that the import does not change anything in the files when it is not asked to modify them

### Next steps
* #504 : It should be possible to process the imported files to rename them using the pattern defined in the settings. Some tests for that should be added too (see InteractiveManualImport_DontMoveAnything_DontRenameAnything)


listenarr.api/Controllers/ManualImportController.cs
```diff
@@ -93,293 +100,190 @@ public ActionResult<object> Preview([FromQuery] string path)
...
+                        .Where(item => !string.IsNullOrWhiteSpace(item.FullPath))
+                        .Select(item => item.FullPath!)
+                        .Where(FileUtils.IsAudioFile))
+                : Array.Empty<FileUtils.AudioMatchProfile>();
+
+            _logger.LogDebug("Manual import batch: {ItemCount} items", orderedItems.Count);
+
+            foreach (var item in orderedItems)
+            {
+                var fileCount = orderedItems.Select(f => f.MatchedAudiobookId).Count();
```
therobbiedavis: This will count every item in the batch regardless of audiobook. Example: if you import 1 file for Book A and 1 file for Book B, both files see fileCount = 2 and are treated as multi-file imports, causing wrong naming.

fe/src/views/library/LibraryImportView.vue
```diff
@@ -175,6 +178,12 @@
       :folders="rootFoldersStore.folders"
     />
   </div>
+
+  <RootFolderFormModal
+    v-if="addRootFolder"
+    @saved="refreshRootFolders"
+    @close="addRootFolder = false"
```
therobbiedavis: This closes the modal without saving and resets addRootFolder but doesn't reset selectedFolderId. Therefore it stays null, so the select continues to show "Choose another folder...". You need to restore the previous folder on close.

listenarr.api/Services/ImportService.cs
```diff
@@ -1247,4 +1247,23 @@ public Task<bool> MoveFileAsync(string sourceFile, string destFile)
             }
         }
     }
+
+    public async Task PerformActionOn(Listenarr.Api.Services.FileMover.FileAction action, string source, string? destination = null, HashSet<string>? usedDestinations = null)
```
therobbiedavis: why not use `usedDestinations`? It's used for collision detection in the filemover.

listenarr.api/Services/IFileMover.cs
```diff
@@ -9,5 +9,14 @@ public interface IFileMover
...
         Task<bool> MoveDirectoryAsync(string sourceDir, string destDir);
         Task<bool> CopyDirectoryAsync(string sourceDir, string destDir);
+
+        /// <summary>
+        /// Perform the given action on the given file
+        /// </summary>
+        /// <param name="action">What we want to do with the file</param>
+        /// <param name="source">File</param>
+        /// <param name="destination">Optional destination of the action</param>
+        /// <param name="usedDestinations">To remove probably</param>
```
therobbiedavis: If we're removing then lets refactor and remove, otherwise lets update this description.

listenarr.domain/Models/Audiobook.cs
```diff
@@ -61,7 +62,13 @@ public class Audiobook
         public long? FileSize { get; set; }
 
         // Base path for multi-file audiobooks (common root directory)
-        public string? BasePath { get; set; }
+        public string? BasePath {
+            get
+            {
+                return FileUtils.NormalizeStoredPath(field);
```
therobbiedavis, line 69: Lets instead normalize on the setter so reads are free

listenarr.domain/Models/Configuration.cs
```diff
@@ -84,7 +85,15 @@ public class WebhookConfiguration
     public class ApplicationSettings
     {
         public int Id { get; set; } = 1; // Singleton pattern - only one settings record
-        public string OutputPath { get; set; } = string.Empty;
+        public string OutputPath 
+        {
+            get 
+            {
+                return FileUtils.NormalizeStoredPath(field);
```
therobbiedavis, line 93: same here, normalize on the setter

fe/src/stores/rootFolders.ts
```diff
@@ -49,5 +49,9 @@ export const useRootFoldersStore = defineStore('rootFolders', () => {
     return r
   }
 
-  return { folders, loading, defaultFolder, load, create, update, remove }
+  function getLast() {
+    return folders.value.at(-1);
```
therobbiedavis: this assumes two people are not creating folders at the same time. We might want RootFolderFormModal to emit the saved entity's ID and resolve it directly.

therobbiedavis (changes requested):
Solid job! Thank you! Just a few changes and a request to please update the changelog!

fe/src/stores/rootFolders.ts
therobbiedavis: I know this probably isn't a valid use-case, but also I've learned long ago not to assume how a user will use your software.

fe/src/views/library/LibraryImportView.vue
T4g1: I see, I updated it to be a button instead so we have no issue with select options to solve. There is another case when there is simply no root folder configured yet that would remain an issue even when remembering the last folder selected anyway

2026-04-14 13_35_00-Listenarr - Audiobook Management

wdyt ?

listenarr.domain/Models/Audiobook.cs
T4g1: I would agree but what about values already in DB then ? Are we sure all values are currently normalized already ?

T4g1:
@therobbiedavis Thanks! I did a pass on your reviews, can you check again to see if we are aligned ?

listenarr.domain/Models/Audiobook.cs
therobbiedavis: Yes, so let's include a one-time normalization migration with this to normalize existing BasePaths

fe/src/views/library/LibraryImportView.vue
therobbiedavis: lgtm!

listenarr.domain/Models/Audiobook.cs
therobbiedavis: Yes, so let's include a one-time normalization migration with this to normalize existing BasePaths

listenarr.domain/Models/Configuration.cs
therobbiedavis: let's include a one-time normalization migration with this to normalize existing OutputPaths, then normalize the setter

CHANGELOG.md
```diff
@@ -5,56 +5,15 @@ All notable changes to this project will be documented in this file.
...
-- **Eliminated golang/stdlib CVEs from Docker image (CVE-2023-29403, CVE-2022-30635):** The Debian-packaged `gosu` binary was compiled with Go 1.19.8 and flagged by scanners. Both `Dockerfile` and `Dockerfile.runtime` now build `gosu` from source in a `golang:1.24-alpine` multi-stage builder stage (pinned to `gosu@1.19`, built with Go 1.24.6), replacing the apt package entirely. The resulting static binary is copied into the runtime image.
-- **Eliminated npm bundled CVEs from Docker image (tar CVE-2026-23950 / CVE-2026-29786 / CVE-2026-31802 / CVE-2026-24842 / CVE-2026-23745, minimatch CVE-2026-26996, cross-spawn CVE-2024-21538):** The NodeSource `nodejs` apt package ships npm with an older bundled node_modules tree (tar 6.2.1, minimatch 9.0.5, cross-spawn 7.0.3). After installing Node.js, both Dockerfiles now install `npm@10.9.8` (pinned) and remove the stale apt npm tree (`/usr/lib/node_modules/npm`, `/usr/bin/npm`, `/usr/bin/npx`), leaving only the upgraded npm in `/usr/local`.
-
-## [0.2.67] - 2026-04-02
-
-### Changed
-- **Reduced redundancy for the manual import modal:** Small improvement that make it easier to add or edit import fields.
+## [0.2.66] - ?
```
therobbiedavis: Changelog needs to be updated from canary or rebased, then your changes would be under 
```## [Unreleased] ```
at the top. We will update this with a version and date before release.

**Review by therobbiedavis — CHANGES_REQUESTED** [review]
@T4g1 Added a couple comments on the existing issues, outside of these it lgtm!

File: listenarr.domain/Models/Audiobook.cs (hunk 5 above)
Comment (T4g1) [reply listenarr.domain/Models/Audiobook.cs:69]: I'm gonna need some pointers on how to achieve that, afaik, migration cannot be used to run arbitrary C# methods on the database and the logic of NormalizeStoredPath seems really complex to translate to SQL

Comment (therobbiedavis) [reply listenarr.domain/Models/Audiobook.cs:69]: Generate a migration like so:
```dotnet ef migrations add NormalizeStoredPaths --project listenarr.infrastructure --startup-project listenarr.api```

then add the following to the generated Up() in the migration:

```
migrationBuilder.Sql(@"
    UPDATE Audiobooks
    SET BasePath = RTRIM(TRIM(BasePath), '/\')
    WHERE BasePath IS NOT NULL AND BasePath != RTRIM(TRIM(BasePath), '/\');

    UPDATE ApplicationSettings
    SET OutputPath = RTRIM(TRIM(OutputPath), '/\')
    WHERE OutputPath != RTRIM(TRIM(OutputPath), '/\');
");

T4g1: Alright but it does not handle PROGR~1 or path like that for windows though

T4g1:
Ready for re-review i believe

listenarr.api/Controllers/ManualImportController.cs
```diff
@@ -93,293 +98,190 @@ public ActionResult<object> Preview([FromQuery] string path)
...
+    /// Import the file into the library
+    /// </summary>
+    /// <param name="item">File to import into the library</param>
+    /// <param name="action">Action to perform on the file</param>
+    /// <param name="sourceDirectory">Directory from which we are importing the file</param>
+    /// <param name="usedDestinations"></param>
+    /// <param name="hasMultipleFile">Indicates if this file is part of multiple files for a same audiobook</param>
+    /// <returns></returns>
+    /// <exception cref="IOException"></exception>
+    private async Task<ManualImportResultDto> ImportFileAsync(
```
therobbiedavis, line 201: ImportFileAsync and GenerateManualImportPathAsync both go back to the database for root folders and settings, even though Start() already fetched them before calling into those methods. We should probably add them as parameters and pass them down to reduce DB hits for data that isn't going to change mid-import.

fe/src/stores/libraryImport.ts
```diff
@@ -461,9 +463,9 @@ export const useLibraryImportStore = defineStore('libraryImport', () => {
         await apiService.startManualImport({
           path: item.folderPath,
           mode: 'interactive',
-          inputMode: inputMode.value,
+          action: action.value,
           includeCompanionFiles: true,
```
therobbiedavis: Let's change this to includeCompanionFiles: action.value !== 'none' that way the backend and the frontend stay in sync, and the backend never tries to process companion files when leaving files in place.

therobbiedavis (changes requested):

listenarr.domain/Models/Audiobook.cs
therobbiedavis: Ah, I forgot about short paths. You're right. Let's just add a TODO comment here and come back to it later because I think it will require more effort than necessary right now.

listenarr.api/Controllers/ManualImportController.cs
T4g1: What about having a cache in IRootFolderService since it's scoped ?

T4g1: Nevermind, it's probably open door to nigthmarish issues if we go that way (or we'd have to handle cache invalidation and things like that probably)

T4g1:
Should be good, let me know if you see other things to change

listenarr.api/Controllers/ManualImportController.cs
therobbiedavis: Could we do the same for settings that we did for rootfolders earlier?

listenarr.api/Models/ManualImport/ManualImportItemDto.cs
```diff
@@ -0,0 +1,43 @@
...
+    [JsonPropertyName("relativePath")]
+    public string RelativePath { get; set; } = string.Empty;
+
+    [JsonPropertyName("fullPath")]
+    [Required]
+    public string? FullPath {
+        get
+        {
+            if (field == null) return null;
+            return Path.GetFullPath(field);
```
therobbiedavis: We may want to add a guard for any traversal attempts here. wdyt?

therobbiedavis (changes requested):

T4g1:
Done :D

therobbiedavis (approved):
lgtm! Just update the changelog from unreleased to 0.2.71 with today's date and we can merge it in (barring any codeql issues)

T4g1:
@therobbiedavis Done, should be ready to merge now :) !

PR #505, opened 2026-04-11
Refactor DownloadClientCategoryFilter as a proper OOP class

T4g1:
## Summary
I'm still trying to make sense of some parts of the code, while doing that I figured this could help to reduce code complexity to start refactoring minor files as well.

This tries to apply proper OOP patterns to the `DownloadClientCategoryFilter` file. I also tried to move files in proper folders to avoid the current situation with all files in the same place which makes them difficult to browse imho

### Added
- `CategoryConfiguration` class

### Removed
- `DownloadClientCategoryFilter` file

## Testing
Test coverage is not impacted as this is just a refactoring, no new feature or behavior change

## Notes
These changes are purely optional


therobbiedavis:
Heads up: #599 restructures the download-client gateway, queue, and adapter boundaries. This refactor appears to touch the older API-layer download-client layout, so it should be rebased and re-evaluated against the post-#599 structure before further review.

Some of the intent may still be useful, but it should fit the new application/infrastructure split rather than the pre-#599 service layout.

Issue #504, opened 2026-04-09
[Feature Request] Allow to rename files to match the configured patterns


T4g1:
> further improvement:
> 
> create the option to process like "file renaming" to local library files after importing them first like described above.
> most *arr apps provide such a workflow for importing local libraries and file management afterwords.
> 
> **sonarr example renaming to configured pattern**
> 
> <img width="1051" height="360" alt="Image" src="https://github.com/user-attachments/assets/cec36a24-03ab-4278-8cfe-27c84ec6a62b" /> 

 _Originally posted by @LoveBootCaptain in [#494](https://github.com/Listenarrs/Listenarr/issues/494#issuecomment-4191398259)_

This seems like more work than the original issue so I took the liberty to split it here

The following may be required to make this work:
* Update front end to expose that feature (how ? where ?)
* Update settings (naming pattern are already there, should this be optional ? what is already done ?)

T4g1:
From next steps here: https://github.com/Listenarrs/Listenarr/pull/503: Related to code coverage, the test added in `ManualImport_MultiFileCollisionTests`: `InteractiveManualImport_DontMoveAnything_DontRenameAnything` could be used as a base for it ?
* Should we add an option to not move file but allow to rename them ?
* Should that option be allowed only on the audiobook view ? (in that case, the test mentionned above is probably not relevant)

LoveBootCaptain:
> * Should we add an option to not move file but allow to rename them ?
> * Should that option be allowed only on the audiobook view ? (in that case, the test mentionned above is probably not relevant)

i think the option should be only in the audiobook view.
when you imported the files with "move" oder "copy" the files should have been renamed already? right?

the option for renaming would only make sense, if you "do nothing" on library import and you decide to rename later or changed your renaming pattern in settings.

lamb0985:
Another issue related to naming structure. If I set single file or multipart file naming with {Title} is is also adding {Subtitle} afterwards. For example, If in a audiobook's metadata I have this:
 
Title: Example
Subtitle: Yup

It is naming the file --> Example - Yup.ext instead of just Example.ext

Is this intended or a bug?

T4g1:
That does sounds like a bug, I created an issue for it here #571, i believe adding the requested feature here does not necesseraly impact the code responsible for that behavior so it's probably better to keep them as separated issues

PR #515, opened 2026-04-17
Fix: inputMode leftovers

### LINKED ISSUE #494: [Feature Request] option for keep local files while library import from root folder

T4g1:
## Summary
* Fix #494, do over of #503 

### Added
* Ability to add root folders while selecting the root folder from which we want to import files
* Ability to leave files in place when importing library
* Ability to set the desired monitoring for audiobooks added through library importation
* 

### Changed
- Added new check to do in CONTRIBUTING.md

### Fixed
- Forgot to adapt some inputMode usage to action

## Testing
* Check that the import does not change anything in the files when it is not asked to modify them

### Next steps
* #504 : It should be possible to process the imported files to rename them using the pattern defined in the settings. Some tests for that should be added too (see InteractiveManualImport_DontMoveAnything_DontRenameAnything)


T4g1:
@therobbiedavis i believe we can delete the branch but i'll leave it up to you

PR #516, opened 2026-04-17
Allows sorting Audiobooks on added date

T4g1:
## Summary
Requested in https://github.com/Listenarrs/Listenarr/discussions/513
Allows user to sort audiobooks on added date and create custom filters for it too

### Added
- Audiobook now store when they were added
- Possibility to sort audiobooks on added date
- Possibility to create custom filters on added date

## Testing

Manualy tested


listenarr.api/Controllers/LibraryController.cs
```diff
@@ -409,6 +408,7 @@ public async Task<IActionResult> AddToLibrary([FromBody] AddToLibraryRequest req
                 SeriesNumber = ToStringOrFirst(metadata.SeriesNumber),
                 Description = ToStringOrFirst(metadata.Description),
                 Publisher = ToStringOrFirst(metadata.Publisher),
+                Added = DateTime.Now,
```
therobbiedavis: The rest of the app uses DateTime.UtcNow everywhere for timestampes. DateTime.Now here will store local server time in the DB and cause a lot of comparison issues later. Can you swap this?

listenarr.api/Services/LibraryAddService.cs
```diff
@@ -118,6 +117,7 @@ public async Task<LibraryAddOperationResult> AddToLibraryAsync(
                 SeriesNumber = ToStringOrFirst(metadata.SeriesNumber),
                 Description = ToStringOrFirst(metadata.Description),
                 Publisher = ToStringOrFirst(metadata.Publisher),
+                Added = DateTime.Now,
```
therobbiedavis: Same DateTime.Now -> DateTIme.UtcNow issue here.

listenarr.domain/Models/Audiobook.cs
```diff
@@ -38,6 +37,7 @@ public class Audiobook
         public string? SeriesNumber { get; set; }
         public List<AudiobookSeriesMembership>? SeriesMemberships { get; set; }
         public string? Description { get; set; }
+        public DateTime Added { get; set; }
```
therobbiedavis: Probably just worth setting the default here to DateTime.UtcNow

fe/src/views/library/AudiobooksView.vue
```diff
@@ -1097,6 +1097,10 @@ const filteredAndSortedAudiobooks = computed(() => {
         av = getAudiobookStatus(a)
         bv = getAudiobookStatus(b)
         break
+      case 'added':
+        av = a.added.toString().toLowerCase()
+        bv = b.added.toString().toLowerCase()
```
therobbiedavis: This will not sort as expected because it's converting a DateTime to a string (e.g. Fri Apr 17 2026...) which will not sort properly. You can use what we already do in the filter evalulator:
`av = new Date(a.added).getTime()`
`bv = new Date(b.added).getTime()`

fe/src/types/index.ts
```diff
@@ -597,6 +597,7 @@ export interface Audiobook {
   seriesNumber?: string
   seriesMemberships?: AudiobookSeriesMembership[]
   description?: string
+  added: Date
```
therobbiedavis: While this is a full UTC timestamp the JSON never delivers this as a date, it should be typed as a string.

fe/src/components/domain/collection/CustomFilterModal.vue
```diff
@@ -93,6 +94,14 @@
                       placeholder="e.g. 2023"
                     />
                   </template>
+                  <template v-else-if="r.field === 'added'">
+                    <input
+                      @input="onDateSelected(r, $event)"
+                      type="date"
+                      class="form-input value-input"
+                      placeholder="e.g. 2023-04-17"
+                    />
```
therobbiedavis: I think we need a value bind here, otherwise won't the datepicker just show blank if the user opens an existing saved filter?

fe/src/views/library/CollectionView.vue
```diff
@@ -1020,6 +1020,7 @@ function mapCatalogItem(book: RemoteCatalogBook, sourcePrefix: 'author-catalog'
     language: book.language,
     publisher: book.publisher,
     narrators: book.narrators || [],
+    added: new Date(),
```
therobbiedavis: I was initially concerned by this because I thought of anything missing added would get the date as now, however after looking at the migration we set a default for any missing, so I think this is ok. This can be marked resolved, just keeping my thoughts here for posterity.

listenarr.infrastructure/Migrations/20260417140631_AddAudiobookAddedDate.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<DateTime>(
+                name: "Added",
+                table: "Audiobooks",
+                type: "TEXT",
+                nullable: false,
+                defaultValue: new DateTime(2026, 1, 1, 0, 0, 0, 0, DateTimeKind.Unspecified));
```
therobbiedavis: Shouldn't we use DateTimeKing.Utc? It's what we use elsewhere. Also why 2026-1-1 as the default for existing rows? Why not DateTime.MinValue?

therobbiedavis (changes requested):

listenarr.infrastructure/Migrations/20260417140631_AddAudiobookAddedDate.cs
T4g1: The reasoning was the following:
- When a user enters a filtering date through the date picker, it will be relative to his timezone thus comparing to UTC could lead to strange behavior. We can fix that by converting the inputed date to UTC based on user known (?) timezone probably
- The default here is to give something in the past (relative to the date this will be merged/released that is not too absurd for the user (should we want to display that added date somewhere in the future) so filtering will work as expected for the user but we can definitely go for DateTime.MinValue if you prefer, non tech people seeing 1970-01-01 will probably freak out comparing to seeing 2026-01-01

listenarr.domain/Models/Audiobook.cs
T4g1: I thought about that but won't it prevent the value to be fetched from anywhere else ?
The setter could be used to normalize given date to UTC though if we really want to

fe/src/types/index.ts
T4g1: I'm not sure I follow what you mean, we expect it to be a Date fetched from the API, the transfer/conversion logic should not determine what data type we want to use right ?

listenarr.infrastructure/Migrations/20260417140631_AddAudiobookAddedDate.cs
T4g1: > We can fix that by converting the inputed date to UTC based on user known (?) timezone probably

Turns out the browser already handles timezone when working with date selector!

listenarr.domain/Models/Audiobook.cs
therobbiedavis: My understanding is it won't prevent fetched values from being assigned, it only applies to new/unset instances.

fe/src/views/library/AudiobooksView.vue
```diff
@@ -1097,6 +1097,10 @@ const filteredAndSortedAudiobooks = computed(() => {
         av = getAudiobookStatus(a)
         bv = getAudiobookStatus(b)
         break
+      case 'added':
+        av = new Date(a.added).getTime().toString()
+        bv = new Date(b.added).getTime().toString()
```
therobbiedavis: Converting it to string is going to break the comparison. Like I said: this will not sort as expected because it's trying to sort literal string like "Fri Apr 17 2026" and "Mon Apr 27 2026" which will not sort as expected.

We need to change this to:
```
av = new Date(a.added).getTime()
bv = new Date(b.added).getTime()
```

Then later before the return add a new conditional, like line 1109:

```
if (typeof av === 'number' && typeof bv === 'number') {
  return (av - bv) * (sortOrder.value === 'asc' ? 1 : -1)
}
```

fe/src/types/index.ts
therobbiedavis: The backend value is a DateTime, but once it comes through JSON the frontend receives a plain string, not a JavaScript Date. Nothing is converting it back automatically.

So added: Date is misleading right now because TS would allow a.added.getTime(), but that would fail at runtime because it's a string. Unless we add a mapping step that converts added with new Date(...), the accurate API type should be added: string.

listenarr.infrastructure/Migrations/20260417140631_AddAudiobookAddedDate.cs
therobbiedavis: The main thing I’m trying to avoid is introducing an Unspecified DateTime into a field that otherwise behaves like a UTC timestamp. If we keep 2026-01-01 as the backfill/default for existing rows, I think it should still be:
`new DateTime(2026, 1, 1, 0, 0, 0, DateTimeKind.Utc)`

I’m less attached to MinValue specifically. The important part is that whatever sentinel/default we choose should be UTC and intentional. DateTimeKind.Unspecified is the bit that can create inconsistent serialization/comparison behavior later.

therobbiedavis (changes requested):

therobbiedavis:
Is this still on your radar @T4g1?

T4g1:
Yes, its on hold but I can update it when you think we should merge it

listenarr.domain/Models/Audiobook.cs
therobbiedavis: Addressed in the rebuilt branch. `Audiobook.Added` is now nullable with a `DateTime.UtcNow` initializer, while both actual library-add paths refresh it immediately before persistence. Existing values loaded by EF still overwrite the initializer normally, and the SQLite mapping restores `DateTimeKind.Utc` when materializing stored timestamps.

fe/src/types/index.ts
therobbiedavis: Addressed. The frontend transport contract is now `added?: string | null`, matching the JSON payload rather than pretending it is a JavaScript `Date`. Consumers explicitly parse the serialized UTC string where timestamp semantics are needed.

listenarr.infrastructure/Migrations/20260417140631_AddAudiobookAddedDate.cs
therobbiedavis: Addressed by removing the fabricated sentinel entirely. The regenerated EF migration adds a nullable `Added` column with no default. After migration, startup repair backfills legacy rows only from durable evidence: earliest `Added` history first, then earliest `AudiobookFile.CreatedAt` as fallback. Rows with no evidence remain `null`, so we never invent a date. Known values are normalized back to UTC on SQLite materialization.

fe/src/views/library/AudiobooksView.vue
therobbiedavis: Addressed. Date Added sorting now parses the serialized timestamp with `Date.parse()` and compares the numeric epoch values. Missing or invalid legacy dates are kept last in both directions, and selecting Date Added defaults to descending/newest-first.

therobbiedavis:
Superseded by #841.

I recreated this work on a maintainer-owned branch so the remaining hardening and review fixes can be landed directly without depending on contributor-fork write access. The replacement PR explicitly credits @T4g1 for the original feature/implementation and preserves the current #516 code state as its starting point.

Closing this PR as superseded rather than rejected.

PR #521, opened 2026-04-17
fix/512: Adds FFprobe for ARM64 and fix remote path mapping while importing downloads

### LINKED ISSUE #512: [bug] Automatic import only imports/moves the first file of multi-part audiobooks
I have a problem regarding import multi-part audiobooks. Every time I download an audiobook that consists multiple MP3/FLAC files Listenarr only moves/imports the first file and leaves the rest in the download folder.

I have no errors in the logfile.

---

**Same behaviour but maybe different cause reported by Phynix in Discord. He gets errors. Text copied from him:**

Looking at my logs, I noticed that the bundled ffprobe seems to crash on my system. It looks like the bundled version is not compatible with ARM architecture. Because ffprobe fails, Listenarr falls back to basic metadata, grabs the first track, renames it to the book's title, and ignores the rest.

Environment:
```
    OS: Debian (running via Docker in CasaOS)
    Architecture: ARM / Raspberry Pi
    Version: 0.2.70.0 (Canary)
```

Logs:
```
 2026-04-08 16:33:31.534 +00:00 [WRN] [Listenarr.Api.Services.SystemProcessRunner] Process runner threw an exception for /app/config/ffmpeg/ffprobe
System.ComponentModel.Win32Exception (8): An error occurred trying to start process '/app/config/ffmpeg/ffprobe' with working directory '/app'. Exec format error
   at System.Diagnostics.Process.ForkAndExecProcess(ProcessStartInfo startInfo, String resolvedFilename, String[] argv, String[] envp, String cwd, Boolean setCredentials, UInt32 userId, UInt32 groupId, UInt32[] groups, Int32& stdinFd, Int32& stdoutFd, Int32& stderrFd, Boolean usesTerminal, Boolean throwOnNoExec)
   at System.Diagnostics.Process.StartCore(ProcessStartInfo startInfo)
   at Listenarr.Api.Services.SystemProcessRunner.RunAsync(ProcessStartInfo startInfo, Int32 timeoutMs, CancellationToken cancellationToken) in /home/runner/work/Listenarr/Listenarr/listenarr.api/Services/SystemProcessRunner.cs:line 38
2026-04-08 16:33:31.534 +00:00 [INF] [Listenarr.Api.Services.MetadataService] Extracted basic metadata from file: File01.mp3
2026-04-08 16:33:31.534 +00:00 [INF] [Listenarr.Api.Services.MetadataRescanService] Updated metadata for file id=1
```

T4g1:
## Summary
* Fix #512 

While investigating that issue, I stumbled upon several issues:

### Filtering of downloaded items
When importing finished downloads, Listenarr tries to figure out which files can be imported, it list files reported by the download client and files found in the download directory.

The issue is that when there is remote path mapping (download client and listenarr running in separate docker for examples), Listenarr can no longer match the files.

The solution is to add a remote path mapping before filtering importable files.

### ffprobe
The ffprobe issue was the original hypothesis for this error, turns out arm architecture was not listed in the binaries URLs we use. I fixed that too so ARM architecture can run ffprobe too.

The file naming service requires disc and track metadata to determines the correct naming to apply. 
The method to do that required both metadata and explicit dick/track data. Those explicit data where always null except in testing. This lead me to believe that the error was originating from that but implicitely, metadata disck and track were also used behind the scene.

By fixing ffprobe, those data should now be correctly retrieved.

To future proof that, i reworked the metadata handling and processing to remove unwanted complexity (and explicit useless disc/track informations). Hopefully, next time an architecture is not supported, the log should be clearer about it.

### Only import one file
After a download is finished, `ProcessMoveOrCopyJobAsync` is called eventualy which seems to be doing two things:
* Move files to destination: For some reason, this does not seems to be working
* Queue a re-scan: I haven't found what that is used for
Then, it updates the download path. Right now, this logic was always setting that value to a particular file (even for multi file downloads)

Then, after this confusing processing, the `CompleteDownloadProcessor.ProcessCompletedDownloadAsync` is called. It use the download path that was set earlier to know which file(s) to process. 

The issue is, if it is a file and the containing folder does not exists, it treats that as a single file import!

So, I fixed that by having `ProcessMoveOrCopyJobAsync` returns a directory and creates it so `ProcessCompletedDownloadAsync` can work as expected

### Added
- Metadata are now used implicitely to determine track/disc when processing multi files download to generate appropriate file names

### Fixed
- Handles ARM64 while browsing for ffprobe download URL
- Handles remote path mapping whild comparing local files to download client reported files
- Fixed `ProcessMoveOrCopyJobAsync` not allowing the rest of back end to process multi-files downloads

## Testing
* Added some test cases that depicts a multi files download processing with remote path mapping set up
* Tested on reald case by releasing a docker build that i used on an ARM64 setup

2026-04-19 22_08_24-t4g1@mediaserver_ _media_wdelements5t_library_audiobooks_Isaac Asimov_Le Cycle d

## Notes
* When upgrading, users should remove the /app/config/ffmpeg folder to make sure Listenarr re-installs the correct architecture on relaunch
* Users should update configuration for multi file naming to: "{Title}-{DiskNumber:00}-{ChapterNumber:00}"
* * Errata: Actually, no because DiskNumber and ChapterNumber are currently treated the same
* The naming pattern computed in `ProcessMoveOrCopyJobAsync` is overwritten while doing the actual moving/copying of files in `CompleteDownloadProcessor.ProcessCompletedDownloadAsync` and that one is not as complete. It uses dick number and chapter as the same value. An issue should be created for that to align both implementation on the same naming computations.


T4g1:
Going back to draft:
```
  "[2026-04-19 20:07:25] Resolved naming metadata: Author='Isaac Asimov', AlbumArtist='Isaac Asimov', Series='Le Cycle de Fondation', Title='Seconde Fondation', Year='2019'",
  "[2026-04-19 20:07:25] Initial destination inside output root: /audiobooks/wdelements5t/Isaac Asimov/Le Cycle de Fondation/Seconde Fondation/Seconde Fondation.jpg",
  "[2026-04-19 20:07:25] Destination dir exists: False PathRoot=/",
  "[2026-04-19 20:07:25] Destination directory does not exist: /audiobooks/wdelements5t/Isaac Asimov/Le Cycle de Fondation/Seconde Fondation. Skipping file move/copy and keeping source: /downloads/complete/Isaac Asimov - Le Cycle de Fondation - Tome 3 - Seconde Fondation/Seconde Fondation Isaac Asimov.jpg",
  "[2026-04-19 20:07:25] Finalizing directory batch using completion path: /downloads/complete/Isaac Asimov - Le Cycle de Fondation - Tome 3 - Seconde Fondation",
  "[2026-04-19 20:07:40] Updated download record with final path: /downloads/complete/Isaac Asimov - Le Cycle de Fondation - Tome 3 - Seconde Fondation",
  "[2026-04-19 20:07:40] Enqueued scan job 8dcac479-1557-45c0-877a-efbe00d5858b for audiobook 4",
  "[2026-04-19 20:07:40] Job completed successfully"
```

`Finalizing directory batch using completion path` step is using the wrong directory

T4g1:
> Going back to draft:
> 
> ```
>   "[2026-04-19 20:07:25] Resolved naming metadata: Author='Isaac Asimov', AlbumArtist='Isaac Asimov', Series='Le Cycle de Fondation', Title='Seconde Fondation', Year='2019'",
>   "[2026-04-19 20:07:25] Initial destination inside output root: /audiobooks/wdelements5t/Isaac Asimov/Le Cycle de Fondation/Seconde Fondation/Seconde Fondation.jpg",
>   "[2026-04-19 20:07:25] Destination dir exists: False PathRoot=/",
>   "[2026-04-19 20:07:25] Destination directory does not exist: /audiobooks/wdelements5t/Isaac Asimov/Le Cycle de Fondation/Seconde Fondation. Skipping file move/copy and keeping source: /downloads/complete/Isaac Asimov - Le Cycle de Fondation - Tome 3 - Seconde Fondation/Seconde Fondation Isaac Asimov.jpg",
>   "[2026-04-19 20:07:25] Finalizing directory batch using completion path: /downloads/complete/Isaac Asimov - Le Cycle de Fondation - Tome 3 - Seconde Fondation",
>   "[2026-04-19 20:07:40] Updated download record with final path: /downloads/complete/Isaac Asimov - Le Cycle de Fondation - Tome 3 - Seconde Fondation",
>   "[2026-04-19 20:07:40] Enqueued scan job 8dcac479-1557-45c0-877a-efbe00d5858b for audiobook 4",
>   "[2026-04-19 20:07:40] Job completed successfully"
> ```
> 
> `Finalizing directory batch using completion path` step is using the wrong directory

This is another issue, this can be handled there: https://github.com/Listenarrs/Listenarr/issues/528

listenarr.api/Services/DownloadProcessingBackgroundService.cs
```diff
@@ -1076,122 +965,110 @@ string FirstNonEmpty(params string?[] candidates)
...
             }
 
-            return null;
-        }
-
-        private async Task FinalizeProcessedDownloadAsync(DownloadProcessingJob job, IDownloadService downloadService, string destinationPath)
-        {
-            await downloadService.ProcessCompletedDownloadAsync(job.DownloadId, destinationPath);
-            job.AddLogEntry($"Updated download record with final path: {destinationPath}");
+            await downloadService.ProcessCompletedDownloadAsync(job.DownloadId, job.SourcePath);
```
therobbiedavis, line 1066: Can you help me understand? I think this is accidentally throwing away the completion path we just calculated above. For directory batches we set `job.DestinationPath` from `completionPath`, but this calls the next import stage with `job.SourcePath` instead. If the files were moved, the source may already be empty; if they were copied, we may re-import from the original location. Should this pass `job.DestinationPath` here, or keep the destination as an explicit parameter?

listenarr.api/Services/DownloadProcessingBackgroundService.cs
```diff
@@ -1076,122 +965,110 @@ string FirstNonEmpty(params string?[] candidates)
...
 
+                // Apply remote path mapping
+                var translationTasks = allowedFiles
+                    .Select(path => remotePathMappingService.TranslatePathAsync(download.DownloadClientId, path));
+                    
+                var tranlatedAllowedFiles = await Task.WhenAll(translationTasks);
+
                 var filteredFiles = importableFiles
-                    .Where(path => allowedFiles.Contains(FileUtils.NormalizeStoredPath(path)))
+                    .Where(tranlatedAllowedFiles.Contains)
```
therobbiedavis, line 1032: This match feels a bit brittle now. The client paths are normalized and translated, but the local candidates are compared as-is, and `Contains` is case-sensitive, right? So a small casing/separator/normalization difference could make this return zero files even though the files are present. Could we normalize both sides and compare with an ordinal-ignore-case set?

listenarr.api/Services/RemotePathMappingService.cs
```diff
@@ -143,17 +142,17 @@ public async Task<string> TranslatePathAsync(string downloadClientId, string rem
...
+        remotePath = FileUtils.NormalizeStoredPath(remotePath);
+        var mappings = await GetPathMappingByClientIdAsync(downloadClientId);
 
         foreach (var mapping in mappings)
         {
-            var normalizedMappingPath = NormalizePath(mapping.RemotePath);
+            var normalizedMappingPath = FileUtils.NormalizeStoredPath(mapping.RemotePath);
 
-            if (normalizedRemotePath.StartsWith(normalizedMappingPath, StringComparison.OrdinalIgnoreCase))
+            if (remotePath.StartsWith(normalizedMappingPath, StringComparison.OrdinalIgnoreCase))
```
therobbiedavis: This prefix check can also match sibling paths. For example, a mapping for `/downloads/book` would match `/downloads/book2`.

listenarr.api/Services/FfmpegService.cs
```diff
@@ -523,6 +520,10 @@ private static bool TryBuildPathUnderRoot(string rootPath, string entryPath, out
 
         private string? GetDownloadUrlForPlatform()
         {
+            if (RuntimeInformation.OSArchitecture == Architecture.Arm64)
```
therobbiedavis: Nice catch adding ARM64 support btw. Currently we only ship ARM64 in the docker image (maybe we can add the executables easily), so this probably works for the current release path. Still, this check is `Architecture.Arm64` rather than `Linux && Arm64`, so any future/source-run macOS or Windows ARM64 runtime would also pick the Linux tarball. Could we scope this branch to Linux ARM64 so the platform logic stays accurate?

tests/Listenarr.Api.Tests/DownloadProcessingTests/DownloadProcessing_FileMissingRetryTests.cs
```diff
@@ -240,46 +230,25 @@ public async Task ProcessMoveOrCopy_DirectorySource_FinalizesOnceAfterCompanionF
...
-            downloadServiceMock.Verify(d => d.ProcessCompletedDownloadAsync(dl.Id, expectedTxtDest), Times.Never);
-
-            try { Directory.Delete(sourceDir, true); }
-            catch (IOException) { /* Best-effort cleanup for temp test directories. */ }
-            catch (UnauthorizedAccessException) { /* Best-effort cleanup for temp test directories. */ }
-
-            try { Directory.Delete(destRoot, true); }
-            catch (IOException) { /* Best-effort cleanup for temp test directories. */ }
-            catch (UnauthorizedAccessException) { /* Best-effort cleanup for temp test directories. */ }
+            downloadServiceMock.Verify(d => d.ProcessCompletedDownloadAsync(download.Id, sourceDir), Times.Once);
```
therobbiedavis, line 216: Do we want to assert the destination/completion path instead in the even that there is a regression?

listenarr.api/Services/FfmpegService.cs
```diff
@@ -666,6 +667,234 @@ private static bool TryBuildPathUnderRoot(string rootPath, string entryPath, out
...
+                startInfo.ArgumentList.Add("-show_format");
+                startInfo.ArgumentList.Add("-show_streams");
+                startInfo.ArgumentList.Add(filePath);
+
+                if (_processRunner == null)
+                {
+                    throw new FfmpegException($"IProcessRunner is not available; cannot run ffprobe for {sanitizedFilePath}");
+                }
+
+                var pr = await _processRunner.RunAsync(startInfo, 10000);
```
therobbiedavis, line 701: Should we treat a non-zero ffprobe exit code as a failure before trying to parse stdout?

therobbiedavis (changes requested):
Looking good! Just some small requests!

listenarr.api/Services/DownloadProcessingBackgroundService.cs
T4g1: For directory import, the calculated destination was always a single file thus failing the import in the following step. It was also doing nothing as it checks for the destination to exist before doing anything. That very same destination we create afterwards.

Regarding single files import though, im not really sure of the impact. Both the pre-computation that leads to this line and the methods called afterwards are trying to move/copy the file into destination. Which is very wrong imho

If the processing of single file is already completed, we should not call this particular ProcessCompletedDownloadAsync

We should not use those path as all-in-one information, depending on the source, destinationPath, finalPath, sourcePath means very different things, this will be part of refactoring I plan on doing in #535 though

DownloadProcessingBackgroundService should be limited to monitoring ongoing download and triggering lifecycle events. It should not actually process the downloads but DownloadMonitorService already does that (and also has logic to move things around), then we have CompleteDownloadHandlingService and CompleteDownloadProcessor which both have processing of download logics :x (and a bunch of other things which also does strange things around the files)

listenarr.api/Services/DownloadProcessingBackgroundService.cs
T4g1: I do not think this is required as importable files comes directly from the framework (Directory.EnumerateFiles) but I'll see about adding it

tests/Listenarr.Api.Tests/DownloadProcessingTests/DownloadProcessing_FileMissingRetryTests.cs
T4g1: I dont get it, the files are already checked just before that line, what do you mean ?

T4g1: Added pre-condition asserts!

listenarr.api/Services/DownloadProcessingBackgroundService.cs
therobbiedavis: That makes sense, and I agree. We should split the responsibilities.

My concern is that ProcessMoveOrCopyJobAsync copies/moves the files, then FinalizeProcessedDownloadAsync calls ProcessCompletedDownloadAsync. So I believe then, in copy mode,  this can cause a second import pass from the source directory, and in move mode the source may already be empty. 

I’m fine with the larger cleanup happening in #535, but I am just worried about the larger functional implications until we get around to #535.

T4g1: For single file, `ProcessMoveOrCopyJobAsync` simply calls `ProcessFileWithEnhancedLogicAsync` which does it's thing then calls `FinalizeProcessedDownloadAsync` where we finaly call `ProcessCompletedDownloadAsync` with the job informations we want.

The thing is: `ProcessFileWithEnhancedLogicAsync` requires settings.OutputPath to be set in order to do anything. That setting is, by default, empty and does nothing. Which is why, I believe, the whole process here does nothing at all except calling `ProcessCompletedDownloadAsync` on a given sourcePath using jobs

I agree that this merge is risky though, this whole thing is very confusing. It does lay ground work for #534 and #535 so I could work on those here if you rather not merge this yet

Or we can merge canary and not release it so we have time to test it

therobbiedavis (approved):

PR #524, opened 2026-04-19
Upgrade .NET 10 / Node 24 + clean architecture migration

therobbiedavis:
# Summary

This one's been a while coming. The API layer was directly importing EF Core types so controllers and services had `ListenArrDbContext` injected all over the place, making it impossible to unit test without spinning up a real database and tightly coupling business logic to persistence details. This PR fixes that, and takes the opportunity to upgrade the runtime stack at the same time.

## Added

- 17 new application-layer repository interfaces under `listenarr.application/Repositories/` (`IDownloadRepository`, `IDownloadHistoryRepository`, `IHistoryRepository`, `IIndexerRepository`, `IUserRepository`, `IUserSessionRepository`, `IApplicationSettingsRepository`, `IApiConfigurationRepository`, `IDownloadClientConfigurationRepository`, `IRemotePathMappingRepository`, `IAudiobookFileRepository`, `IMoveJobRepository`, `IMonitoredAuthorRepository`, `IMonitoredSeriesRepository`, `IProcessExecutionLogRepository`, `IDownloadProcessingJobRepository`, `IRootFolderRepository`)
- 13 new EF Core repository implementations under `listenarr.infrastructure/Repositories/` (`EfHistoryRepository`, `EfIndexerRepository`, `EfUserRepository`, `EfUserSessionRepository`, `EfApplicationSettingsRepository`, `EfApiConfigurationRepository`, `EfDownloadClientConfigurationRepository`, `EfRemotePathMappingRepository`, `EfAudiobookFileRepository`, `EfMoveJobRepository`, `EfMonitoredAuthorRepository`, `EfMonitoredSeriesRepository`, `EfProcessExecutionLogRepository`)
- `IDatabaseConnectionProvider` / `EfDatabaseConnectionProvider` for exposing raw connection info without leaking EF types
- `QueueStats` domain model for download processing queue metrics
- CI layering enforcement: `run-tests.yml` now fails the build if any file outside `Program.cs` in the API layer imports from `listenarr.infrastructure`, or if any `async void` appears in production code
- CI now installs Node 24, runs frontend build, lint, type-check, and tests as part of the test job
- Targeted download query methods on `IDownloadRepository`: `GetCompletionCandidatesAsync`, `GetActiveForMonitoringAsync`, `GetRecentAsync`, `GetActiveAudiobookIdsAsync` — replaces hot-path `GetAllAsync()` + in-memory LINQ in several background services

## Changed

- Upgraded .NET 8 → .NET 10 (`global.json`, `Directory.Build.props`, both Dockerfiles)
- Upgraded gosu builder from `golang:1.24-alpine` → `golang:1.26-alpine` to pull in a patched Go stdlib
- Upgraded Node 20 → 24 and npm 10.9.8 → 11.12.1 in `Dockerfile.runtime`
- `EfDownloadProcessingJobRepository` and `EfRootFolderRepository` moved from `listenarr.api/Repositories/` to `listenarr.infrastructure/Repositories/` where they belong
- All controllers and services that previously injected `ListenArrDbContext` or `IDbContextFactory<ListenArrDbContext>` now inject the appropriate application-layer interface(s)
- `InfrastructureServiceRegistrationExtensions.AddListenarrInfrastructure()` is now the single DI registration surface for all repositories; `Program.cs` is the only API-layer file that references infrastructure
- Replaced `AddDbContext + manual SimpleDbContextFactory singleton` with a single `AddDbContextFactory<ListenArrDbContext>(ServiceLifetime.Singleton)` call, which registers both the factory and a scoped context automatically
- `run-tests.yml` adds concurrency group (cancel-in-progress on re-push), switches from matrix to env vars, upgrades `setup-dotnet` to v4, and adds a publish step for linux-x64
- `vite.config.ts` adds explicit type annotation for the visualizer plugin to satisfy TypeScript strict checking
- Frontend test stubs updated to use the new repository interfaces throughout

## Fixed

- Pinned `picomatch@4.0.4` and `brace-expansion@2.0.3` inside the npm install in `Dockerfile.runtime` to resolve CVEs present in the versions bundled with npm 11
- `TestDownloadRepository.ToQueueTrackedDownloadProjection` was missing `AudiobookId` and `Language` fields present in the EF implementation — queue display tests would pass even if downstream code relying on those fields was broken
- `TestDownloadProcessingJobRepository.GetDueRetryJobsAsync` was returning results in insertion order instead of `Priority DESC, NextRetryAt ASC`, diverging from the EF implementation
- `TestDownloadProcessingJobRepository.GetStatsAsync` was returning an empty `QueueStats` object instead of computing real counts — any test asserting on queue metrics would always pass
- `GetDueRetryJobsAsync` lacked a `CancellationToken` parameter on both the interface and implementation, inconsistent with every other async method in the same repository

## Removed

- `.github/workflows/ci.yml` — old CI workflow that targeted `main`/`master` branches and was never triggered; superseded by `run-tests.yml`
- `listenarr.api/Extensions/ServiceProviderExtensions.cs` — no longer needed after DI wiring moved to infrastructure
- `listenarr.api/Repositories/EfDownloadProcessingJobRepository.cs`, `EfRootFolderRepository.cs` — moved to infrastructure
- `listenarr.api/Repositories/IDownloadProcessingJobRepository.cs`, `IDownloadRepository.cs`, `IRootFolderRepository.cs` — moved to application layer
- `SimpleDbContextFactory` — made redundant by `AddDbContextFactory`


listenarr.infrastructure/Services/AudiobookRepository.cs
```diff
@@ -386,6 +407,36 @@ private static string NormalizeAuthorName(string? value)
...
+            var allowedColumns = new HashSet<string>(columns, StringComparer.Ordinal);
+
+            foreach (var col in columns)
+            {
+                if (!allowedColumns.Contains(col) || !System.Text.RegularExpressions.Regex.IsMatch(col, @"^[A-Za-z_][A-Za-z0-9_]*$"))
+                    continue;
+
+                var sql = $"UPDATE Audiobooks SET {col} = json_array(json_extract({col}, '$')) WHERE {col} IS NOT NULL AND json_valid({col})=1 AND json_type({col}) NOT IN ('array','object')";
+                await _db.Database.ExecuteSqlRawAsync(sql, ct);
+            }
```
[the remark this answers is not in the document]
therobbiedavis: Fixed — removed the redundant allowedColumns HashSet (it was always a superset of columns, so the .Contains() check was dead code) and moved the regex guard into a .Where() on the array itself.

listenarr.api/Services/CompletedDownloadHandlingService.cs
```diff
@@ -127,18 +124,19 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
             {
                 // Find all downloads that are Completed/ImportPending but not yet imported (FinalPath still empty)
-                var completedDownloads = await dbContext.Downloads
+                var allDownloads = await downloadRepo.GetAllAsync();
+                var completedDownloads = allDownloads
                     .Where(d =>
                         (d.Status == DownloadStatus.Completed || d.Status == DownloadStatus.ImportPending) &&
                         string.IsNullOrEmpty(d.FinalPath))
-                    .ToListAsync(cancellationToken);
+                    .ToList();
```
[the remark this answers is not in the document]
therobbiedavis: Fixed — replaced GetAllAsync() + in-memory filter with GetCompletionCandidatesAsync(500), which issues a server-side query scoped to Completed/ImportPending/Processing rows. The in-memory filter is now just the FinalPath check on a small result set.

listenarr.api/Services/CompletedDownloadHandlingService.cs
```diff
@@ -312,13 +310,14 @@ private async Task ProcessCompletedDownloadsAsync(CancellationToken cancellation
...
+            IDownloadRepository downloadRepo,
             IServiceScope scope,
             CancellationToken cancellationToken)
         {
-            var movedDownloads = await dbContext.Downloads
+            var allDownloads = await downloadRepo.GetAllAsync();
+            var movedDownloads = allDownloads
                 .Where(d => d.Status == DownloadStatus.Moved && !string.IsNullOrEmpty(d.DownloadClientId))
-                .ToListAsync(cancellationToken);
+                .ToList();
```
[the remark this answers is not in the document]
therobbiedavis: Fixed — replaced GetAllAsync() + in-memory filter with GetActiveForMonitoringAsync(), which already has a server-side predicate that covers the Moved+DownloadClientId case. The in-memory .Where() then narrows to just Moved rows from that smaller set.

listenarr.api/Services/MoveQueueService.cs
```diff
@@ -76,8 +75,8 @@ public bool TryGetJob(Guid id, out MoveJob? job)
             try
             {
                 using var scope = _scopeFactory.CreateScope();
-                var db = scope.ServiceProvider.GetRequiredService<ListenArrDbContext>();
-                job = db.MoveJobs.FirstOrDefault(j => j.Id == id);
+                var moveJobRepo = scope.ServiceProvider.GetRequiredService<IMoveJobRepository>();
+                job = moveJobRepo.GetByIdAsync(id).GetAwaiter().GetResult();
                 if (job != null) _jobs[id] = job;
                 return job != null;
```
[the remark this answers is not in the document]
therobbiedavis: Acknowledged. TryGetJob and UpdateJobStatus have always been synchronous call sites — the original code used EF's synchronous FirstOrDefault() directly. In ASP.NET Core there's no SynchronizationContext on thread pool threads so the deadlock risk is lower than in classic frameworks, but the thread pool saturation concern is real. Tracked as follow-up work to make both methods async and update callers.

listenarr.api/Services/MoveQueueService.cs
```diff
@@ -100,15 +99,14 @@ public void UpdateJobStatus(Guid id, string status, string? error = null)
...
+                var dbJob = moveJobRepo.GetByIdAsync(id).GetAwaiter().GetResult();
                 if (dbJob != null)
                 {
                     dbJob.Status = status;
                     dbJob.Error = error;
                     dbJob.UpdatedAt = DateTime.UtcNow;
-                    db.MoveJobs.Update(dbJob);
-                    db.SaveChanges();
+                    moveJobRepo.UpdateAsync(dbJob).GetAwaiter().GetResult();
                 }
```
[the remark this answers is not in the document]
therobbiedavis: Same as TryGetJob above — both methods will need to go async together. Tracked as follow-up.

listenarr.application/Services/IAudiobookRepository.cs
```diff
@@ -10,6 +11,9 @@ namespace Listenarr.Api.Services
     public interface IAudiobookRepository
     {
         Task<List<Audiobook>> GetAllAsync();
+        Task<List<Audiobook>> GetByIdsWithFilesAsync(IEnumerable<int> ids, CancellationToken ct = default);
+        Task<List<Audiobook>> GetMonitoredAudiobooksForSearchAsync(DateTime cutoff, CancellationToken ct = default);
+        Task NormalizeJsonColumnsAsync(CancellationToken ct = default);
```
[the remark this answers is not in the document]
therobbiedavis: The file intentionally keeps the Listenarr.Api.Services namespace to avoid a cross-codebase churn wave in this PR — there's a comment in the file explaining this. Namespace correction will be its own PR once the dust settles.

listenarr.api/Services/UnmatchedScanBackgroundService.cs
```diff
@@ -112,23 +112,22 @@ await _hubContext.Clients.All.SendAsync(
...
-                .ToListAsync(ct);
+            var trackedFromFiles = await fileRepo.GetAllFilePathsAsync(ct);
 
-            var trackedFromAudiobooks = await db.Audiobooks
+            var allAudiobooks = await audiobookRepo.GetAllAsync();
+            var trackedFromAudiobooks = allAudiobooks
                 .Where(a => a.FilePath != null)
                 .Select(a => a.FilePath!)
-                .ToListAsync(ct);
+                .ToList();
```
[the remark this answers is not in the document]
therobbiedavis: Valid — GetAllAsync() on AudiobookRepository eagerly loads Files, ExternalIdentifiers, and SeriesMemberships so the full-table scan is heavier than it looks. Tracked as follow-up to add GetLegacyFilePathsAsync() that projects just FilePath values.

listenarr.api/Services/ConfigurationService.cs
```diff
@@ -243,19 +182,8 @@ public async Task SaveApplicationSettingsAsync(ApplicationSettings settings)
...
-                // server-side trigger matching. Detect and decode that case here.
-                //
-                // We keep this defensive normalization server-side for robustness,
-                // but the real fix would be to avoid producing double-encoded JSON
-                // from the frontend. This helper centralizes the detection/decoding
-                // logic so it's consistently applied to both the global triggers
-                // and per-webhook trigger lists.
                 try
                 {
                     settings.EnabledNotificationTriggers = NormalizeTriggerList(settings.EnabledNotificationTriggers) ?? new List<string>();
```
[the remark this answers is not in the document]
therobbiedavis: Fixed — moved the existing-settings fetch to before the normalization block so the null-check runs while EnabledNotificationTriggers is still null (not yet coerced to []). Added the same if-null-preserve pattern used for Webhooks. Partial saves that omit the field now fall back to the stored value instead of silently clearing all triggers.

listenarr.api/Services/RootFolderService.cs
```diff
@@ -99,121 +73,56 @@ public async Task<RootFolder> UpdateAsync(RootFolder root, bool moveFiles = fals
...
 
-            // Update root folder and audiobooks in the same context to ensure transaction consistency
-            using (var ctx = await _dbFactory.CreateDbContextAsync())
+            List<(int audiobookId, string original, string target)> moves = new();
+            if (!string.Equals(oldPath, newPath, StringComparison.OrdinalIgnoreCase))
             {
-                // Re-load the entity in this context so it's properly tracked
-                var trackedRoot = await ctx.RootFolders.FindAsync(root.Id);
-                if (trackedRoot == null) throw new KeyNotFoundException("Root folder not found");
+                moves = await _repo.MigrateAudiobookPathsAsync(oldPath, newPath);
```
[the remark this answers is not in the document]
therobbiedavis: Valid concern — MigrateAudiobookPathsAsync and the root folder UpdateAsync are two separate commits. Wrapping them in a single unit of work requires either a shared DbContext across both repository calls or a raw transaction scope. Tracked as a follow-up; the window for a split-state failure is small in practice (the root folder update is the second, lighter write) but it is a real consistency gap.

tests/Listenarr.Api.Tests/TestServiceFactory.cs
```diff
@@ -70,30 +70,45 @@ public static ServiceProvider BuildServiceProvider(Action<IServiceCollection>? c
...
+                if (db != null)
+                {
+                    return new Listenarr.Infrastructure.Repositories.EfIndexerRepository(db);
+                }
+                var mock = new Mock<Listenarr.Application.Repositories.IIndexerRepository>();
+                mock.Setup(r => r.GetAllAsync(It.IsAny<System.Threading.CancellationToken>())).ReturnsAsync(new System.Collections.Generic.List<Listenarr.Domain.Models.Indexer>());
+                mock.Setup(r => r.GetEnabledAsync(It.IsAny<bool>(), It.IsAny<System.Threading.CancellationToken>())).ReturnsAsync(new System.Collections.Generic.List<Listenarr.Domain.Models.Indexer>());
+                mock.Setup(r => r.GetByIdAsync(It.IsAny<int>(), It.IsAny<System.Threading.CancellationToken>())).ReturnsAsync((Listenarr.Domain.Models.Indexer?)null);
+                return mock.Object;
+            });
```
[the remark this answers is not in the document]
therobbiedavis: Fixed — always returns a mock now. The DbContext path was removed; resolving a scoped DbContext into a singleton factory violates DI lifetime rules and would throw with validateScopes enabled.

package.json
```diff
@@ -10,7 +10,7 @@
     "pretest": "npm run version:sync",
     "dev": "concurrently \"npm run dev:api\" \"npm run dev:web\" --names \"API,WEB\" --prefix-colors \"blue,green\"",
     "dev:api": "cd listenarr.api && dotnet run",
-    "dev:web": "cd fe && npm run dev",
+    "dev:web": "wait-on http-get://localhost:4545/api/v1/system/health -t 120000 -i 1000 --httpTimeout 3000 && cd fe && npm run dev",
```
T4g1: This should probably not be here but could be better used inside a script to run both dev:api and de:web though

listenarr.api/Controllers/DownloadsController.cs
```diff
@@ -13,14 +12,14 @@ namespace Listenarr.Api.Controllers;
 [Tags("Downloads")]
 public class DownloadsController : ControllerBase
 {
-    private readonly ListenArrDbContext _dbContext;
+    private readonly IDownloadRepository _downloadRepo;
```
T4g1: Concerning naming: When do we want to abbreviate or not ?
Why not use _downloadRepository ?

listenarr.api/Controllers/HistoryController.cs
```diff
@@ -31,12 +29,12 @@ namespace Listenarr.Api.Controllers
     [Tags("History")]
     public class HistoryController : ControllerBase
     {
-        private readonly ListenArrDbContext _dbContext;
+        private readonly Listenarr.Application.Repositories.IHistoryRepository _history;
```
T4g1: The Listenarr.Application.Repositories is useless given the using that was added

listenarr.api/Controllers/HistoryController.cs
```diff
@@ -31,12 +29,12 @@ namespace Listenarr.Api.Controllers
...
         private readonly ILogger<HistoryController> _logger;
 
-        public HistoryController(ListenArrDbContext dbContext, ILogger<HistoryController> logger)
+        public HistoryController(Listenarr.Application.Repositories.IHistoryRepository history, ILogger<HistoryController> logger)
```
T4g1: Same here

listenarr.api/Services/Scoring/CompositeScorer.cs
```diff
@@ -1,5 +1,4 @@
-using Listenarr.Domain.Models; // SearchResult, QualityProfile
-using Listenarr.Infrastructure.Models; // Indexer
+using Listenarr.Domain.Models; // SearchResult, QualityProfile, Indexer
```
T4g1: We probably do not want such comments as they will quickly be outdated and thus misleading

listenarr.application/Repositories/IAudiobookRepository.cs
T4g1, line 1: Why is this not in Repositories/ directory ?

listenarr.infrastructure/Repositories/AudiobookRepository.cs
T4g1, line 1: Should be in repositories too

listenarr.application/Services/IDatabaseConnectionProvider.cs
T4g1, line 1: This is confusing, as i understand the provider in such architecture is not the same thing as this, maybe rename to IDatabaseConnectionService instead ?

listenarr.infrastructure/Extensions/InfrastructureServiceRegistrationExtensions.cs
```diff
@@ -10,12 +13,53 @@ namespace Listenarr.Infrastructure.Extensions
     /// </summary>
     public static class InfrastructureServiceRegistrationExtensions
     {
-        public static IServiceCollection AddListenarrInfrastructure(this IServiceCollection services)
+        /// <summary>
+        /// Registers all infrastructure services. When <paramref name="sqliteDbPath"/> is provided
+        /// the DbContext, DbContextOptions, and IDbContextFactory are also registered (production
+        /// and integration-test paths). Omit the parameter only when the caller has already
+        /// registered a DbContext (e.g. unit tests using UseInMemoryDatabase).
```
T4g1: Testing specifity should not drive the implementation. 
This method should get the proper DbContextOptionsBuilder to use (defining sqlite or in memory) instead of the sqlPath
Then, testing like DependencyInjectionTests.cs should give the DbContextOptionsBuilder to AddListenarrInfrastructure instead fo adding them to the service themselves (and simplify that comment)

tests/Listenarr.Api.Tests/AudioFileService_UpdateAudiobookFieldsTests.cs
```diff
@@ -28,6 +28,9 @@ public async Task EnsureAudiobookFileAsync_PopulatesAudiobookFilePathAndSize()
             // Build service provider with required services (including a mock metadata service)
             var services = new ServiceCollection();
             services.AddSingleton<ListenArrDbContext>(db);
+            services.AddSingleton<Listenarr.Application.Repositories.IAudiobookFileRepository>(_ => new Listenarr.Infrastructure.Repositories.EfAudiobookFileRepository(db));
```
T4g1: General remark about the full path: Listenarr.Application.Repositories
This gets complicated to maintain once files are moved or changed from namespace to sub namespace ect.
We should probably rely more on good "using" practices instead

tests/Listenarr.Api.Tests/TestServiceFactory.cs
```diff
@@ -70,30 +70,41 @@ public static ServiceProvider BuildServiceProvider(Action<IServiceCollection>? c
...
             // ListenArrDbContext if present, otherwise fall back to an in-memory test repo.
-            services.AddSingleton<Listenarr.Api.Repositories.IDownloadRepository>(sp =>
+            services.AddSingleton<Listenarr.Application.Repositories.IDownloadRepository>(sp =>
             {
                 var dbFactory = sp.GetService<IDbContextFactory<ListenArrDbContext>>();
                 if (dbFactory != null)
                 {
-                    var logger = sp.GetRequiredService<ILogger<Listenarr.Api.Repositories.EfDownloadRepository>>();
-                    return new Listenarr.Api.Repositories.EfDownloadRepository(dbFactory, logger);
+                    var logger = sp.GetRequiredService<ILogger<Listenarr.Infrastructure.Repositories.EfDownloadRepository>>();
```
T4g1: Random example of why we dont want those fully qualified namespace probably: Listenarr.Api.Repositories to Listenarr.Infrastructure.Repositories

T4g1 (commented):
I feel like Sqlite and In memory for test could benefit from being their own provider but looking at the implementation it would mean defining their own repositories too

listenarr.infrastructure/Services/EfDatabaseConnectionProvider.cs
T4g1, line 1: Service instead of Provider ?

listenarr.infrastructure/Services/EfDatabaseConnectionProvider.cs
```diff
@@ -0,0 +1,25 @@
...
+        public EfDatabaseConnectionProvider(ListenArrDbContext dbContext)
+        {
+            _dbContext = dbContext;
+        }
+
+        public async Task<DbConnection> GetOpenConnectionAsync()
+        {
+            var conn = _dbContext.Database.GetDbConnection();
+            if (conn.State != System.Data.ConnectionState.Open)
+                await conn.OpenAsync();
```
T4g1: Never closed

Make this IDisposable and close it there maybe ? Not sure this in an issue though

CONTRIBUTING.md
```diff
@@ -147,10 +147,41 @@ This project follows a layered pattern: domain models in `listenarr.domain`, EF
...
+PR ──feature──► canary (alpha)
+PR ──feature──► canary
+PR ──feature──► canary
+                canary ──code freeze──► beta (candidate)
+PR ──feature──► canary   (development continues)
+PR ──fix──────► beta     (org members stabilise the candidate)
+PR ──fix──────► beta
+                beta ──────────────────► main (stable release)
+                beta ──rebase──────────► canary (fixes synced forward)
+```
```
T4g1, line 174: I believe you can use mermaid code in markdown like this:
```mermaid
sequenceDiagram
    participant main as main (stable)
    participant testing as beta (candidate)
    participant develop as canary (alpha)
    participant PR
    activate develop
    PR ->> develop: Feature
    PR ->> develop: Feature
    PR ->> develop: Feature
    develop ->> testing: Code freeze 
    activate testing
    PR ->> develop: Feature
    PR ->> testing: Fix
    PR ->> testing: Fix
    PR ->> develop: Feature
    testing ->> main: Release
    testing ->> develop: Rebase
    deactivate testing
    deactivate develop
```

CONTRIBUTING.md
```diff
@@ -147,10 +147,41 @@ This project follows a layered pattern: domain models in `listenarr.domain`, EF
...
+7. PRs to `main` will be closed without review.
+
 ### Pull Request Guidelines
 
 **Branch naming:**
-- Create PRs from feature branches, not from `develop` in your fork
+- Create PRs from feature branches, not from `beta` or `canary` in your fork
 - Use meaningful branch names that describe what is being added/fixed
 
 Good examples:
```
T4g1, line 198: imho: Using issue numbers in the branch name can be usefull too and tend to enforce having an issue opened and previously debated/discussed to base the PR on (to not get random unrelated PRs)

We could enforce it or strongly recommend it (or don't care too :p)

See https://deepsource.com/blog/git-branch-naming-conventions for example

Also the feature/ bugfix/ enhancement/ is commonly used by im still waiting to find a good use case for those, i would recommend ditching them if we dont have one because if we enforce issue numbering too, it will be easier to browse branches by issue number

CONTRIBUTING.md
```diff
@@ -147,10 +147,41 @@ This project follows a layered pattern: domain models in `listenarr.domain`, EF
...
+4. Org members open **fix PRs** targeting `beta` to stabilise the release candidate.
+5. Once stable, `beta` is merged into `main` and a release tag is created.
+6. After the release, `beta` is **rebased back into `canary`** so all fixes are carried forward.
+7. PRs to `main` will be closed without review.
+
 ### Pull Request Guidelines
 
 **Branch naming:**
-- Create PRs from feature branches, not from `develop` in your fork
+- Create PRs from feature branches, not from `beta` or `canary` in your fork
```
T4g1: This is not related to branch naming is it ? Maybe specify to start feature branch from latest canary too

README.md
```diff
@@ -91,7 +91,8 @@ docker run -d \
 - ~~`latest` / `stable`~~ - Latest stable release (available soon)
 - `canary` - Latest canary build (pre-release)
 - `canary-X.Y.Z` - Specific canary version
-- `nightly-X.Y.Z` - Specific nightly version
+- `beta` - Latest beta build (pre-release)
```
T4g1, line 94: release candidate instead ?

README.md
```diff
@@ -290,16 +291,24 @@ This means you can run the helper once and enter your public domain (for example
...
 
+**Contribution branching model:**
+- Feature PRs (all contributors) always target `canary`.
+- When ready for a release candidate, `canary` is code-frozen and merged into `beta` by an org member.
+- Feature development continues into `canary` during the beta window.
+- Org members open fix PRs against `beta` to stabilise the release candidate.
+- `beta` is merged into `main` for the stable release, then rebased back into `canary` to carry fixes forward.
+- PRs directly to `main` will be closed without review.
+
 Version numbers are automatically incremented:
```
T4g1, line 308: I'm not sure we should automate version numbering, usualy, patch is incremented on fixe/minor change, minor is incremented on minor/big change and major on major/outstanding changes. 
Here, any public release would mean a major version increment which seems a bit extreme

README.md
```diff
@@ -465,8 +474,9 @@ docker build -f listenarr.api/Dockerfile.runtime -t <your-image> listenarr.api/p
 ### Version Management
 
 Application versions are managed in `listenarr.api/Listenarr.Api.csproj` with a `<Version>` element. CI automatically bumps versions on builds:
```
T4g1, line 476: Im not so sure it's a good idea, why not base the workflow on manual tagging instead to keep control on it ?

.github/workflows/beta.yml
```diff
@@ -11,10 +11,10 @@ permissions:
 
 on:
```
T4g1, line 12: We probably want to trigger release/build on manual tag creation instead, what do you think ?

T4g1 (commented):
I cant really review the workflow themselves because i got no idea how this works but regarding the process, I'm not convinced by automatic version numbering and automatic release.

I would recommend a manual tagging approach that then triggers the workflow if possible

Also, the changelog updates could probably be delayed until we merge to either beta or main (its easier for contributors too that way as they dont have to be involved in the software lifecycle and can focus on the feature): We would simply have to look at the git log or PR history to know what we should include in it or not

listenarr.application/Services/IDatabaseConnectionProvider.cs
therobbiedavis: I'd keep the name as-is. Provider fits because the interface just hands you an open connection, nothing more. Calling it a Service would be more confusing, I think, since that implies business logic, but this is purely plumbing.

T4g1: I'd argue that it should not be in a Service directory then

README.md
therobbiedavis: I prefer canary, beta, release terminology.

tests/Listenarr.Api.Tests/AudioFileService_UpdateAudiobookFieldsTests.cs
therobbiedavis: This should be fixed across the repo.

.github/workflows/validate-pr-labels.yml
```diff
@@ -0,0 +1,38 @@
...
+                `Current labels: ${labels.length ? labels.join(', ') : '(none)'}`
+              );
+            } else if (found.length > 1) {
+              core.setFailed(
+                `PR #${context.payload.pull_request.number} has multiple version labels (${found.join(', ')}). ` +
+                `Only one of ${valid.join(', ')} is allowed.`
+              );
+            } else {
+              core.info(`✅ Version label '${found[0]}' found on PR #${context.payload.pull_request.number}.`);
+            }
```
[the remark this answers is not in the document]
therobbiedavis: Resolved.

.github/workflows/beta.yml
```diff
@@ -106,7 +106,7 @@ jobs:
       - name: Abort obsolete run
         if: steps.inspect_target.outputs.bump_state == 'superseded'
         run: |
-          echo "::error::Current develop version (${{ steps.inspect_target.outputs.BRANCH_VERSION }}) is newer than target version (${{ steps.resolve-version.outputs.VERSION }}). This run is obsolete and will not publish stale artifacts."
+          echo "::error::Current beta version (${{ steps.inspect_target.outputs.BRANCH_VERSION }}) is newer than target version (${{ steps.resolve-version.outputs.VERSION }}). This run is obsolete and will not publish stale artifacts."
           exit 1
 
       - name: Persist bumped version to csproj
```
T4g1: Probably not relevant as only canary does logic on that ?

.github/workflows/canary.yml
T4g1, line 1: General note about those: There's a lot of logic duplication, we can factorize those so only one files contains the logic while others reuse it (see this probably: https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows)

For example:
- Beta should probably only care about how the docker image is called while another workflow could handle all the release/build logic (this can be achieved with [env](https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows) i believe

- Canary should know how to handle version tags and increment project version and then trigger a build like beta does

CONTRIBUTING.md
```diff
@@ -163,11 +208,15 @@ Bad examples:
...
 
 **PR process:**
-1. **Target branch**: Only make pull requests to `canary`, never `main` or `develop`
-   - PRs to `main` and `develop` will be commented on and closed
+1. **Target branch**:
+   - **Feature branches** → `canary` (all contributors)
+   - **Hotfixes** → `beta` (org members only)
+   - PRs to `main` will be commented on and closed
+   - Never open a PR directly from `beta` or `canary` in your fork — always use a dedicated feature branch
+   > **Note for maintainers:** The `Validate PR version label` check only blocks merges if it is registered as a required status check under **Settings → Branches → canary branch protection rule**. Without that, a PR without a version label can still be merged (the build will then fail at the canary workflow step instead).
```
T4g1, line 219: I'm not sure I undestand fully what it implies: Does that means that for each new PR, we have to change the branch protection rule ? If not, can contributors leverage that information in some way ? I believe we can keep the comment in the workflow itself and remove this, what do you think ?

CONTRIBUTING.md
```diff
@@ -187,7 +236,7 @@ Bad examples:
 - [ ] All tests pass
 - [ ] No console errors or warnings
 - [ ] Documentation updated (if needed)
-- [ ] Rebased on latest `canary` branch
+- [ ] Rebased on latest `canary` branch (or `beta` for org hotfixes)
```
T4g1, line 239: Note for later: Tests are typicaly run on Linux in Github, it could be good to add a check to run the test on windows too before validating as it behaves differently regarding folder naming capabilities (ive started tagging that kind of tests so we can "dotnet test --filter "OS=Windows" or something

listenarr.api/Program.cs
```diff
@@ -194,7 +194,7 @@ ex is IOException
     .MinimumLevel.Override("Microsoft.EntityFrameworkCore", LogEventLevel.Warning)
     .MinimumLevel.Override("Microsoft.EntityFrameworkCore.Database.Command", LogEventLevel.Warning)
     // Enable debug logging for Transmission adapter to troubleshoot RPC issues
-    .MinimumLevel.Override("Listenarr.Api.Services.Adapters.TransmissionAdapter", LogEventLevel.Debug)
+    .MinimumLevel.Override("TransmissionAdapter", LogEventLevel.Debug)
```
T4g1: About this: I believe Transmission RPC API is relatively stable, this has a huge impact on the size of logs (one day one my end is like 500Mb of logs), i was thinking about filtering those out, RPC issues can be debugged quite easily with a Postman and the request so storing reply is probably a bit overkill, wdyt ?

tests/Listenarr.Api.Tests/DependencyInjectionTests.cs
```diff
@@ -15,13 +32,11 @@ public void InfrastructureRegistrations_ResolveIAudiobookRepository()
         {
             var services = new ServiceCollection();
 
-            // Register an in-memory DbContext so repository dependencies are satisfied.
-            services.AddDbContext<ListenArrDbContext>(options =>
+            // Register infrastructure implementations (the extension lives in Infrastructure project).
+            // Pass the InMemory provider so the extension wires up IDbContextFactory<ListenArrDbContext>.
+            services.AddListenarrInfrastructure(options =>
```
T4g1, line 37: 💯

T4g1 (commented):
A lot of nice changes, i would just recommend cleaning those workflow to avoid logic duplication but otherwise i can't wait to see this merged 👍

listenarr.api/Program.cs
T4g1: (i do have a lot of unrelated torrents seedings so torrent-get is really huge)

CONTRIBUTING.md
therobbiedavis: No, it just means from here on we need to make sure that a PR has a patch/minor/major label otherwise the build will fail. This is just documentation to explain that.

.editorconfig
```diff
@@ -0,0 +1,45 @@
...
+indent_size = 2
+
+# ─── JSON ─────────────────────────────────────────────────────────────────────
+[*.json]
+indent_style = space
+indent_size = 2
+
+# ─── Markdown ─────────────────────────────────────────────────────────────────
+[*.md]
+trim_trailing_whitespace = false
```
T4g1: Why not true ?

therobbiedavis: No real reason, just default to preserve hard line breaks but we don't use that I don't believe. I changed it.

listenarr.api/Program.cs
therobbiedavis: Removed

CONTRIBUTING.md
therobbiedavis: Agree, for later though

.github/workflows/canary.yml
therobbiedavis: Update this to remove logic duplication

.github/workflows/beta.yml
therobbiedavis: correct, accidental dead code

therobbiedavis:
@T4g1 Good for hopefully a last review here

.github/workflows/canary.yml
T4g1: Nice, it's a lot cleaner like that!

T4g1:
lgtm! Can't wait to see how the rebase for other PR will go 😭

T4g1 (approved):

PR #532, opened 2026-04-22
Add pre-commit linter and pre-push tests

T4g1:
## Summary

Using Husky, this allows to define pre-commit and pre-push hooks to enforce formatting and tests runs when doing commit and push

All of this can be bypassed using `--no-verify` when using the associated git command (so we can add them to Github CI if we want to enforce them more seriously)

Only those files are relevant for the review:
* .husky
* package.json

The rest is a preliminary run of `dotnet format` to align the codebase on the currently defined rules

## Notes
* The `dotnet format` command on commit is a bit time consuming (roughly 19s on my setup)


package.json
```diff
@@ -16,10 +16,16 @@
...
+    "test:backend": "dotnet test",
+    "test:frontend": "cd fe && npm run test:unit",
+    "prepare": "husky"
   },
   "dependencies": {
     "concurrently": "^9.2.1",
     "wait-on": "^8.0.3"
+  },
+  "devDependencies": {
+    "husky": "^9.1.7"
```
therobbiedavis, line 39: you need to commit the regenerated package-lock.

.husky/pre-commit
therobbiedavis, line 1: Can we also add these checks to our test workflow somehow?

therobbiedavis (changes requested):

.husky/pre-commit
T4g1: Yes, it should be possible but I haven't tried anything in that regard yet. If you want and have the time for it, you can take the lead on this to see what's possible

Also, we can configure specific codes the formater should fix, do we want to be more intrusive on those ? Or do we keep defaults as they are right now ? (we could remove unused using or enforce namespace for example)

therobbiedavis (approved):

T4g1:
I cannot review my own PR but I approve your changes! But didn't you talked about enforcing those on the CI workflows too ?

therobbiedavis:
> I cannot review my own PR but I approve your changes! But didn't you talked about enforcing those on the CI workflows too ?

So yes, but I changed my mind to only test for bad formatting so it can't get merged unless it passes the lint. However I forgot to add the backend check.

PR #548, opened 2026-04-26
Testing framework

### LINKED ISSUE #534: [Architecture] Testing framework
### Context

Right now, tests are all located in the same folder, most of them have to define cumbersome setups before doing real testing, those definitions are duplicated everywhere

We need a framework for testing that can abstract the setup required so tests can focus only on defining specific data, manipulation and verification

Mock setups should also be abstracted

T4g1:
## Summary
* Fix #534 
See the test README for most of the explanation

### Added
- Builder pattern for most commonly used objects in tests
- Revamped tests directory structure
- Added commonly defined objects and shortcuts for common repositories

### Changed
- Disabled startup logging for tests
- Simplified environment handling and better isolation of project/configurations

## Testing
- As little change to the actual codebase as possible is done here so existing unit/integration test only

## Next steps
Best efforts:
- Set namespaces to correct values
- Add proper tagging on all tests
- Apply simplified framework to improve test maintainability


tests/Common/BaseTests.cs
```diff
@@ -0,0 +1,74 @@
...
+            _audiobookFileRepository = _provider.GetRequiredService<IAudiobookFileRepository>();
+            _downloadProcessingJobRepository = _provider.GetRequiredService<IDownloadProcessingJobRepository>();
+            _indexerRepository = _provider.GetRequiredService<IIndexerRepository>();
+        }
+
+        public virtual async Task InitializeAsync()
+        {
+        }
+
+        public virtual async Task DisposeAsync()
```
therobbiedavis, line 78: Don't we need to dispose our created stuff here? Otherwise files created from other tests could leak into other tests, right?

tests/Common/TempFileService.cs
```diff
@@ -1,23 +1,35 @@
...
-        public BaseTests()
+        public TempFileService()
         {
             _tempFolder = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString());
             Directory.CreateDirectory(_tempFolder);
+        }
+
+        public virtual Task InitializeAsync()
+        {
+            throw new NotImplementedException();
```
therobbiedavis: This could just be a return since the constructor always sets it up, right?

tests/Common/MockUtils.cs
```diff
@@ -0,0 +1,183 @@
...
+                new Mock<IHttpClientFactory>().Object,
+                new Mock<IAppMetricsService>().Object);
+        }
+
+        public static ServiceProvider CreateServiceProvider(string outputPath = "")
+        {
+            return CreateServiceProvider(new Mock<IImportItemResolutionService>().Object, outputPath);
+        }
+
+        public static ServiceProvider CreateServiceProvider(IImportItemResolutionService importItemResolutionService, string outputPath = "", DownloadClientConfiguration downloadClientConfiguration = null)
```
therobbiedavis: Should be `DownloadClientConfiguration? downloadClientConfiguration = null`

tests/Common/MockUtils.cs
```diff
@@ -0,0 +1,183 @@
...
+
+            // Set job to processing (the outer loop normally does this)
+            job.Status = ProcessingJobStatus.Processing;
+            job.StartedAt = DateTime.UtcNow;
+            await queueService.UpdateJobAsync(job);
+
+            return job;
+        }
+
+        public static TransmissionAdapter CreateTransmissionAdapter(ServiceProvider provider, Mock<ITorrentFileDownloader>? torrentFileDOwnloader = null)
```
therobbiedavis: `torrentFileDOwnloader = null` typo (capital O, should be `torrentFileDownloader = null`

tests/Common/TestUtils.cs
therobbiedavis: Calling this by string seems like it could lead to potential maintainability issues. Maybe we mark these methods as internal so we can use them in tests?

tests/Listenarr.Tests.csproj
therobbiedavis, line 6: eventually we'll need to fix these instead of suppress them.

tests/Builders/DownloadBuilder.cs
```diff
@@ -0,0 +1,112 @@
...
+            return this;
+        }
+
+        public DownloadBuilder WithBlockMessage(string value)
+        {
+            _download.ImportBlockMessages.Add(value);
+            return this;
+        }
+
+        public DownloadBuilder WithImportAttemps(int value)
```
therobbiedavis: Typo `WithImportAttemps` should be `WithImportAttempts`

tests/Builders/DownloadClientConfigurationBuilder.cs
```diff
@@ -0,0 +1,103 @@
...
+        private Dictionary<string, object> _settings = [];
+
+        public DownloadClientConfigurationBuilder()
+        {
+            _downloadClientConfiguration.Id = Guid.NewGuid().ToString();
+            _downloadClientConfiguration.IsEnabled = true;
+            _downloadClientConfiguration.Host = "localhost";
+            _downloadClientConfiguration.Port = 8080;
+            _downloadClientConfiguration.Type = "qbittorrent";
+        }
```
therobbiedavis, line 17: Which is best, having IsEnabled true and Type qbittorrent set by default or setting this to false/"" and explicitly setting with each test? For instance we'll already need to override Type for any other client test.

tests/Builders/IndexerBuilder.cs
```diff
@@ -0,0 +1,97 @@
...
+namespace Listenarr.Tests.Builders
+{
+    public class IndexerBuilder
+    {
+        private readonly Indexer _indexer = new();
+        private Dictionary<string, string> _additionalSettings = [];
+
+        public IndexerBuilder()
+        {
+            _indexer.Id = 1;
```
therobbiedavis: Just need to make sure no test builds two indexers or more otherwise this ID will collide.

tests/Builders/IndexerBuilder.cs
```diff
@@ -0,0 +1,97 @@
...
+            _indexer.Name = "";
+            _indexer.Type = "torrent";
+            _indexer.Implementation = "Custom";
+            _indexer.Url = "";
+            _indexer.ApiKey = "";
+            _indexer.IsEnabled = true;
+            _indexer.EnableInteractiveSearch = true;
```
therobbiedavis, line 22: Some default concern here as DownloadClientConfigurationBuilder()

tests/Builders/IndexerBuilder.cs
```diff
@@ -0,0 +1,97 @@
...
+            return this;
+        }
+
+        public IndexerBuilder Enabled()
+        {
+            _indexer.IsEnabled = true;
+            return this;
+        }
+
+        public IndexerBuilder DIsabled()
```
therobbiedavis: Typo `DIsabled` instead of `Disabled`

tests/Builders/AudiobookBuilder.cs
```diff
@@ -0,0 +1,57 @@
...
+
+namespace Listenarr.Tests.Builders
+{
+    public class AudiobookBuilder
+    {
+        private readonly Audiobook _audiobook = new();
+
+        public AudiobookBuilder()
+        {
+            _audiobook.Id = 1;
```
therobbiedavis: Same collision concern as IndexerBuilder()

tests/README.md
```diff
@@ -0,0 +1,71 @@
...
+```
+* If you overide one or more service this way, you should call `Init()`
+    * You should add data to repository only after you have manually called `Init()` 
+    * If you did not alter any service, this is done for you in the test constructor already
+
+The `Init()` method create the following things:
+* A `ServiceProvider` using the services you injected and default ones added for you. This provider is accessible using `_provider`
+* A reference to the most useful repositories to save you some line of codes:
+    * `IDownloadRepository` under `_downloadRepository`
+    * `IDownloadClientConfigurationRepository` under `_idownloadClientConfigurationRepository`
```
therobbiedavis: Typo `_idownloadClientConfigurationRepository` instead of `_downloadClientConfigurationRepository`

listenarr.api/Services/NotificationPayloadBuilder.cs
```diff
@@ -359,14 +353,14 @@ static string Truncate(string? value, int max)
             {
                 try
                 {
-                    Console.WriteLine($"DEBUG: Attempting to download image for attachment: {absoluteImageUrl}");
+                    //Console.WriteLine($"DEBUG: Attempting to download image for attachment: {absoluteImageUrl}");
```
therobbiedavis: We can just remove these commented lines

listenarr.api/Services/NotificationService.cs
```diff
@@ -269,7 +264,7 @@ async Task HandleFailedResponseAsync(HttpResponseMessage response)
                         apiVersion: ApiVersionPathBuilder.ResolveApiVersion(_httpContextAccessor?.HttpContext, startup?.ApiVersion)
                     );
 
-                    Console.WriteLine($"DEBUG: NotificationService received attachment? {attachment != null}");
+                    //Console.WriteLine($"DEBUG: NotificationService received attachment? {attachment != null}");
```
therobbiedavis: Same here, we can remove this commented line

fe/package.json
```diff
@@ -19,7 +19,8 @@
     "dev:api": "cd ../listenarr.api && dotnet run",
     "build": "run-p type-check \"build-only {@}\" --",
     "preview": "vite preview",
-    "test:unit": "vitest",
+    "test": "vitest",
+    "test:unit": "vitest run --reporter=dot --silent",
```
therobbiedavis, line 23: Have might need to call out in the CONTRIBUTING doc which we expect devs to run. I assume we'll use test:unit for CI, but if a dev runs that would they think everything is good since it suppresses errors?

fe/package.json
therobbiedavis: Looks like this needs to be rebased since we're on 0.3.0 now

therobbiedavis (changes requested):
Some questions and requests

tests/Common/BaseTests.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        public IApplicationSettingsRepository _applicationSettingsRepository;
+        public IDownloadRepository _downloadRepository;
+        public IDownloadClientConfigurationRepository _downloadClientConfigurationRepository;
+        public IRemotePathMappingRepository _remotePathMappingRepository;
+        public IHistoryRepository _historyRepository;
+        public IAudiobookRepository _audiobookRepository;
+        public IAudiobookFileRepository _audiobookFileRepository;
+        public IDownloadProcessingJobRepository _downloadProcessingJobRepository;
+        public IIndexerRepository _indexerRepository;
```
lzinga: These are implementation details of the test base class. They should be protected to prevent tests from accidentally passing these around or external code depending on them. The underscore prefix convention also implies they're private/protected

tests/Builders/AudiobookBuilder.cs
```diff
@@ -0,0 +1,57 @@
...
+            return this;
+        }
+
+        public AudiobookBuilder WithSeries(string value)
+        {
+            _audiobook.Series = value;
+            return this;
+        }
+
+        public AudiobookBuilder PublishedOn(DateOnly value)
```
lzinga: Should this be "WithPublishedOn" to follow the same naming pattern? Minor but slightly inconsistent.

tests/README.md
```diff
@@ -0,0 +1,71 @@
...
+* Test name should be the name of the class you are trying to test followed by `Tests`. Example for `DownloadService`: `DownloadServiceTests`
+* Your test class should be in a path mirroring the main project being tested by it (example: `Adaper` tests are in `Features/Api/Services/Adapters`)
+* Your test class should inherits `BaseTests`. Example for `DownloadService`: `DownloadServiceTests : BaseTests`
+* Your test class should define the following tags:
+```
+[Trait("Name", "DownloadServiceTests")]
+[Trait("Category", "DownloadService")]
+```
+* You can override `InitializeAsync` to define common setup for all test methods in your test class
+* You can override any service for dependancy injection using:
```
lzinga: dependancy -> dependency

tests/README.md
```diff
@@ -0,0 +1,71 @@
...
+```
+_services.AddScoped(...)
+_services.AddSingleton(...)
+_services.AddTransient(...)
+
+# Dont forget to call Init() afterwards
+```
+* If you overide one or more service this way, you should call `Init()`
```
lzinga: overide -> override

tests/Common/TestUtils.cs
lzinga: this builds its own DI container separate from the test's _services right? so IDownloadService and IConfigurationService on lines 44-45 are just bare mocks that return nothing. wouldn't that mean tests using this helper aren't actually hitting the real service logic, they're just checking it doesn't blow up?

lzinga (commented):
Took a quick look through.

tests/Builders/DownloadClientConfigurationBuilder.cs
T4g1: I believe that a test should have to explicit this if it cares about the value. Most tests requires a valid client configuration but few actually cares about what client is configured. It's also the lazy approach and leads to less line of codes (and less line means less bugs hopefully :p )

fe/package.json
T4g1: I does not suppress errors at all, here is the kind of output i get with those settings:
2026-05-04 11_33_04-C__Windows_system32_cmd exe

For the frontend: If a test is erroneous, a red cross will be shown and a report will give details about the tests that failed.
No red cross for the backend but a summary will still give details about failures

tests/Common/BaseTests.cs
T4g1: This is the responsibility of the TempFileService class but I forgot to manually dispose of it (as I created it manually with new) so indeed, files are still there ^^' Note that it make sure to use a GUID behind the scene so no collision si guaranteed but the filesystem was not cleaned after the tests :x

tests/Common/MockUtils.cs
T4g1: Also, those method should be removed

tests/Common/TestUtils.cs
T4g1: The method was not used so I removed it (fyi: it's a method that is already in place in several tests, I agree that we could improve that too)

tests/Builders/AudiobookBuilder.cs
T4g1: Yeah, I was a bit puzzled by this, should we make it more fluent for the developper that stumble upon test cases written with this or consistent in the naming of the builder's method ? The main purpose of using builders here is to help understand faster what the test does and what kind of data it manipulates

Here I will go for WithPublishedDate probably but the question remains

tests/Common/TestUtils.cs
T4g1: Yes, fortunately, this method is no longer used so I removed it. They should indeed use the test's service collection instead

listenarr.api/Program.cs
```diff
@@ -39,87 +33,52 @@
...
-            // (for example a transient IO issue or an unexpected path layout), fall
-            // back to the default builder which will use the running assembly's
-            // base directory. Log to console so developers can see the fallback.
-            Console.WriteLine($"[Listenarr] Warning: failed to use content root '{projectDir}' - falling back to default. Error: {ex.Message}");
-            builder = WebApplication.CreateBuilder(args ?? Array.Empty<string>());
+            Console.WriteLine($"[Listenarr] LISTENARR_CONTENT_ROOT '{contentRootOverride}' does not exist; creating it");
+            Directory.CreateDirectory(contentRootOverride);
         }
+
+        builder.Environment.ContentRootPath = contentRootOverride;
```
therobbiedavis: I think this override needs to happen before `WebApplication.CreateBuilder`. Right now it looks like the builder is created with `AppContext.BaseDirectory` on line 36, and then `ContentRootPath` is mutated here afterward. Some default config/file-provider setup has already happened by then, so `LISTENARR_CONTENT_ROOT` may not fully behave like the real content root. Could we resolve/create the override path first, then pass it into `WebApplicationOptions.ContentRootPath`?

tests/Mocks/FfmpegServiceMock.cs
```diff
@@ -0,0 +1,50 @@
+using Listenarr.Api.Services;
+using Listenarr.Domain.Models;
+using Listenarr.Tests.Builders;
+using Listenarr.Tests.Common;
+
+namespace Listenarr.Tests.Mocks
+{
+    public class FfmpegServiceMock : IFfmpegService
+    {
+        private TempFileService _fileService = new();
```
therobbiedavis, line 10: This creates its own `TempFileService`, but this is never disposed.

therobbiedavis (changes requested):

listenarr.api/Program.cs
```diff
@@ -350,9 +318,9 @@ ex is IOException
 // Add default HTTP client for other services
 builder.Services.AddHttpClient();
 
-// Add named HttpClient for download operations (qBittorrent, Transmission, etc.)
+// Add named HttpClient for sabnzbd
 // Prevents socket exhaustion by reusing connections
-builder.Services.AddHttpClient("DownloadClient")
+builder.Services.AddHttpClient("sabnzbd")
```
therobbiedavis: Shouldn't this be `"DownloadClient"`? SABnzbd is already registered below, and `DownloadMonitorService` still asks for `"DownloadClient"`. Registering the same named client

listenarr.api/Program.cs
```diff
@@ -39,87 +33,56 @@
 using Listenarr.Api.Extensions;
 using Listenarr.Infrastructure.Extensions;
 
-// Check for special CLI helpers before building the web host
-// Set ContentRootPath to a reliable value for local dev, but leave Docker/production unaffected.
-var isDocker = Environment.GetEnvironmentVariable("DOTNET_RUNNING_IN_CONTAINER") == "true";
-var isDev = Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT") == "Development";
+var contentRootPath = AppContext.BaseDirectory;
+var environmentName = "Production";
```
therobbiedavis: Should this respect `ASPNETCORE_ENVIRONMENT` / `DOTNET_ENVIRONMENT` instead of defaulting straight to `"Production"`? As written it looks like local `Development` / `Integration` launch profiles won’t affect `builder.Environment` unless the test-host override kicks in.

tests/Common/BaseTests.cs
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        public virtual async Task InitializeAsync()
+        {
+            await FileService.InitializeAsync();
+        }
+
+        public virtual async Task DisposeAsync()
+        {
+            await FileService.DisposeAsync();
```
therobbiedavis, line 80: Should we dispose the service provider here too?

tests/Builders/DownloadClientConfigurationBuilder.cs
```diff
@@ -0,0 +1,103 @@
...
+
+        public DownloadClientConfigurationBuilder WithoutSsl()
+        {
+            _downloadClientConfiguration.UseSSL = false;
+            return this;
+        }
+
+        public DownloadClientConfigurationBuilder WithSettings(string key, string value)
+        {
+            _downloadClientConfiguration.Settings[key] = value;
```
therobbiedavis: I think this should write to `_settings` instead.

listenarr.api/Services/NotificationService.cs
therobbiedavis: Still leftover

tests/Mocks/ListenarrWebApplicationFactory.cs
```diff
@@ -16,16 +16,18 @@
  * along with this program. If not, see <https://www.gnu.org/licenses/>.
  */
 using Listenarr.Api.Services;
```
therobbiedavis, line 18: Small cleanup concern here, should `base.Dispose(disposing)` run before deleting the temp SQLite files? The test host may still hold the DB connection until the factory is disposed.

therobbiedavis (changes requested):

listenarr.api/Program.cs
T4g1: My bad, I did not saw it was already defined

tests/Common/BaseTests.cs
```diff
@@ -0,0 +1,101 @@
...
+            nameof(_downloadProcessingJobRepository),
+            nameof(_indexerRepository),
+            nameof(_downloadHistoryRepository),
+            nameof(_qualityProfileRepository),
+            nameof(_moveJobRepository)
+        )]
+        public void Init()
+        {
+            _services ??= new ServiceCollectionBuilder().Build();
+            _provider = _services.BuildServiceProvider();
```
therobbiedavis, line 57: Not a blocker but one logic thing I just noticed (Sorry!). I think `Init()` can replace an existing `_provider` without disposing it. Since the constructor already calls `Init()`, tests that add mocks and call `Init()` again leave the first provider behind, and `DisposeAsync()` only disposes the latest one.

listenarr.api/Program.cs
```diff
@@ -350,7 +328,6 @@ ex is IOException
 // Add default HTTP client for other services
 builder.Services.AddHttpClient();
 
-// Add named HttpClient for download operations (qBittorrent, Transmission, etc.)
 // Prevents socket exhaustion by reusing connections
 builder.Services.AddHttpClient("DownloadClient")
     .ConfigureHttpClient(client =>
```
therobbiedavis, line 333: Looks like some adapter-related registrations are still split between Program.cs and the extension methods. `DownloadClientsOptions` / its validator and `IImportItemResolutionService` are registered here on line 368, then `AddListenarrAdapters()` registers them again later. Is this ok or should we double register?

therobbiedavis (commented):
Both non-blocking, but just wanting to make sure we don't have any drift or artifacts.

therobbiedavis (approved):

tests/Common/BaseTests.cs
T4g1: Don't be sorry, it's a very good point, I completely disregarded the cleaning of these objects so I will update this

Issue #545, opened 2026-04-25
Why do webhooks require HTTPS URLs?


lhanson:
Using the latest `canary` release installed today, I'm trying to set up a notification webhook for a ntfy server on my LAN. I haven't set up TLS certificates because it seems like overkill for services on a network not exposed to the internet. It seems like the frontend [requires](https://github.com/Listenarrs/Listenarr/blob/b0185591df3f326915e38a2418cc86c01a673a0f/fe/src/views/settings/NotificationsTab.vue#L494) notification webhook URLs to be https:

<img width="679" height="460" alt="Image" src="https://github.com/user-attachments/assets/443f19d7-0663-417a-8a92-a234762fc454" />

Is it an intentional limitation to only allow HTTPS for notifications?

therobbiedavis:
@lhanson Not intentional, just an oversight as all my self-hosted stuff is either hosted via reverse-proxy or on the same internal only network and thereby accessible via docker container names.

therobbiedavis:
Closed as part of #555

PR #555, opened 2026-04-29
Allow HTTP URLs for notification webhooks

lhanson:
## Summary

Addresses #545 

### Changed
- Allow either HTTP or HTTPS for notification webhook URLs


therobbiedavis (changes requested):
@lhanson Simple and easy. Thanks! However we currently do not have any tests for this, could you add regression coverage for this in `NotificationsTab.spec.ts`? That spec currently only covers the loading state, so the URL scheme behavior is untested.

Suggested cases:
- accepts `https://ntfy.tld.com/topic`
- accepts `http://ntfy.local/topic`
- rejects a non-HTTP(S) scheme such as `ftp://ntfy.local/topic`

That would prove the reported LAN/self-hosted webhook case and guard against accidentally restoring HTTPS-only validation later.

lhanson:
Test cases added, thanks for those suggestions.

therobbiedavis:
@lhanson Can you rebase this? There is now a rebase conflict so I can not do it from my side unless I commit it myself.

Thanks!

therobbiedavis:
@lhanson The unit test is failing.

```
/home/runner/work/Listenarr/Listenarr/fe/src/__tests__/NotificationsTab.spec.ts
Error:   58:32  error  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any

lhanson:
Should be fixed now.

therobbiedavis (approved):

PR #565, opened 2026-05-01
Add {SeriesNumber:00} zero-padding support (#522)

### LINKED ISSUE #522: SeriesNumber does not have a pad option
When specifying a renaming format, the {SerieslNumber} does not have a {SeriesNumber:00} option.

This exists for {DiskNumber}.

Can it be added please so that my books appear in the correct order when there are more than 9 of them.

lzinga:
## Summary
 
 Add `{SeriesNumber:00}` zero-padding support to file naming patterns, 
 matching the existing `{DiskNumber:00}` and `{ChapterNumber:00}` behavior.
 Fixes #522
 
 ## Changes
 
 ### Added
 - `{SeriesNumber:00}` documented in Settings help text (folder and file pattern modals)
 - 5 backend tests for SeriesNumber zero-padding (single digit, two-digit, decimal, null, unformatted)
 
 ### Changed
 - Frontend preview logic now handles SeriesNumber zero-padding alongside DiskNumber/ChapterNumber
 
 ## Testing
 
 - Backend: `dotnet test --filter "FileNamingService_SeriesNumber"` — 5/5 pass
 - Manual: Verified backend formatting engine already supports `{SeriesNumber:00}` generically — no backend code 
changes needed, only frontend help text and preview
 
 ## Notes
 
 - No backend changes required — `FileNamingService.ApplyNamingPattern` already handles any `{Variable:format}` via 
`int.TryParse` + `ToString(format)`
 - Decimal series positions (e.g., 4.5) use the raw value since they can't be int-parsed — consistent with 
DiskNumber/ChapterNumber behavior

### DIFF
--- a/fe/src/components/settings/FileManagementSection.vue
+++ b/fe/src/components/settings/FileManagementSection.vue
@@ -152,7 +152,7 @@
               <li><code>{Asin}</code> - Audible ASIN</li>
               <li>
                 <code>{SeriesNumber}</code> or <code>{SeriesNumber:00}</code> - Position in series (00 =
-                zero-padded)
+                zero-padded; decimal positions like 4.5 are not padded)
               </li>
               <li><code>{Year}</code> - Publication year</li>
             </ul>
@@ -175,7 +175,7 @@
               <li><code>{Asin}</code> - Audible ASIN</li>
               <li>
                 <code>{SeriesNumber}</code> or <code>{SeriesNumber:00}</code> - Position in series (00 =
-                zero-padded)
+                zero-padded; decimal positions like 4.5 are not padded)
               </li>
               <li>
                 <code>{DiskNumber}</code> or <code>{DiskNumber:00}</code> - Disk/part number (00 =
@@ -253,7 +253,7 @@ function applyPattern(
   // Replace all variables with sample values
   for (const [key, value] of Object.entries(sampleVariables)) {
     if (key === 'DiskNumber' || key === 'ChapterNumber' || key === 'SeriesNumber') {
-      // Handle zero-padding for disk and chapter numbers using the sample value
+      // Handle zero-padding for disk, chapter, and series numbers using the sample value
       const paddedRegex = new RegExp(`\\{${key}:00\\}`, 'g')
       const paddedSample = value.toString().padStart(2, '0')
       result = result.replace(paddedRegex, paddedSample)
--- a/fe/src/components/settings/FileManagementSection.vue
+++ b/fe/src/components/settings/FileManagementSection.vue
@@ -150,7 +150,10 @@
               <li><code>{Publisher}</code> - Publisher name</li>
               <li><code>{Language}</code> - Metadata language</li>
               <li><code>{Asin}</code> - Audible ASIN</li>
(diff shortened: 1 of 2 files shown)


fe/src/components/settings/FileManagementSection.vue
```diff
@@ -246,7 +252,7 @@ function applyPattern(
 
   // Replace all variables with sample values
   for (const [key, value] of Object.entries(sampleVariables)) {
-    if (key === 'DiskNumber' || key === 'ChapterNumber') {
+    if (key === 'DiskNumber' || key === 'ChapterNumber' || key === 'SeriesNumber') {
       // Handle zero-padding for disk and chapter numbers using the sample value
```
therobbiedavis: Now a stale comment. Could you update to "disk, chapter, and series numbers" to keep it accurate.

fe/src/components/settings/FileManagementSection.vue
```diff
@@ -150,7 +150,10 @@
               <li><code>{Publisher}</code> - Publisher name</li>
               <li><code>{Language}</code> - Metadata language</li>
               <li><code>{Asin}</code> - Audible ASIN</li>
-              <li><code>{SeriesNumber}</code> - Position in series</li>
+              <li>
+                <code>{SeriesNumber}</code> or <code>{SeriesNumber:00}</code> - Position in series (00 =
```
therobbiedavis, line 154: Might be worth mentioning the caveat that decimal position like `4.5` won't be zero padded, unless you want to refactor to take that into account.

tests/Features/Api/Services/Import_PatternIntegrationTests.cs
```diff
@@ -323,5 +323,163 @@ public async Task FileNamingService_BothDiskAndChapter_UsesMultiFilePattern()
...
+                Series = null,
+                SeriesPosition = null
+            };
+
+            // Act
+            var result = await service.GenerateFilePathAsync(metadata, ".m4b");
+
+            // Assert - no series number, so the variable and surrounding separator should be cleaned up
+            Assert.Contains("Standalone Book", result);
+            Assert.DoesNotContain("00", result);
```
therobbiedavis, line 445: May also want:
`Assert.DoesNotContain(" - .", result);`
to make sure the filename is clean

therobbiedavis (changes requested):
@lzinga Some very minor comments

therobbiedavis:
@lzinga I'm fine with this direction, but should we pad decimals?

lzinga:
> @lzinga I'm fine with this direction, but should we pad decimals?

Padding decimals gets ambiguous (e.g. should 4.5 become 04.5, `04.50`, etc.), so I leaned toward leaving them as-is. This seems to follow what [Sonarr](https://wiki.servarr.com/sonarr/settings#standard-episode-format) and Radarr do.

Audiobooks are different than tv shows of course and as things grow, there may be a case for providing extended options. However, right now I don't think it's needed and we can always make a separate issue for it to be discussed further in the future.

therobbiedavis (approved):
lgtm

PR #553, opened 2026-04-28
Image performance optimizations & API hardening

therobbiedavis:
## Summary

This PR cleans up audiobook image loading, moves browser auth to HttpOnly cookie sessions, tightens configuration/API key access rules, slims the library API payload, and improves UI navigation responsiveness.

The frontend now uses direct backend image URLs instead of image tokens or blob/object URL plumbing. Audiobook, author, series, collection, wanted, and modal image paths all go through the simpler protected image helper, with lazy-loaded audiobook covers using shimmer placeholders.

The library list endpoint is now significantly leaner: the backend loads audiobooks without file includes, fetches compact format summaries and per-audiobook file counts in parallel, and returns a slim payload. This reduces memory and I/O on large libraries.

UI navigation is snappier: clicking a sidebar link immediately marks it as active (optimistic `pendingNavPath`) without waiting for async route guards to resolve, route transitions use a short page-fade to avoid blank flashes, and the startup-config cache window is extended to 30 s to reduce per-navigation waits.

It also updates the Add New search flow so searches only run when submitted, improves keyboard behavior, adds a public bootstrap config endpoint for pre-auth SPA startup, and fixes secret redaction/API key management rules so public remote callers cannot read unredacted secrets when auth is disabled.

This branch also includes cross-platform git hooks for local quality enforcement (pre-commit: lint-staged + layering + async void; pre-push: version sync + backend format + frontend type check + frontend tests) and runtime/CI maintenance for Docker image scanning and build notification behavior.

### Added
- Public startup bootstrap config endpoint for safe pre-auth SPA initialization
- `StartupBootstrapConfig` model
- Cookie-session auth tests and frontend auth store tests
- API image URL and image-cache tests
- API key management access tests
- `AppPathService` for shared app/config/cache path resolution
- `RequireAdminOrApiKeyWhenAuthenticationEnabled` filter
- `RequireAdministratorSessionWhenAuthenticationEnabled` filter
- `/system/ready` endpoint for local tooling and dev server readiness checks
- Audiobook shimmer placeholders while images lazy load
- `IAudiobookRepository.GetAllNoFilesAsync` — loads audiobooks without eagerly loading files
- `IAudiobookFileRepository.GetFormatSummariesAsync` — compact per-audiobook format summaries
- `IAudiobookFileRepository.GetCountsByAudiobookIdAsync` — file counts by audiobook id
- Optimistic sidebar active state (`pendingNavPath`) set on click, cleared after navigation resolves
[...]

PR #566, opened 2026-05-01
Add unauthenticated /ping health endpoint for TrueNAS

lzinga:
## Summary
 
 Adds a lightweight `/ping` endpoint at the root URL for use as a Docker/TrueNAS liveness probe. This matches the 
convention used by Sonarr, Radarr, Lidarr, and other *arr apps, where TrueNAS app catalog templates call `curl 
/ping` for health checks.
 
 ## Changes
 
 ### Added
 - `GET /ping` endpoint in SystemController — returns `{"status":"OK"}` with no authentication required
 - `<remarks>` XML doc explaining why the route uses a leading `/` to override the controller prefix
 - Integration tests for /ping (auth disabled + auth enabled scenarios)
 
 ## Testing
 
 - 2 integration tests in `PingEndpointTests.cs`:
   - `Ping_Returns200_WhenAuthDisabled` — verifies 200 OK without auth
   - `Ping_Returns200_WhenAuthEnabled_WithoutCredentials` — verifies [AllowAnonymous] bypasses auth middleware
 - Both pass against `ListenarrWebApplicationFactory`
 - Verified Sonarr's TrueNAS template uses `healthcheck.set_test("curl", {"path": "/ping"})` at root
 
 ## Notes
 
 - The leading `/` in `[HttpGet("/ping")]` is intentional — it overrides the controller's `api/v{version}/system` 
prefix so the endpoint lives at the root, matching the *arr convention
 
 ### Follow-up: TrueNAS App Catalog Submission
 
 Once this PR is approved and merged, the next step is to submit Listenarr to the [TrueNAS Apps 
Catalog](https://github.com/truenas/apps):
 
 1. **Fork** `truenas/apps` into the Listenarrs org
 2. **Create the app definition** at `ix-dev/community/listenarr/` modeled after [Sonarr's 
definition](https://github.com/truenas/apps/tree/master/ix-dev/community/sonarr), which includes:
    - `app.yaml` — metadata (name, version, categories, icon, screenshots, maintainers, run_as_context)
    - `item.yaml` — catalog entry (categories, icon_url, tags)
    - `ix_values.yaml` — default image config (Docker image repo + tag)
    - `questions.yaml` — user config schema (port, storage mounts, env vars, resources)
    - `templates/docker-compose.yaml` — Jinja2 template with `c1.healthcheck.set_test("curl", {"port": ..., "path": 
"/ping"})`
    - `templates/test_values/basic-values.yaml` — CI test scenario (required)
    - `README.md` — short app description
 3. **Publish a Docker image** (e.g., `ghcr.io/listenarrs/listenarr`) — required by `ix_values.yaml`
 4. **Host icon + screenshots** on a public URL for `app.yaml`
 5. **Submit a PR** to `truenas/apps` targeting the community train
 6. Per [TrueNAS CONTRIBUTIONS.md](https://github.com/truenas/apps/blob/master/CONTRIBUTIONS.md): open an issue 
first, run `python .github/scripts/ci.py` locally to validate, and open a draft PR early for maintainer feedback


listenarr.api/Controllers/SystemController.cs
```diff
@@ -86,6 +86,22 @@ public ActionResult<StorageInfo> GetStorageInfo()
...
+        /// Lightweight liveness probe for Docker/TrueNAS health checks.
+        /// Returns 200 OK with no authentication required and no service calls.
+        /// </summary>
+        /// <remarks>
+        /// The leading "/" overrides the controller's route prefix so this endpoint
+        /// is reachable at GET /ping (root) instead of /api/v1/system/ping.
+        /// TrueNAS app catalog templates (and other *arr apps) expect /ping at the root.
+        /// </remarks>
+        [AllowAnonymous]
+        [HttpGet("/ping")]
```
therobbiedavis, line 98: On #553 I have added a /ready which effectively does similar but for the wait-on startup. I don't think we need two endpoints for the same function but the question is which endpoint follows standards? I'm willing to refactor if we want to go the /ping route.

tests/Listenarr.Api.Tests/PingEndpointTests.cs
```diff
@@ -0,0 +1,70 @@
...
+    {
+        private readonly ListenarrWebApplicationFactory _factory;
+
+        public PingEndpointTests(ListenarrWebApplicationFactory factory)
+        {
+            _factory = factory;
+        }
+
+        [Fact]
+        public async Task Ping_Returns200_WhenAuthDisabled()
```
therobbiedavis, line 37: This test just uses whatever auth state the factory default is, whereas the second test explicitly sets auth as enabled. I think we should be explicit here as well and set auth as disabled.

listenarr.api/Controllers/SystemController.cs
therobbiedavis: Also might be worth going ahead and allowing `HEAD` requests here so uptime monitors, which a lot default to HEAD requests, will be able to resolve this.

therobbiedavis (changes requested):
Again some minor changes

listenarr.api/Controllers/SystemController.cs
lzinga: > On #553 I have added a /ready which effectively does similar but for the wait-on startup. I don't think we need two endpoints for the same function but the question is which endpoint follows standards? I'm willing to refactor if we want to go the /ping route.

Looking at /ready in #553 it's currently just a static 200 response same as /ping would be. The key differences here are:
 

1. `/ping` lives at root (no /api/v1/system prefix) which is what TrueNAS and other *arr apps expect (Sonarr, Radarr)
2.  `/ping` has to always return 200 no matter what state the app is in. If any startup logic (DB checks, service init, etc.) ever gets added to it and fails during boot, TrueNAS would restart the container thinking it's dead when it's really just booting up.

 We could point the TrueNAS template at /api/v1/system/ready instead but then we're the odd one out in the *arr
 catalog and relying on that endpoint never getting readiness logic added that could fail during migrations.
 
 I think both can coexist. `/ping` is the "is the process alive" contract for container orchestrators, /ready can
 evolve into actual readiness checks down the road if needed. If we go down this path then ping can evolve to have the checks as well.
 
 ```csharp
  public ActionResult Ping()
 {
     // During startup, always report alive
     if (!_readinessService.IsReady)
         return Ok(new { status = "OK" });
 
     // After startup, also check DB connectivity
     try
     {
         _configurationService.GetConfiguration();
     }
     catch
     {
         return StatusCode(500, new { status = "Error" });
     }
     return Ok(new { status = "OK" });
 }
 ```
 
 Ultimately, I am fine with any method, I just chose /ping as that is what radarr/sonarr and truenas use. I have gone ahead and fixed the other issues but will leave this one open for now until we figure out the best path forward.

therobbiedavis: Lets do both then, /ping for is process alive, and /ready for has application started

tests/Listenarr.Api.Tests/PingEndpointTests.cs
```diff
@@ -0,0 +1,80 @@
...
+        {
+            using var factory = _factory.WithWebHostBuilder(builder =>
+            {
+                builder.ConfigureServices(services =>
+                {
+                    services.AddSingleton<Listenarr.Api.Services.IStartupConfigService>(sp =>
+                    {
+                        return new TestStartupConfigService(new StartupConfig { AuthenticationRequired = "Disabled" });
```
therobbiedavis, line 45: The rest of the test suite uses "false" instead of "Disabled". It returns false here only because it falls through the allowlist same as any other unknown string. Can you change is to false to be consistent and explicit?

tests/Listenarr.Api.Tests/PingEndpointTests.cs
```diff
@@ -0,0 +1,80 @@
+/*
```
therobbiedavis, line 1: I think we should also add a test case for the new [HttpHead("/ping")]. Which in that case it might be worth looking at going ahead and creating a private helper to make the test fixture setups easier.

therobbiedavis (approved):
@lzinga There are a couple non-blockers, so approving for now. If you want to address the non-blockers, feel free, and i'll re-review.

Issue #572, opened 2026-05-08
[Architecture] Notification subscriber/publisher


T4g1:
The current status of notifications is strongly coupled to particular services and spread accross several of them. There are multiple endpoints depending on what is happening and the responsability to trigger notification must be known and shared by any class.

This is not optimal as ideally, we'd want to not have to care about what should or should'nt be notified when developping new features. Therefore I suggest this architecture, which is already quite aligned with what was planned:
* Single entry point for all notifications: `NotificationService`
* Subscriber model: `INotificationSubscriber` can subscribe to the `NotificationService`
* Event driven: `NotificationService` exposes a generic method to publish information on topics/channels: Like `Notify/Publish(channel/topic, content)`
* Typed events: Let's have an enumeration to centralize available topics and documentation: Like `enum NotificationChannel`

```mermaid
graph TD
    %% Source
    AppCore[Listenarr] -- "NotificationEvent" --> NS[NotificationService]

    %% Orchestrateur
    subgraph Notification Service
        NS
    end

    %% Interface
    NS -- "INotificationSubscriber" --> Interface((INotificationSubscriber))
    Interface --> DH[DownloadHub]
    Interface --> LH[LogHub]
    Interface --> Discord[Discord Provider]
    Interface --> PB[Pushbullet Provider]

    %% Abonnés Internes (SignalR)
    subgraph SignalR Hubs
        DH[DownloadHub]
        LH[LogHub]
    end

    %% Abonnés Externes (Webhooks)
    subgraph External / Webhooks
        Discord[Discord Provider]
        PB[Pushbullet Provider]
    end

    %% Clients Finaux
    DH -- "Push: ProgressUpdate" --> Browser[Browser]
    LH -- "Push: NewLogLine" --> Dashboard[Admin Dashboard]
    Discord -- "HTTP POST" --> DiscordAPI[Discord]
```

Take a look here for example: 
https://github.com/Listenarrs/Listenarr/blob/8846b98efe329a53e333f732f4bbd365382e50ac/listenarr.api/Services/NotificationService.cs#L244-L245

Currently, `NotificationService` implements specific interfaces logics, this makes it difficult to maintain as we add more interfaces or when we need to update/test existing ones
All those should becomes specific classes that implements `INotificationSubscriber` instead.

Also, regarding the directory structure: I suggest that the interface should be placed under `listenarr.application/Interfaces` while implementations should go in `listenarr.infrastructure/Notifications` or something like that

[...]

PR #576, opened 2026-05-10
Download importation improvements + Clarification of project structure

### LINKED ISSUE #492: [Architecture] Clarify project structure
Documented Project Structure in CONTRIBUTING (https://github.com/Listenarrs/Listenarr/blob/canary/CONTRIBUTING.md) is not up to date. At least the following is missing:
* `listenarr.application`
* `listenarr.domain`
* `listenarr.infrastructure`

Among other things, the following is not clear:
* Do we put data structures in `listenarr.domain` ?
* Does the project follows typical DDD (Domain-Driven Design) or not ?
* What kind of data structures should be in `listenarr.api/Models` ?
* Should we put converters/mapper in:
** domain to/from infrastructure: `listenarr.infrastructure` ?
** domain to/from api: `listenarr.api` ?
** domain to/from domain: `listenarr.domain` ?

### LINKED ISSUE #535: [Architecture] Download processing
### Context

Responsability about the download monitoring and processing responsability is not clearly defined in the code base, multiple files share that responsability with no real orchestration which leads to inconsistencies in download handling

We should rework that to provide a clear unique way of processing download lifecycle and remove redundancies

### Ideas

We have multiple responsabilities to indentify and handle:
* Create download: That is already handled by download clients
* Monitor download: One background service should be limited to monitoring progress on downloads (with no knowledge of download client nor what is done on them). It should know the lifecycle of download (as a graph of possible statuses) and trigger callbacks when download status changes

If we go for a job based architecture (which make it easier to recover from unexpected interruption):
* We could have IDownloadHandler: Each of those would subscribe to the download monitoring service updates on specific topic (download completed, download started, download processed, ...)
* The IDownloadHandler for completed download could then start file processing jobs: Source, Destination, Action
* Process files: Only one service should handle file logic and response to file processing jobs

I believe this could give a good flexibility and ultimately could also be used for manual/automatic imports requested by the user (as it would simply means creating a file processing job)

T4g1:
## Summary
* Closes #535 
* Closes #492 

There's a lot of changes here mostly because of the file re-structuration done in order to align better on the planified directory structure.

Regarding the download refactoring, here is a slightly simplified overview of how it works now:

```mermaid
sequenceDiagram
    participant Monitor as DownloadMonitor
    participant Gateway as DownloadClientGateway
    participant Adapter as IDownloadClientAdapter
    participant JobService as DownloadProcessingJobService
    participant Processor as DownloadProcessingJobProcessor
    participant ImportService as DownloadImportService
    participant FileMover as FileMover

    Note over Monitor, Adapter: Phase 1: Monitoring
    loop Every X seconds
        Monitor->>Gateway: FetchDownloadsAsync(client, downloads)
        Gateway->>Adapter: FetchDownloadsAsync(client, downloads)
        Adapter-->>Gateway: List<Download>
        Gateway-->>Monitor: List<Download>
    end

    Note over Monitor, JobService: Phase 2: Completion Trigger
    opt Status == Completed
        Monitor->>JobService: EnqueueAsync(download)
        JobService-->>Monitor: Task Created
    end

    Note over Processor, FileMover: Phase 3: Processing & Import
    Processor->>JobService: GetNextJobAsync()
    Processor->>Gateway: GetDownloadedFiles(download)
    Gateway->>Adapter: GetQueueItemAsync(client, download)
    Adapter-->>Gateway: QueueItem
    Gateway-->>Processor: QueueItem with list of files

    Processor->>ImportService: ImportDownloadFilesAsync(audiobook, files)
    loop For each file
        ImportService->>FileMover: PerformAction(Action, Source, Destination)
        FileMover-->>ImportService: Success
    end
    ImportService-->>Processor: File Imported
```

The main things to note is that all duplication of responsability regarding download completion has now been removed. Import operations are handled only with jobs so in case of system failure, the job can be reset and restarted effortlessly

## Changes
* File structure reorganized to better respect the planified architecture
* Services responsabilities updated
* Remote path mapping should be done by IDownloadClientGateway only
* Download should have only the path where the client gateway reported they where placed
* Audiobook base path is now the source of truth to move/import files instead of each services applying it's own logic
* Only IFileMover should be allowed to perform action on files and directories
* Clarified responsabilities for DownloadMonitorService

## Testing
* Add more tests and rewrote existing ones to comply with the new processing
* This probably needs more manual testing

## Next steps
* Notification, history and orphaned download management may have suffered some regression
* Regarding notification, an issue is already prepared for it (it may even handle history/logs if done correctly: #572)
* Tests are still organized by mirroring the old file structure, they should be updated too sometime
* The profile/quality filtering for imported file should be reworked: #582


listenarr.application/Downloads/DownloadImportService.cs
```diff
@@ -0,0 +1,583 @@
...
+                                {
+                                    filename = Path.GetFileName(filename);
+                                }
+                            }
+
+                            var destPathForFile = CombineWithOptionalBase(destDirForFile, filename);
+
+                            res.Success = await fileMover.PerformActionOn(completedFileAction, file, destPathForFile);
+                            res.FinalPath = destPathForFile;
+                            res.Success = true;
```
therobbiedavis: Looks like this always flips res.Success back to true after the mover runs. If the move/copy fails, should we leave the mover result intact?

listenarr.application/Downloads/DownloadProcessingJobProcessor.cs
```diff
@@ -0,0 +1,247 @@
...
+                await downloadProcessingJobService.UpdateJobAsync(job.ScheduleRetry("No importable files found in source directory"));
+                return;
+            }
+
+            job.AddLogEntry($"Found {files.Count} importable file(s) to process");
+
+            try
+            {
+                var downloadImportService = scope.ServiceProvider.GetRequiredService<IDownloadImportService>();
+                await downloadImportService.ImportDownloadFilesAsync(audiobook, files, cancellationToken);
```
therobbiedavis: It looks like we don’t check the import results before completing the job. If imports can return only failed/skipped results without throwing, this could mark a download as Moved even when nothing useful was imported.

listenarr.api/Attributes/LocalOrAdminAttribute.cs
```diff
@@ -0,0 +1,21 @@
...
+    public class LocalOrAdminAttribute : Attribute, IAuthorizationFilter
+    {
+        public void OnAuthorization(AuthorizationFilterContext context)
+        {
+            var isLoopback = SecurityRequestUtils.IsLoopbackRequest(context.HttpContext);
+            var isAuth = SecurityRequestUtils.IsAuthenticatedAdminOrApiKey(context.HttpContext);
+
+            if (!isLoopback && !isAuth)
+            {
+                context.Result = new ForbidResult();
```
therobbiedavis: I couldn’t find a default auth scheme registration, so ForbidResult can throw instead of returning 403. A direct 403 result would match the old guard behavior more safely.

listenarr.domain/Models/DownloadProcessingJob.cs
```diff
@@ -132,49 +132,83 @@ public class DownloadProcessingJob
...
             if (RetryCount >= MaxRetries)
             {
                 MarkAsFailed($"Max retries ({MaxRetries}) exceeded");
-                return;
+                return this;
             }
 
             RetryCount++;
-            Status = ProcessingJobStatus.Retry;
+            Status = ProcessingJobStatus.Pending;
```
therobbiedavis, line 204: This retry state confuses me. ScheduleRetry() sets NextRetryAt, but once the job is Pending, GetNextPendingAsync() appears to pick it up without checking that timestamp. Shouldn't pending jobs with a future NextRetryAt be skipped?

therobbiedavis (changes requested):
Found what seem to be some logic issues

listenarr.application/Downloads/DownloadProcessingJobService.cs
```diff
@@ -0,0 +1,128 @@
...
+            return job.Id;
+        }
+
+        public async Task<DownloadProcessingJob?> GetNextJobAsync()
+        {
+            try
+            {
+                var jobs = await jobRepository.GetJobsByStatusAsync(ProcessingJobStatus.Pending);
+                return jobs
+                    .Where(j => j.NextRetryAt == null || j.NextRetryAt >= DateTime.UtcNow)
```
therobbiedavis: I think this comparison is flipped. `NextRetryAt` is the time the job should wait until, so a future value should keep it out of the queue for now. With `>=`, future retries are eligible immediately, and once the retry time actually passes they stop matching. I think this should be `NextRetryAt == null || NextRetryAt <= DateTime.UtcNow`.

listenarr.application/Downloads/DownloadProcessingJobProcessor.cs
```diff
@@ -0,0 +1,268 @@
...
+            foreach (var result in results)
+            {
+                if (!result.Success)
+                {
+                    await downloadProcessingJobService.UpdateJobAsync(job.MarkAsFailed($"Unable to import at least one file for the job (see the log entries)"));
+                    return;
+                }
+            }
+
+            await downloadService.UpdateAsync(download.Imported());
```
therobbiedavis, line 284: I think we should be a bit stricter here before calling the download imported. If any of the audio files that belong to this download were skipped or not imported, marking the whole thing as `Moved` feels risky, since the cleanup path can later remove the client item/DB record. Companion files can probably stay best-effort, but I’d expect all related audio files to import successfully before this job is considered complete.

listenarr.application/Downloads/DownloadImportService.cs
```diff
@@ -0,0 +1,579 @@
...
+                                    ? Path.GetRelativePath(sourceRootPath, file)
+                                    : Path.GetFileName(file);
+                                if (relativePath.StartsWith("..", StringComparison.Ordinal))
+                                {
+                                    relativePath = Path.GetFileName(file);
+                                }
+
+                                var destination = CombineWithOptionalBase(batchDestinationRoot, relativePath);
+
+                                if (!await fileMover.PerformActionOn(completedFileAction, file, destination))
```
therobbiedavis, line 154: Small logic issue here: if `PerformActionOn` returns false for a companion file, we add a failure result but then immediately add a success result for the same file too. Could this be an `else`, or should we `continue` after recording the failure?

therobbiedavis (changes requested):

listenarr.application/Downloads/DownloadProcessingJobService.cs
T4g1: Damn, I will add a test for that case

listenarr.application/Downloads/DownloadImportService.cs
T4g1: I'll add a test for that too ^^' If you think of more test i could do, send them to me!

listenarr.application/Downloads/DownloadProcessingJobProcessor.cs
T4g1: But with this PR, we skip only on two cases:
- Companion file in an import that has not managed to process any audio file: In that case, two possible scenario:
  - Some audio file are already success = false so the download will not be marked as imported
  - There are no audio file to import: Maybe a previous attempt already moved them, should this be a failure ?
- Better audio quality file already exists:
  - Seems safe to ditch the file and says we have succeeded at doing nothing ?
What do you think ?

therobbiedavis: Fair point, I checked and you're right. It doesn't look like we have any test cases for these scenarios though, and I definitely think we should.

therobbiedavis (changes requested):

listenarr.application/Downloads/DownloadProcessingJobProcessor.cs
therobbiedavis: So the new import-service tests cover the skip behavior, but I think the missing coverage is at the processor boundary: one test proving lower-quality skipped audio intentionally completes/marks `Moved`, and one test proving failed audio import results fail the job and do not mark the download `Moved`. Evenutually I like for this to be more semantically appropriate, so can you add a TODO here as well to update to multiple markings: `Imported`, `Skipped`, and `Blocked` so we can branch logic easier.

therobbiedavis:
@T4g1 Let's focus on getting this one merged in since we have a lot of other PRs that will need to be refactored and rebased for this and the previous release. Then we should probably let's hold on structural/foundational PRs until we can create some coordinated sprints.

listenarr.application/Downloads/DownloadProcessingJobProcessor.cs
T4g1: So I went in this direction:
* If files downloaded and files on disk do not match: Retry
* If import success: Check if at least one file was added as an audiobookfile
* If none were added as audiobook file: Check if audiobook files exists on this audiobook (suggesting quality skip)
* If none exist: Fail
* Otherwise, success

What do you think ? Should we still allow import of partial downloads ? If files are inconsistent with what download client reports, I figured it's probably better to ask for manual intervention as the process might produce unexpected results down the line

listenarr.application/Downloads/DownloadItemService.cs
```diff
@@ -0,0 +1,112 @@
...
+
+            if (string.IsNullOrEmpty(download.DownloadPath))
+            {
+                logger.LogDebug($"Download {download.Id} has no path configured, unable to locate where the files should be");
+                return [];
+            }
+
+            try
+            {
+                var importableFiles = Directory.EnumerateFiles(download.DownloadPath, "*.*", SearchOption.AllDirectories)
```
therobbiedavis: I think this can regress cases where `DownloadPath` is the file itself. The processor still accepts file or directory paths, but this now always calls `Directory.EnumerateFiles(...)`, so a file path will throw and the job will retry instead of importing.

Was there a reason we moved away from the previous file-path handling here?

listenarr.application/Downloads/DownloadProcessingJobProcessor.cs
```diff
@@ -0,0 +1,298 @@
...
+                    await downloadProcessingJobService.UpdateJobAsync(job.MarkAsFailed($"Unable to import at least one file for the job (see the log entries)"));
+                    return;
+                }
+
+                wasRegisteredToAudiobook |= result.WasRegisteredToAudiobook;
+            }
+
+            if (!wasRegisteredToAudiobook)
+            {
+                // If the audiobbook already had some audiobook file, this download has probably been skipped
```
therobbiedavis: Tiny typo while you’re here: `audiobbook` (two b's) -> `audiobook`. 😆

listenarr.application/Downloads/DownloadProcessingJobProcessor.cs
therobbiedavis: This direction makes sense to me. Retrying on a disk mismatch feels safer than allowing a partial import and then entering cleanup.

The new tests cover the cases I was worried about: lower-quality skips can complete, and missing reported files keep the job retrying. I left one separate note on the single-file `DownloadPath` path.

therobbiedavis (changes requested):

listenarr.application/Downloads/DownloadItemService.cs
T4g1: woops, i tried a direction where Download would not be required but I had issue finding a valid path to list files then, i reverted but this check disapeared in the process, my bad

Also, thinking about it, I wonder, shouldn't we loop through queue item source files and check those where File.Exist is true instead of browsing a directory and trying to match files to the source files ? (which seems to effectively do the same things in the end but is more error prone), wdyt ?

T4g1: Here is what i did: Every commit except for the last one is focused on rebasing (fixing localization issue) and reverting the change to the control you mention

Then, I added the last commit (https://github.com/Listenarrs/Listenarr/pull/576/changes/df29955b20fd39f853e127002aa36145454dc4be) to see what it would look like with the simplification i suggested

Let me know if you rather keep the first implementation or not

therobbiedavis: I like the simplification, but depending on how the conversation goes in Discord, we may need to add some additional guards.

listenarr.application/Interfaces/IDownloadClientAdapter.cs
```diff
@@ -63,7 +73,10 @@ Task<DownloadClientItem> GetImportItemAsync(
             CancellationToken ct = default);
 
         /// <summary>
-        /// Legacy method for backward compatibility
+        /// Retrieves the information about a given download as a queue item
+        /// The adapter should return:
+        /// - Either a list of files under SourceFiles
+        /// - Either a RemotePath that can be a file or a directory if the adapter is unable to tell
```
therobbiedavis: I think this should be ContentPath, not RemotePath.

therobbiedavis (approved):
Based on our conversation in Discord this now aligns with the direction we discussed.

It now treats non-empty SourceFiles as authoritative, but falls back to ContentPath when SourceFiles is null or empty. It also maps paths before filesystem checks, expands directory ContentPath, keeps file ContentPath, logs scan failures, returns [] on unresolved paths, and leaves retry orchestration in the processor.

I did leave one, non-blocking clean up.

Future direction note: ideally the adapter should return import-ready `SourceFiles`. That means the adapter should use the shared remote path mapping helper/service, resolve file-vs-directory cases, normalize the paths, and log why it could not resolve anything.

Long term, the gateway should not be doing path mapping, `ContentPath` fallback scanning, or missing-path diagnostics as part of normal import resolution. It should mostly pass through the adapter’s resolved result.

The contract I’d like us to move toward in #592 is:
- `SourceFiles = null` means the adapter could not determine source files.
- `SourceFiles = []` means the adapter determined there are no importable files.
- Non-empty `SourceFiles` should already be mapped, normalized, validated, and ready for the processor.
- `ContentPath` should become a legacy/fallback field, not the normal import path.

T4g1:
> Based on our conversation in Discord this now aligns with the direction we discussed.
> 
> It now treats non-empty SourceFiles as authoritative, but falls back to ContentPath when SourceFiles is null or empty. It also maps paths before filesystem checks, expands directory ContentPath, keeps file ContentPath, logs scan failures, returns [] on unresolved paths, and leaves retry orchestration in the processor.
> 
> I did leave one, non-blocking clean up.
> 
> Future direction note: ideally the adapter should return import-ready `SourceFiles`. That means the adapter should use the shared remote path mapping helper/service, resolve file-vs-directory cases, normalize the paths, and log why it could not resolve anything.
> 
> Long term, the gateway should not be doing path mapping, `ContentPath` fallback scanning, or missing-path diagnostics as part of normal import resolution. It should mostly pass through the adapter’s resolved result.
> 
> The contract I’d like us to move toward in #592 is:
> 
> * `SourceFiles = null` means the adapter could not determine source files.
> * `SourceFiles = []` means the adapter determined there are no importable files.
> * Non-empty `SourceFiles` should already be mapped, normalized, validated, and ready for the processor.
> * `ContentPath` should become a legacy/fallback field, not the normal import path.

You probably should comment part of that on the issue so it's not forgotten when we merge here!

PR #579, opened 2026-05-11
Fixing naming pattern adding {Subtitle}

### LINKED ISSUE #571: Implicit patterns added to configured naming patterns
> Another issue related to naming structure. If I set single file or multipart file naming with {Title} is is also adding {Subtitle} afterwards. For example, If in a audiobook's metadata I have this:
>  
> Title: Example
> Subtitle: Yup
> 
> It is naming the file --> Example - Yup.ext instead of just Example.ext
> 
> Is this intended or a bug? 

 _Originally posted by @lamb0985 in [#504](https://github.com/Listenarrs/Listenarr/issues/504#issuecomment-4368939814)_

Sounds like we should either make it explicit (change the configured naming pattern and notify the user) or apply given naming patterns exactly
Either way, it's worth investigating

therobbiedavis:
## Summary

Fixes #571 by making configured naming patterns apply exactly as written. `{Title}` now resolves to only the audiobook title instead of implicitly appending the subtitle, while `{Subtitle}` remains available when users explicitly include it.

### Added
- Added backend regression coverage for single-file and multi-file manual imports using `{Title}` with audiobook subtitles present.
- Added organize/rename preview regression coverage to ensure title-only patterns do not include subtitles.
- Added frontend coverage for `{Subtitle}` replacement in the audiobook detail path estimate.

### Changed
- Updated audiobook detail path estimation to replace backend-supported metadata tokens including `{Subtitle}`, `{Edition}`, `{Narrator}`, `{Publisher}`, `{Language}`, `{Asin}`, and `{Quality}`.

### Fixed
- Removed implicit title/subtitle merging from manual import naming.
- Removed implicit title/subtitle merging from organize/rename preview naming.

## Testing

- `dotnet test tests/Listenarr.Tests.csproj --filter "FullyQualifiedName~ManualImportControllerTests|FullyQualifiedName~RenameServiceTests|FullyQualifiedName~FileNamingService_PatternSelectionTests" /p:DefaultItemExcludes=**/obj/**`
- `npm run test:unit -- --run src/__tests__/AudiobookDetailView.spec.ts`

## Notes

`{Subtitle}` was already listed in the File Management settings help. This change makes the backend and detail-view preview honor that token explicitly without changing `{Title}` output.


fe/src/views/library/AudiobookDetailView.vue
```diff
@@ -1046,27 +1048,38 @@ const displayBasePath = computed(() => {
...
+  const narrator = audiobook.value?.narrators?.filter(Boolean).join(', ') || ''
+  const publisher = audiobook.value?.publisher || ''
+  const language = audiobook.value?.language || ''
+  const asin = audiobook.value?.asin || ''
   const year = audiobook.value?.publishYear || ''
   const seriesNumber = audiobook.value?.seriesNumber || ''
+  const quality = audiobook.value?.quality || ''
 
   // Basic variable replacement mirroring server pattern keys
-  let relative = pattern
```
T4g1, line 1048: We could maybe avoid duplicating the pattern logic in the front end, why not have some kind of controller with endpoint to apply pattern for audiobook/file path instead that use directly the application service logic ?

.husky/pre-commit
```diff
@@ -1,5 +1,29 @@
 #!/usr/bin/env sh
 set -e
 
+# Some Windows Git clients launch hooks with a reduced PATH that omits Node.
+if ! command -v node >/dev/null 2>&1 && [ -d "/c/Program Files/nodejs" ]; then
+  export PATH="/c/Program Files/nodejs:$PATH"
+fi
+
 echo "Running staged lint and format checks..."
-npm run lint:staged
```
T4g1: Why not keep the commands unified with what package.json defines ? We probably want to keep the behavior united so contributors can run simple npm command to make sure commit/push will work

.husky/pre-commit
```diff
@@ -1,5 +1,29 @@
...
+node scripts/lint-staged.mjs
+
+echo "Checking layering rules and async void..."
+VIOLATIONS=0
```
T4g1: Same here, we should probably make this as a sperate script with an easy `npm run test:arch` for example (nice to see that being added btw :D )

.husky/pre-push
T4g1, line 1: Same remark here for the package.json commands to keep an unified behavior

tests/Features/Api/Controllers/ManualImportControllerTests.cs
T4g1, line 1: General remark for the tests but you can leverage the new framework to make tests more concise (builder pattern, temp file/directory management, service providers)

.husky/pre-commit
```diff
@@ -6,24 +6,22 @@ if ! command -v node >/dev/null 2>&1 && [ -d "/c/Program Files/nodejs" ]; then
   export PATH="/c/Program Files/nodejs:$PATH"
 fi
 
-echo "Running staged lint and format checks..."
-node scripts/lint-staged.mjs
+resolve_node() {
```
T4g1: We could probably avoid duplicating the logic by using a common script that defines this function. Also, just before this, you modify the PATH which could also be done in this function instead right ?

tests/Builders/QualityProfileBuilder.cs
```diff
@@ -0,0 +1,34 @@
+using Listenarr.Domain.Models;
+
+namespace Listenarr.Tests.Builders
+{
+    public class QualityProfileBuilder
+    {
+        private readonly QualityProfile _qualityProfile = new()
+        {
+            Name = "Test Profile",
```
T4g1, line 12: The default builder should probably handle the ID so this object is valid for database insertion

listenarr.api/Services/FileNamingService.cs
```diff
@@ -161,174 +156,15 @@ public async Task<string> GenerateFilePathAsync(
         /// </summary>
         public string ApplyNamingPattern(string pattern, Dictionary<string, object> variables, bool treatAsFilename = false)
```
T4g1, line 158: We probably want to drop this method as it's no longer the responsability of this file

listenarr.api/Services/FileNamingService.cs
T4g1, line 1: i'd say this should go in Naming directory too but this will probably conflict with the PR that takes care of that then :x

listenarr.application/Audiobooks/AudiobookPathPreviewService.cs
```diff
@@ -0,0 +1,151 @@
...
+        }
+
+        private string ComputeBaseDirectoryFromPattern(
+            Audiobook audiobook,
+            string rootPath,
+            string fileNamingPattern)
+        {
+            var directoryPattern = BuildDirectoryPattern(fileNamingPattern, audiobook);
+
+            var variables = new Dictionary<string, object>
```
T4g1: This should probably be the responsibility of the NamingService, building this data structure is very specific to that service only

listenarr.api/Services/FileNamingService.cs
```diff
@@ -161,174 +156,15 @@ public async Task<string> GenerateFilePathAsync(
...
-                result = extensionSeparator >= 0
-                    ? deviceNameStem + "_" + result[extensionSeparator..]
-                    : result + "_";
-            }
-
-            return result;
+            return _namingPatternService.SanitizePathComponent(pathComponent);
         }
 
         private Dictionary<string, object> BuildVariables(AudioMetadata metadata)
```
T4g1, line 163: Duplicated logic with ComputeBaseDirectoryFromPattern

listenarr.application/Interfaces/IAudiobookPathPreviewService.cs
T4g1, line 1: General remark: I would leave the interfaces in the interfaces directory, that way other layers only requires using Listenarr.Application.Interfaces and nothing else idealy

listenarr.application/Naming/INamingPatternService.cs
T4g1, line 1: Same remark for interfaces location

listenarr.application/Naming/NamingPatternService.cs
```diff
@@ -0,0 +1,177 @@
...
+        };
+
+        public string ApplyNamingPattern(
+            string pattern,
+            Dictionary<string, object> variables,
+            bool treatAsFilename = false)
+        {
+            if (string.IsNullOrWhiteSpace(pattern))
+            {
+                return "Unknown";
```
T4g1: If we give an empty pattern, we expect an empty path no ? Returning Unknown is probably problematic if we apply that for folderNamingPattern for example and the user does not want any subdirectory for some reason

listenarr.application/Audiobooks/AudiobookPathPreviewService.cs
```diff
@@ -0,0 +1,151 @@
...
+            Audiobook audiobook,
+            string? destinationRoot = null,
+            CancellationToken ct = default)
+        {
+            var settings = await settingsRepository.GetAsync(ct) ?? new ApplicationSettings();
+            var root = !string.IsNullOrWhiteSpace(destinationRoot)
+                ? destinationRoot
+                : settings.OutputPath;
+
+            var namingPattern = !string.IsNullOrWhiteSpace(settings.FolderNamingPattern)
```
T4g1: I do not think we should use the FileNamingPattern to build a basePath for the audiobook
Download import is based on the file naming pattern while this should remain based on folder naming pattern

listenarr.application/Interfaces/IAudiobookPathPreviewService.cs
```diff
@@ -0,0 +1,12 @@
+using Listenarr.Domain.Models;
+
+namespace Listenarr.Application.Audiobooks
+{
+    public interface IAudiobookPathPreviewService
+    {
+        Task<PathPreviewResult> PreviewAsync(
+            Audiobook audiobook,
+            string? destinationRoot = null,
```
T4g1, line 10: I find this a bit confusing, I assume this is the prefered root folder the user wants for this audiobook ? Maybe add a comment or explicitely name it selectedRootFolder ? Also, why do we allow null values ? To create a base path for an audiobook, we absolutely need a root folder right ? The controller should probably safeguard that so user "trust" component stays in API layer

T4g1: Also, we could directly use the RootFolder object if controller does the preliminary checks

listenarr.application/Audiobooks/AudiobookPathPreviewService.cs
```diff
@@ -0,0 +1,151 @@
...
+        }
+
+        private string SanitizeOptional(string? value)
+        {
+            return string.IsNullOrWhiteSpace(value)
+                ? string.Empty
+                : namingPatternService.SanitizePathComponent(value);
+        }
+
+        private static string BuildDirectoryPattern(string fileNamingPattern, Audiobook audiobook)
```
T4g1: This feels strange to have here, i expect the responsibility of AudioobookPathPreviewService to be limited to building a path from a pre-configured pattern.
This method should probably be used while configuring those patterns instead to validate them eventually (I haven't look at that part of the code yet but I expect that to be in the IConfigurationService probably)

listenarr.application/Audiobooks/AudiobookPathPreviewService.cs
```diff
@@ -0,0 +1,151 @@
...
+                directoryPattern = Regex.Replace(directoryPattern, @"\{Series[^}]*\}", string.Empty, RegexOptions.IgnoreCase);
+                directoryPattern = Regex.Replace(directoryPattern, @"[\\/]\s*[\\/]", "/");
+                directoryPattern = Regex.Replace(directoryPattern, @"^\s*[\\/]", string.Empty);
+                directoryPattern = Regex.Replace(directoryPattern, @"[\\/]\s*$", string.Empty);
+            }
+
+            return directoryPattern;
+        }
+
+        private static string ResolvePathWithOptionalBase(string? basePath, string candidatePath)
```
T4g1: I would put that in FileUtils or something as this is static
Also, why not put basePath as an optional argument instead of a mandatory null when we do not use it ? (method name should then be updated as ResolvePath)

listenarr.application/Naming/INamingPatternService.cs
```diff
@@ -0,0 +1,12 @@
+namespace Listenarr.Application.Naming
+{
+    public interface INamingPatternService
+    {
+        string ApplyNamingPattern(
+            string pattern,
+            Dictionary<string, object> variables,
+            bool treatAsFilename = false);
+
+        string SanitizePathComponent(string pathComponent);
```
T4g1: This seems specific to the internal working of this service so it should probably not be exposed by the interface

T4g1 (changes requested):
I believe we can improve the separation of concerns and thus avoid code duplication even further here

listenarr.api/Services/FileNamingService.cs
T4g1: As this will be merged after the file re-structuration, i leave this open as a reminder to make sure we no longer have a Services directory once this has been rebased

PR #580, opened 2026-05-12
fix(nzbget): preserve Basic auth across same-host non-downgrade redirects

kevinheneveld:
## Summary

Adding NZBGet as a download client could fail with `NZBGet XML-RPC error: Unauthorized` even when the same credentials worked from a manual `curl`. The trigger is a reverse proxy in front of NZBGet (Caddy/Nginx/Traefik HTTPS upgrade, trailing-slash normalization, etc.) returning a 30x — .NET's default `HttpClientHandler` follows the redirect but strips `Authorization` to prevent credential leakage to unintended hosts, so NZBGet sees the follow-up without auth and 401s.

## Approach

Revised after [@T4g1's review](https://github.com/Listenarrs/Listenarr/pull/580#issuecomment-...) — the original commit on this branch (URL-embedded `user:pass@host`) was abandoned because it would also forward credentials across cross-host or HTTPS→HTTP redirects, defeating the very security control it was working around.

The new approach treats this as a redirect-handling concern rather than an auth-encoding concern:

* The named `"nzbget"` HttpClient is now configured with `AllowAutoRedirect = false` and a custom `NzbgetSafeRedirectHandler` in its delegating-handler chain.
* On 301/302/303/307/308, the handler validates that the next URL is **(a)** the same host:port as the previous request and **(b)** not an HTTPS→HTTP downgrade. Only then does it re-apply the `Authorization` header on the next request.
* 307/308 preserve method and (buffered) body per RFC 7231; 301/302/303 change to GET, matching `HttpClientHandler`'s defaults.
* Hop cap of 5 to defend against redirect loops.
* On any rule violation (cross-host, scheme downgrade, missing `Location`, redirect loop), the handler throws a typed `NzbgetSafeRedirectException` with a descriptive message. `TestConnectionAsync` surfaces that message verbatim so users see the actual misconfiguration instead of a generic `"Authentication failed"` / `"network error"`.

The handler is wired into both `Program.cs` and the equivalent `ServiceRegistrationExtensions.AddListenarrHttpClients` registration, so every code path that uses the named `"nzbget"` HttpClient is covered — including `FetchDownloadsAsync` which calls `/jsonrpc` (the JSON-RPC gap T4g1 flagged in the same review). `AddViaJsonRpcAsync` delegates through `CallXmlRpcAsync` and is covered transitively.

The Authorization header is now the sole auth path. NZBGet's own `WebServer.cpp` accepts `Authorization: Basic` natively (see `daemon/remote/WebServer.cpp` `ParseHeaders()`), so URL-embedded credentials were never required.

## Why not let `HttpClientHandler` handle redirects with its built-in safety?

Considered, but `HttpClientHandler.AllowAutoRedirect = true` is exactly what causes the original bug: it strips `Authorization` on every hop, unconditionally, so any 30x kills auth. The framework has no per-host re-apply hook. A `DelegatingHandler` with `AllowAutoRedirect = false` is the established pattern in this codebase (used in `TransmissionAdapter`, `TorrentFileDownloader`, `ImageCacheService`, `MyAnonamouseHelper`, all with similar comments about preserving auth/cookies across redirects).

## Test plan

* [x] `dotnet test --filter "FullyQualifiedName~NzbgetAdapterTests"` passes locally (13 tests including new redirect-flow coverage)
* [x] Full suite green: `dotnet test` → 641 passed, 0 failed
* [ ] Manual: configure NZBGet as a download client in Listenarr against an instance behind a reverse proxy that issues a same-host HTTPS upgrade; "Test" returns connected
* [ ] Manual: point at an attacker.example URL and confirm the cross-host error message surfaces correctly

### DIFF
--- a/listenarr.infrastructure/Adapters/NzbgetSafeRedirectHandler.cs
+++ b/listenarr.infrastructure/Adapters/NzbgetSafeRedirectHandler.cs
@@ -1,0 +1,231 @@
+/*
+ * Listenarr - Audiobook Management System
+ * Copyright (C) 2024-2026 Listenarr Contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+using System.Net;
+using Microsoft.Extensions.Logging;
+
+namespace Listenarr.Infrastructure.Adapters
+{
+    /// <summary>
+    /// Thrown by <see cref="NzbgetSafeRedirectHandler"/> when a redirect is refused (cross-host,
+    /// HTTPS->HTTP downgrade, missing Location, or loop). Carries a descriptive message that
+    /// the adapter surfaces to the user instead of the generic "network error" fallback.
+    /// </summary>
+    public sealed class NzbgetSafeRedirectException : HttpRequestException
+    {
+        public NzbgetSafeRedirectException(string message) : base(message) { }
+    }
+
+    /// <summary>
+    /// Manual redirect handler for the NZBGet HttpClient.
+    ///
+    /// The named "nzbget" HttpClient is configured with <c>AllowAutoRedirect = false</c> because
+    /// .NET's default redirect-follower strips <c>Authorization</c> headers across hops to prevent
+    /// credential leakage to unintended hosts. That protection breaks NZBGet setups where a reverse
+    /// proxy in front of NZBGet issues a 30x (typical cases: a Caddy/Nginx HTTPS upgrade, or
+    /// trailing-slash normalization on a configured base URL) — the redirect succeeds but the
+    /// subsequent request hits NZBGet without auth and returns 401 Unauthorized.
+    ///
+    /// This handler re-applies the <c>Authorization</c> header on a redirect, but only when both
+    /// safety rules pass:
+    ///   1. The redirect target is the same host:port as the original request.
+    ///   2. The redirect does not downgrade from HTTPS to HTTP.
+    ///
+    /// On any rule violation the handler throws a descriptive <see cref="HttpRequestException"/>
+    /// instead of silently dropping auth, so misconfigured proxies surface as actionable errors
+    /// rather than mysterious "Unauthorized" failures.
+    /// </summary>
+    public sealed class NzbgetSafeRedirectHandler : DelegatingHandler
+    {
+        internal const int MaxRedirects = 5;
+
+        private readonly ILogger<NzbgetSafeRedirectHandler>? _logger;
+
+        public NzbgetSafeRedirectHandler()
+        {
+        }
+
+        public NzbgetSafeRedirectHandler(ILogger<NzbgetSafeRedirectHandler> logger)
+        {
+            _logger = logger;
+        }
+
+        protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            // Buffer request content once so we can replay POST bodies across 307/308 redirects.
+            if (request.Content != null)
+            {
+                await request.Content.LoadIntoBufferAsync().ConfigureAwait(false);
+            }
+
+            var currentRequest = request;
+            HttpRequestMessage? clonedRequest = null;
+            var hops = 0;
+
+            try
+            {
+                while (true)
+                {
+                    var response = await base.SendAsync(currentRequest, cancellationToken).ConfigureAwait(false);
+
+                    if (!IsRedirectStatus(response.StatusCode))
+                    {
+                        return response;
+                    }
+
+                    if (hops >= MaxRedirects)
+                    {
+                        var sanitized = SanitizeUri(currentRequest.RequestUri);
+                        response.Dispose();
+                        throw new NzbgetSafeRedirectException(
+                            $"NZBGet request to {sanitized} hit the redirect cap of {MaxRedirects} hops. " +
+                            "This likely indicates a redirect loop in your reverse-proxy or NZBGet base-URL configuration.");
+                    }
+
+                    var location = response.Headers.Location;
+                    if (location == null)
+                    {
+                        var sanitized = SanitizeUri(currentRequest.RequestUri);
+                        var status = response.StatusCode;
+                        response.Dispose();
+                        throw new NzbgetSafeRedirectException(
+                            $"NZBGet responded {(int)status} {status} to {sanitized} but did not include a Location header. " +
+                            "Cannot follow the redirect; check the NZBGet (or proxy) server logs.");
+                    }
+
+                    var nextUri = location.IsAbsoluteUri
+                        ? location
+                        : new Uri(currentRequest.RequestUri!, location);
+
+                    if (!IsSameOriginHost(currentRequest.RequestUri!, nextUri))
+                    {
+                        var fromUri = SanitizeUri(currentRequest.RequestUri);
+                        var toUri = SanitizeUri(nextUri);
+                        response.Dispose();
+                        throw new NzbgetSafeRedirectException(
+                            $"NZBGet redirected from {fromUri} to a different host ({toUri}). " +
+                            "Authorization credentials would be forwarded to an unexpected host, so the redirect was blocked. " +
+                            "Fix the NZBGet base URL or reverse-proxy rewrite so the response is returned directly.");
+                    }
+
+                    if (IsSchemeDowngrade(currentRequest.RequestUri!, nextUri))
+                    {
+                        var fromUri = SanitizeUri(currentRequest.RequestUri);
+                        var toUri = SanitizeUri(nextUri);
+                        response.Dispose();
+                        throw new NzbgetSafeRedirectException(
+                            $"NZBGet redirected from HTTPS ({fromUri}) to HTTP ({toUri}). " +
+                            "Credentials would be sent in clear, so the redirect was blocked. " +
+                            "Configure NZBGet (or its reverse proxy) to serve only HTTPS.");
+                    }
+
+                    var nextRequest = BuildRedirectedRequest(currentRequest, response.StatusCode, nextUri);
+                    _logger?.LogDebug(
+                        "NZBGet redirect {Status} {From} -> {To} (hop {Hop}/{Cap})",
+                        (int)response.StatusCode,
+                        SanitizeUri(currentRequest.RequestUri),
+                        SanitizeUri(nextUri),
+                        hops + 1,
+                        MaxRedirects);
+
+                    response.Dispose();
+                    if (clonedRequest != null)
+                    {
+                        clonedRequest.Dispose();
+                    }
+                    clonedRequest = nextRequest;
+                    currentRequest = nextRequest;
+                    hops++;
+                }
+            }
+            catch
+            {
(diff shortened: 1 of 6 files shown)


T4g1:
I see in the adapter that it mention using JSON-RPC too but I can't find reference to call made using that instead of XML-RPC, at first glance I would suggest to make sure JSON-RPC is not used or that this change does not applies to it too

Also, i believe the auto redirect that strips the authorization does that to prevent credential leakage through malevolent redirects, this change breaks that security right ? Wouldn't it be better to define a custom redirection handler that validates the domain is the same and it's not downgrading (HTTPS -> HTTP) then add back the relevant headers instead ?

kevinheneveld:
Thanks @T4g1 — both points were on target and I've pivoted the PR accordingly. Quick summary of what changed:

**On the JSON-RPC code path:** you were right that it exists and was missed. `NzbgetAdapter.FetchDownloadsAsync` calls `/jsonrpc` directly via `BuildUri(client, "/jsonrpc")`, and `AddViaJsonRpcAsync` delegates through `CallXmlRpcAsync` so it's covered transitively. The pivoted approach below is registered at the HttpClient level (the named `"nzbget"` client), so both endpoints get the same treatment without per-call wiring.

**On the security concern about URL-embedded creds:** also right — `user:pass@host` survives redirects to *any* host (cross-domain, scheme-downgraded, anything), which defeats the protection the framework's redirect-strip is implementing rather than implementing it properly. I checked NZBGet's own `WebServer.cpp` (`ParseHeaders()`) and it accepts the `Authorization: Basic` header natively, so URL-embedded creds were never required — the original commit's justification was wrong. I also checked whether `/xmlrpc` actually 30x's on its own: it doesn't (only `/nzbget` redirects to `/nzbget/`), so the realistic 3xx case is a reverse proxy in front of NZBGet doing HTTPS upgrade or path normalization.

**Implementation:** new `NzbgetSafeRedirectHandler : DelegatingHandler` wired into the named `"nzbget"` HttpClient with `AllowAutoRedirect = false`. On 30x it validates same-host + no-HTTPS-downgrade before re-applying the `Authorization` header. On rule violation (cross-host, scheme downgrade, missing Location, or redirect loop) it throws a typed `NzbgetSafeRedirectException` with a descriptive message that `TestConnectionAsync` surfaces to the user, so misconfigurations are obvious instead of showing up as mystery `Unauthorized` errors.

PR body and tests rewritten — `[Theory]` over 301/302/303/307/308 for the re-apply path, plus negative tests for cross-host, downgrade, and loop. Force-pushed to `fix/nzbget-xmlrpc-auth`. Take another look when you have a moment.

tests/Features/Infrastructure/Adapters/NzbgetAdapterTests.cs
```diff
@@ -40,26 +48,42 @@ public TestHttpClientFactory(HttpClient client)
...
+        private static HttpClient BuildNzbgetClient(DelegatingHandlerMock leaf)
+        {
+            var redirectHandler = new NzbgetSafeRedirectHandler { InnerHandler = leaf };
+            return new HttpClient(redirectHandler);
+        }
+
+        private static NzbgetAdapter BuildAdapter(HttpClient http) =>
+            new(new TestHttpClientFactory(http), Mock.Of<INzbUrlResolver>(), NullLogger<NzbgetAdapter>.Instance);
+
+        private static HttpResponseMessage OkXml(string body) => new(HttpStatusCode.OK)
```
T4g1: Why not use GetCannedResponse from MockUtils ?

tests/Features/Infrastructure/Adapters/NzbgetAdapterTests.cs
```diff
@@ -40,26 +48,42 @@ public TestHttpClientFactory(HttpClient client)
             public HttpClient CreateClient(string name) => _client;
         }
 
+        // Wrap the safe-redirect handler around a DelegatingHandlerMock so unit tests exercise
+        // the production redirect pipeline. Matches how the named "nzbget" HttpClient is configured
+        // in Program.cs / ServiceRegistrationExtensions.
+        private static HttpClient BuildNzbgetClient(DelegatingHandlerMock leaf)
```
T4g1, line 54: Can't we re-use the NzbgetApiMock instead of creating a new one for each test ?

listenarr.infrastructure/Adapters/NzbgetSafeRedirectHandler.cs
```diff
@@ -0,0 +1,231 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Listenarr.Infrastructure.Adapters
+{
+    /// <summary>
+    /// Thrown by <see cref="NzbgetSafeRedirectHandler"/> when a redirect is refused (cross-host,
+    /// HTTPS->HTTP downgrade, missing Location, or loop). Carries a descriptive message that
+    /// the adapter surfaces to the user instead of the generic "network error" fallback.
+    /// </summary>
+    public sealed class NzbgetSafeRedirectException : HttpRequestException
```
T4g1: I'm not a huge fan of having multiple classes in one file, @therobbiedavis we should probably define guideline for this, wdyt ?

listenarr.infrastructure/Adapters/NzbgetSafeRedirectHandler.cs
```diff
@@ -0,0 +1,231 @@
...
+            {
+                if (string.Equals(header.Key, "Authorization", StringComparison.OrdinalIgnoreCase)) continue;
+                if (string.Equals(header.Key, "Host", StringComparison.OrdinalIgnoreCase)) continue;
+                next.Headers.TryAddWithoutValidation(header.Key, header.Value);
+            }
+
+            return next;
+        }
+
+        private static bool IsRedirectStatus(HttpStatusCode code) => code switch
```
T4g1: We have duplicated code with this in some other areas already, lets make this a util somewhere

listenarr.infrastructure/Adapters/NzbgetSafeRedirectHandler.cs
```diff
@@ -0,0 +1,231 @@
...
+        {
+            HttpStatusCode.MovedPermanently => true,   // 301
+            HttpStatusCode.Found => true,              // 302
+            HttpStatusCode.SeeOther => true,           // 303
+            HttpStatusCode.TemporaryRedirect => true,  // 307
+            HttpStatusCode.PermanentRedirect => true,  // 308
+            _ => false
+        };
+
+        private static bool IsSameOriginHost(Uri from, Uri to) =>
```
T4g1: The following methods also sounds like they are not restricted to NZBGet context, maybe those should move too to avoid AI duplicating code everywhere

listenarr.infrastructure/Adapters/NzbgetSafeRedirectHandler.cs
```diff
@@ -0,0 +1,231 @@
...
+    {
+        internal const int MaxRedirects = 5;
+
+        private readonly ILogger<NzbgetSafeRedirectHandler>? _logger;
+
+        public NzbgetSafeRedirectHandler()
+        {
+        }
+
+        public NzbgetSafeRedirectHandler(ILogger<NzbgetSafeRedirectHandler> logger)
```
T4g1: Let's remove the test only empty constructor and solve https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0290 here (test should not drive the implementation details)

listenarr.infrastructure/Adapters/NzbgetSafeRedirectHandler.cs
```diff
@@ -0,0 +1,231 @@
...
+
+            var currentRequest = request;
+            HttpRequestMessage? clonedRequest = null;
+            var hops = 0;
+
+            try
+            {
+                while (true)
+                {
+                    var response = await base.SendAsync(currentRequest, cancellationToken).ConfigureAwait(false);
```
T4g1, line 65: Why not use the using keyword instead of manual dispose down the line ?

listenarr.infrastructure/Adapters/NzbgetSafeRedirectHandler.cs
```diff
@@ -0,0 +1,231 @@
...
+
+                    if (!IsRedirectStatus(response.StatusCode))
+                    {
+                        return response;
+                    }
+
+                    if (hops >= MaxRedirects)
```
T4g1: This should be the while loop guard probably

T4g1 (changes requested):

kevinheneveld:
Thanks @T4g1 — worked through these in `0ec9f8c7` (additive commit, no force-push). Point by point:

**Handler**
- **`IsRedirectStatus` duplication** → removed; the handler now calls the existing `OutboundRequestSecurity.IsRedirectStatusCode` (same 301/302/303/307/308 set).
- **Non-NZBGet helpers** → moved `IsSameOriginHost`/`IsSchemeDowngrade` into `OutboundRequestSecurity` as `IsSameHostAndPort`/`IsHttpsToHttpDowngrade`, and dropped the bespoke `SanitizeUri` in favour of the existing `LogRedaction.SanitizeUrl` (which also strips the query string).
- **Test-only ctor / IDE0290** → replaced both constructors with a primary constructor taking a non-nullable `ILogger`. DI always supplies it; tests pass `NullLogger`. No more test-driven empty ctor.
- **`using` instead of manual dispose** → redirect responses are now scoped with `using`; the terminal (returned) response stays owned by the caller.
- **Loop guard** → the redirect cap is now the loop bound (`for (hops = 0; hops <= MaxRedirects; hops++)`) rather than an `if` inside `while(true)`. Behaviour is unchanged (original request + up to 5 follows); I added an `Assert.Equal(6, requestCount)` to the loop-bail test to pin that and catch any `<`-vs-`<=` regression.
- **Multiple classes in one file** → split `NzbgetSafeRedirectException` into its own file. I went with one-class-per-file as the safe default, but happy to follow whatever guideline you and @therobbiedavis settle on.

**Tests**
- **`GetCannedResponse`** → done; dropped the local `OkXml` helper and use `MockUtils.GetCannedResponse(body, "text/xml")`.
- **Reuse `NzbgetApiMock`** → I looked at this but I'm not sure it fits here: every test in this file uses `DelegatingHandlerMock` specifically to (a) **capture** the outgoing request (asserting URI normalization, the absence of URL-embedded creds, the re-applied `Authorization` header) and (b) **sequence** responses (redirect on hop 1, then 200). `NzbgetApiMock` is a route-based canned-response mock — it doesn't capture the request or let a test return a redirect-then-OK sequence, so I don't think it can drive these cases. Did you have a particular shape in mind (e.g. centralizing the canned version/history XML bodies), or is the capturing handler acceptable for this file?

Full suite green (685/685).

listenarr.infrastructure/Adapters/NzbgetSafeRedirectHandler.cs
```diff
@@ -74,83 +54,77 @@ protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage
 
             var currentRequest = request;
             HttpRequestMessage? clonedRequest = null;
```
T4g1, line 56: I don't get the purpose of this clonedRequest ?

tests/Features/Infrastructure/Adapters/NzbgetAdapterTests.cs
T4g1: To answer back at your reply: 
* capture the outgoing request: We inherit BaseApiMock so NzbgetApiMock has access to the last request made, isn't that enough ?
*  For the ones only defining redirection, I agree, we can keep them

T4g1 (changes requested):

listenarr.infrastructure/Adapters/NzbgetSafeRedirectHandler.cs
T4g1: We still have the exact same method in ImageCacheService

therobbiedavis:
Heads up: #599 restructures NZBGet into focused infrastructure workflows/helpers, so this PR will need a rebase and likely a port of the redirect/auth behavior into the new NZBGet request/XML-RPC path.

The underlying fix may still be valid, but it should be reviewed against the post-#599 NZBGet structure rather than the old adapter layout.

kevinheneveld:

PR #583, opened 2026-05-12
feat: add opt-in full metadata rescan for existing audiobooks

kevinheneveld:
Summary
- Library Import gets an opt-in full metadata rescan that queues library-wide backfill jobs.
- Edit Audiobook gets a "Fill missing from file" action for force-refresh scans on a single book.
- Scan behavior now distinguishes library-wide backfill safety from explicit per-book cleanup.

Validation
- Backend: 642 tests
- Frontend: 348 tests
- Live smoke: deployed on media and verified HTTP 200 on / and /library-import

Note
- This branch also contains the separate Dockerfile runtime-permissions fix commit (8a65297) used to keep the live container readable during deploy.


fe/src/components/domain/audiobook/EditAudiobookModal.vue
```diff
@@ -1067,6 +1082,27 @@ async function resolveAudiobookForEditing(audiobook: Audiobook): Promise<Audiobo
...
+async function fillMissingMetadata() {
+  const audiobook = props.audiobook
+  if (!audiobook?.id || fillingMissingMetadata.value) return
+
+  fillingMissingMetadata.value = true
+  try {
+    await apiService.scanAudiobook(audiobook.id, undefined, true)
+    toast.success(
+      'Metadata refresh enqueued',
+      'Scanning files for missing metadata. Reopen this audiobook to see the updated fields.',
```
T4g1, line 1094: We probably want to leverage SignalR for that instead

fe/src/views/library/LibraryImportView.vue
```diff
@@ -322,7 +335,26 @@ async function onFolderChange() {
 
 async function startScan() {
   if (!selectedFolderId.value) return
-  await store.triggerScan(selectedFolderId.value)
+
+  const scanPromise = store.triggerScan(selectedFolderId.value)
+
+  if (fullRescanMetadata.value) {
```
T4g1, line 341: I'm not sure about the use case for this feature here, as a user, i probably do not expect this feature to be here, wdyt ?

Also, now when we start scan, the scan.status is no longer the valid source of status for the ongoing scans which feels difficult to maintain eventualy

listenarr.api/Services/AudioFileService.cs
```diff
@@ -38,7 +38,7 @@ public AudioFileService(IServiceScopeFactory scopeFactory, ILogger<AudioFileServ
             _limiter = limiter;
         }
 
-        public async Task<bool> EnsureAudiobookFileAsync(int audiobookId, string filePath, string? source = "scan")
+        public async Task<bool> EnsureAudiobookFileAsync(int audiobookId, string filePath, string? source = "scan", bool forceMetadataRefresh = false)
```
T4g1, line 41: About separation of concerns here: 
* The purpose of this method is to link a file to an audiobook
* Coincidentaly, it also creates it when it does not exist and determines the metadata
So: Metadata logic should probably be extracted from this in a method that takes an AudiobookFile and refresh it's metadata (with a default strategy of backfilling) so we can drop the forceMetadataRefresh

listenarr.api/Services/AudioFileService.cs
```diff
@@ -293,6 +307,180 @@ public async Task<bool> EnsureAudiobookFileAsync(int audiobookId, string filePat
             }
         }
 
+        /// <summary>
+        /// Re-extracts metadata from an already-tracked file and backfills any blank library-level
+        /// fields on the parent audiobook record. Used by the force-refresh scan flow to enrich
+        /// existing books without touching their AudiobookFile rows.
+        /// </summary>
+        private async Task RefreshAudiobookMetadataFromFileAsync(
+            int audiobookId,
```
T4g1, line 316: This should take an AudiobookFile as input instead given the commentary and the comment should probably not infer its usage location as this is a service that could be used elsewhere

listenarr.api/Services/AudioFileService.cs
T4g1: I see that you already defined RefreshAudiobookMetadataFromFileAsync, that could be used instead inconditionaly once the audiobook file has been created here

listenarr.api/Services/AudioFileService.cs
```diff
@@ -293,6 +307,180 @@ public async Task<bool> EnsureAudiobookFileAsync(int audiobookId, string filePat
...
+                identifiersChanged = true;
+                anyChange = true;
+            }
+
+            return anyChange;
+        }
+
+        // Splits only on unambiguous list separators. Commas are excluded because
+        // "Last, First"-style single-author tags are common and would split into two authors.
+        private static List<string> SplitListTag(string value) =>
```
T4g1, line 473: This could probably be made into a non specific method by allowing to give the list of separator as a parameter and move it into StringUtils probably

listenarr.api/Services/AudioFileService.cs
```diff
@@ -293,6 +307,180 @@ public async Task<bool> EnsureAudiobookFileAsync(int audiobookId, string filePat
...
+            value.Split(new[] { ';', '/' }, StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries)
+                .Where(s => !string.IsNullOrWhiteSpace(s))
+                .ToList();
+
+        private static string? FirstNonEmpty(params string?[] candidates)
```
T4g1, line 478: Same remark, this is not AudiobookFile specific

listenarr.api/Services/ScanBackgroundService.cs
```diff
@@ -333,8 +352,10 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
                                     using var afScope = _scopeFactory.CreateScope();
                                     var audioFileService = afScope.ServiceProvider.GetRequiredService<IAudioFileService>();
 
-                                    // Store absolute path - metadata extraction needs full path
-                                    var created = await audioFileService.EnsureAudiobookFileAsync(audiobook.Id, filePath, "scan");
+                                    // Store absolute path - metadata extraction needs full path.
+                                    // forceMetadataRefresh=true causes already-tracked files to re-run promotion
+                                    // and backfill blank audiobook fields.
+                                    var created = await audioFileService.EnsureAudiobookFileAsync(audiobook.Id, filePath, "scan", job.ForceMetadataRefresh);
```
T4g1, line 358: We should split the call here for separation of concerns:
* Keep the original EnsurreAudiobookFile
* Call RefreshMetadata explicitely with the proper (created or existing) audiobook file instead

listenarr.api/Services/ScanBackgroundService.cs
```diff
@@ -107,6 +107,25 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
 
                             if (usedBasePath && (string.IsNullOrEmpty(scanRoot) || !Directory.Exists(scanRoot)))
                             {
+                                if (job.SkipMissingBasePathCleanup)
```
T4g1, line 110: Audiobook base path are soon to be mandatory, so im not sure we should add specific logic assuming they are not

listenarr.api/Services/ScanQueueService.cs
```diff
@@ -31,40 +31,46 @@ public ScanQueueService(ILogger<ScanQueueService> logger)
             _logger = logger;
         }
 
-        public async Task<Guid> EnqueueScanAsync(Audiobook audiobook, string? path = null)
+        public async Task<Guid> EnqueueScanAsync(Audiobook audiobook, string? path = null, bool forceMetadataRefresh = false, bool skipMissingBasePathCleanup = false)
```
T4g1, line 34: In summary, i believe:
* skipMissingBasePathCleanup will be outdated very soon
* forceMetadataRefresh should be the default behavior or at least, we can probably drop the "forced" in it

listenarr.api/Services/Metadata/MetadataService.cs
```diff
@@ -263,6 +263,46 @@ public Task WriteAsinTagAsync(string filePath, string asin)
...
+                return Task.FromResult<(byte[]?, string?)>((picture.Data.Data.ToArray(), ext));
+            }
+            catch (Exception ex) when (ex is not OperationCanceledException && ex is not OutOfMemoryException && ex is not StackOverflowException)
+            {
+                _logger.LogDebug(ex, "Could not extract embedded cover from {File}", LogRedaction.SanitizeFilePath(filePath));
+                return Task.FromResult<(byte[]?, string?)>((null, null));
+            }
+        }
+
+        private static string? InferExtensionFromMagicBytes(byte[]? bytes)
```
T4g1, line 295: This should probably be part of the infrastructure layer instead

listenarr.api/Services/Metadata/MetadataService.cs
```diff
@@ -263,6 +263,46 @@ public Task WriteAsinTagAsync(string filePath, string asin)
             }
         }
 
+        public Task<(byte[]? Bytes, string? Extension)> ExtractEmbeddedCoverAsync(string filePath)
+        {
+            try
```
T4g1, line 268: We should make sure to have settings.EnableMetadataProcessing set

T4g1 (changes requested):
While doing the review, Im wondering, why do we need to explicitly ask for the backfill ? Do we want a scan that does not update the metadata ? If so, why would we want to scan in the first place ? Shouldn't we backfill by default when user issues a re-scan ? Depending on which path we take here, this may probably be simplified

listenarr.api/Services/Metadata/MetadataService.cs
T4g1: Note about this comment: In the current state of the repository, if you do move that to the infrastructure layer, you will have layer violation exceptions as MetadataService should not be in the API layer, so you can probably ignore this (i leave it open because depending on when we merge #576 this might become relevant again)

therobbiedavis:
> While doing the review, Im wondering, why do we need to explicitly ask for the backfill ? Do we want a scan that does not update the metadata ? If so, why would we want to scan in the first place ? Shouldn't we backfill by default when user issues a re-scan ? Depending on which path we take here, this may probably be simplified

@kevinheneveld Could you help explain the use-case here?

kevinheneveld:
Good question. The original thinking was to give users an escape hatch in case they'd hand-edited metadata and didn't want a file re-scan to clobber it. But that rationale doesn't actually hold: `RefreshAudiobookMetadataFromFileAsync` only fills *blank* library fields — every assignment is guarded against overwriting existing values — so backfill is already non-destructive by design. There's no scenario where defaulting it loses user edits.

Given that, the opt-in flag is just noise. I'd rather close this PR and revisit metadata-on-rescan as a smaller, focused change once the foundational work lands — at which point `forceMetadataRefresh` / `skipMissingBasePathCleanup` are likely obsolete anyway (per the other review notes), and the separation-of-concerns refactoring in `AudioFileService` would be better done against the new structure than retrofitted here. Thanks for the thorough review — it'll carry over.

PR #585, opened 2026-05-13
Add list view + per-grouping persistence for the audiobooks library

### LINKED ISSUE #595: [Enhancement] List view missing for ?group=authors and ?group=series
## Summary
The audiobooks toolbar exposes a list/grid view toggle, but the grouped views (`?group=authors` and `?group=series`) only render as a grid. Toggling to list mode while grouped is silently ignored — the toolbar icon flips, but the page keeps rendering grid cards.

Root cause: the list-mode template in `fe/src/views/library/AudiobooksView.vue` is nested inside the `groupBy === 'books'` branch. The grouped-view branch has no `viewMode === 'list'` rendering at all.

## Steps to reproduce
1. Open the audiobooks library.
2. Click the toolbar list/grid toggle so the icon shows list mode.
3. Open the grouping dropdown and switch to Authors (or Series).
4. The view stays as a grid of cards. The list-mode toggle has no effect for grouped views.

## Expected behavior
List mode should render a row-per-collection list (cover thumb, collection name, book count). Clicking a row should navigate to the collection page, matching how the grid cards behave.

## Notes
Re-opened here from kevinheneveld/Listenarr#1 per discussion on #585 — proposed fix is in that PR.

kevinheneveld:
## Summary

Two related changes to the audiobooks library's grouped views, combined into one PR (was previously split across #585 and #590; #590 is being closed in favor of this).

**1. List view for grouped authors and series.** The audiobooks toolbar exposes a list/grid view toggle, but the grouped views (`?group=authors`, `?group=series`) only had a grid template. Toggling list mode while grouped was silently a no-op. This PR adds a parallel list rendering for grouped collections: one row per collection (cover thumb, collection name, book count). Clicking a row navigates to the collection page, matching the grid-card behavior.

**2. Per-grouping persistence of the view mode.** Previously the load-from-storage was nested inside an `if (scrollContainer.value)` block that only runs for the Books grouping — so switching to list mode while grouped by Author or Series, then navigating away, would silently revert to grid. This switches to three per-grouping keys (`listenarr.viewMode.books`, `.authors`, `.series`) and reads the relevant one unconditionally on mount + whenever `groupBy` changes. A one-time migration seeds all three keys from the legacy `listenarr.viewMode` value so existing users keep their preference.

## Implementation notes

- Template change is localized to the grouped-view block in `fe/src/views/library/AudiobooksView.vue`. The existing grid is wrapped in `v-else`; new `v-if=\"viewMode === 'list'\"` branch above it.
- CSS reuses the existing `.audiobooks-list` / `.audiobook-list-item` patterns; collection rows have their own three-column grid template (cover / name / count) since they don't carry status badges or per-item actions.
- Persistence watch moved out from inside the scrollContainer guard to a top-level watch, so it fires regardless of grouping. Removed the now-unused \`stopPersistViewModeWatch\` cleanup handle.
- Initial viewMode is set inline at declaration: \`ref<...>(loadViewModeFor(groupBy.value))\`.
- Migration is additive — the legacy key is never deleted, just no longer read once the per-grouping keys exist.
- \`viewMode\` and \`toggleViewMode\` are exposed via \`defineExpose\` so tests can drive the view deterministically.

## Test plan

- [x] \`cd fe && npm run test:unit\` — 14 / 14 AudiobooksView tests passing (4 new: list-view rendering for authors + series, per-grouping persistence, legacy migration)
- [x] \`cd fe && npm run type-check\` — clean
- [x] Rebased onto canary tip (post-1.0.1); resolved one spec-file conflict (type-field adjacency) and re-ported the persistence change onto the new initializeVirtualScroller refactor

Closes #595
Supersedes #590


T4g1:
Small note regarding the issue linked to this: Is there a particular reason why you created it on the fork instead of here ? It might be better to open them here instead if you expect feedback on those (or perhaps there is a way to see forked repository issues from here ?)

kevinheneveld:
Good point. My workflow tracks issues on the fork as a working backlog, but you're right that anything expecting upstream feedback belongs here. I'll re-open this one on Listenarrs/Listenarr and relink it.

kevinheneveld:
Re-filed the linked issue here: #595 (was kevinheneveld/Listenarr#1). Will update the PR body to point at the new one.

fe/src/views/library/AudiobooksView.vue
```diff
@@ -211,7 +211,55 @@
...
+          @keydown.space.prevent="navigateToCollection(collection)"
+        >
+          <img
+            class="list-thumb"
+            :src="
+              getProtectedImageSrc(
+                groupBy === 'authors'
+                  ? getAuthorImageUrl(collection)
+                  : collection.coverUrls && collection.coverUrls[0],
+                `${groupBy}-list:${collection.name}`,
```
therobbiedavis: getProtectedImageSrc() treats its second argument as the actual fallback image URL, not as an image/cache key. With the current value, coverless author/series rows get a src like authors-list:Author A or series-list:Series 1 until the image error handler replaces it. Could this pass getPlaceholderUrl() instead?

fe/src/views/library/AudiobooksView.vue
```diff
@@ -211,7 +211,55 @@
...
+              ) || getPlaceholderUrl()
+            "
+            :alt="collection.name"
+            loading="lazy"
+            decoding="async"
```
therobbiedavis, line 256: This new author list path bypasses the existing author-cover lookup behavior. Grid author cards call into ensureAuthorCover() through handleAuthorImageError() and the observeAuthorCards() marker selector at AudiobooksView.vue:1519, but these list rows only use the generic handleImageError and do not render a data-author-name target. For authors without authorAsins, list mode will stay on the placeholder instead of fetching the lookup cover.

I’d avoid fixing this only by swapping @error to handleAuthorImageError: once my previous comment changes the fallback to a valid placeholder, no error fires for missing covers, so lookup still won’t happen. A safer fix would make list author rows participate in the same lazy observer path, and add a test where an author has no authorAsins but apiService.getAuthorLookup() returns a cover.

therobbiedavis (changes requested):
Thanks for taking care of this; it was actually on my list of todo's eventually. I left a couple of inline comments around the new grouped-list image path: the fallback argument is currently not a real image URL, and the author list rows do not participate in the existing author-cover lookup flow used by grid cards.

kevinheneveld:
Thanks @therobbiedavis — both addressed in `31aeaf8b` (additive commit, no force-push):

- **Fallback URL:** the list thumb now passes `getPlaceholderUrl()` as the `getProtectedImageSrc` fallback instead of the `${groupBy}-list:${name}` cache-key string.
- **Author-cover lookup in list mode:** rather than just swapping `@error` (which, as you noted, won't fire once the fallback is a valid placeholder), the author list rows now participate in the same lazy observer path as grid cards — the list thumb carries `data-author-name`/`data-author-has-cover`, and the `observeAuthorCards()` selector now matches any `[data-author-name]` within `.author-collection` (decoupled from the grid-only `.audiobook-poster-container` layout class so the list row doesn't inherit grid positioning). I also re-run `observeAuthorCards()` on a view-mode switch, since the `viewMode` watch wasn't re-observing and grid/list render different markers.
- **Test:** added a case where an author with no `authorAsins` whose `getAuthorLookup()` returns a cover gets it applied in list view.

s3ntin3l8:
Confirmed this PR on my running system (since I noticed the same issue and it kept bothering me). Everything seems in order.

kevinheneveld:

PR #589, opened 2026-05-14
Show ready / total book counts with quality-color cue on grouped collections

kevinheneveld:
## Summary
- For audiobook collections grouped by author or series, the count column previously showed just the library total (e.g. `12 books`). This adds a ready / total breakdown — \`8 / 12 books\` — where the numerator counts books with a playable file present (statuses `quality-match` or `quality-mismatch`).
- The numerator is colored:
  - **Green** when every present file meets the quality profile.
  - **Orange** when at least one file is below the profile but still playable.
  - **Neutral** when no files are present yet.
- Applied to both the new list-view row and the grid-card bottom placard (visible when item-details is enabled). The small corner count badge on grid cards stays as just the library total to keep the visual uncluttered.
- Build/PR note: **stacked on top of #585** (the list-view rendering for grouped collections). #585 should be merged first; otherwise the list-view template this PR colors doesn't exist yet.

## Implementation notes
- New per-collection fields on `groupedCollections`: `readyCount` and `qualityMismatchCount`. Computed inline by calling `computeAudiobookStatus` (the pure helper in `fe/src/utils/audiobookStatus.ts`).
- `qualityProfiles` ref and `activeDownloadAudiobookIds` computed are moved earlier in the file so `groupedCollections` can depend on them. The status-cache computed (`audiobookStatusById`) declared further down stays where it is — `groupedCollections` calls the pure function directly to avoid a temporal-dead-zone error during initial watcher setup.
- Colors reuse the existing `#2ecc71` / `#f39c12` palette from the audiobook-status legend.

## Test plan
- [x] `cd fe && npm run test:unit` — 351 / 351 passing (351 new test asserts `readyCount` is populated from books-with-files data)
- [x] `npx vue-tsc --noEmit -p tsconfig.app.json` — clean
- [x] `cd tests && dotnet test` — passing (no backend changes)

Closes kevinheneveld/Listenarr#2

PR #590, opened 2026-05-14
Persist audiobooks list/grid view mode per grouping

kevinheneveld:
## Summary
- The audiobooks library toolbar exposes a list/grid view toggle, persisted to localStorage. Previously the load-from-storage was nested inside an \`if (scrollContainer.value)\` block that only runs for the Books grouping — so switching to list mode while grouped by Author or Series, then navigating away, would silently revert to grid.
- This switches to three per-grouping keys (\`listenarr.viewMode.books\`, \`.authors\`, \`.series\`) and reads the relevant one unconditionally on mount + whenever \`groupBy\` changes. So each grouping independently remembers list vs grid — useful for users who want, e.g., a list of authors but cards for books.
- A one-time migration seeds all three keys from the legacy \`listenarr.viewMode\` value the first time it's read, so existing users keep their preference.

## Implementation notes
- Persistence watch moved out from inside \`onMounted\`'s scrollContainer guard to a top-level watch, so it fires regardless of grouping.
- Removed the now-unused \`stopPersistViewModeWatch\` cleanup handle.
- Initial viewMode is set inline at declaration: \`ref<...>(loadViewModeFor(groupBy.value))\`.
- Migration is additive — the legacy key is never deleted, just no longer read once the per-grouping keys exist.

## Test plan
- [x] \`cd fe && npm run test:unit\` — 350 / 350 passing (2 new tests: per-grouping persistence + legacy migration)
- [x] \`npx vue-tsc --noEmit -p tsconfig.app.json\` — clean
- [x] \`cd tests && dotnet test\` — passing (no backend changes)

Closes kevinheneveld/Listenarr#3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.2.72]
+
+### Added
+- **List view for grouped authors and series:** Switching to list view (`viewMode === 'list'`) while grouping the audiobooks library by author or series now renders a row-per-collection list instead of silently falling back to the grid. Each row shows the collection cover, name, and book count, and clicking it navigates to the collection page — matching how the list-mode toggle already works for the Books grouping.
+
 ## [0.2.71] - 2026-04-17
 
 ### Added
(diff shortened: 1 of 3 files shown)


therobbiedavis:
@kevinheneveld Hey there! I appreciate the PR's however I do want to let you know that we're focusing on some foundational changes first and you'll need to rebase your PRs after those are merged, so you may not want to open too many at this time.

kevinheneveld:
Got it, sorry — opened #591 (a bug fix for automatic search matching music albums) before I saw this note. Happy to wait. I'll convert these to draft so they're out of your review queue until the foundational changes land, and I'll rebase them then. No rush from my end.

kevinheneveld:
Quick update — all five of my open PRs (#580, #585, #589, #590, #591) have been rebased onto canary (v0.4.1). #585, #589, #590, #591 are still drafted per your earlier note; #580 is the nzbget auth fix and is ready when you have time. Whenever the foundation phase feels open enough for the feature PRs again, happy to undraft any/all of them. No urgency.

kevinheneveld:
Closing in favor of #585, which I've expanded to carry both the grouped list view and the per-grouping view-mode persistence as two commits on a single branch. They're related enough that splitting them into two PRs was creating churn (rebase conflicts on the same file) without giving you cleaner review units — happy to re-split if you'd prefer them separate. #585 is rebased onto current canary.

PR #591, opened 2026-05-14
fix(search): reject automatic-search results whose title is irrelevant to the audiobook

kevinheneveld:
## Summary

Adds a context-aware relevance filter to the automatic-search pipeline so it stops queuing downloads that share a single common token with the audiobook but are otherwise unrelated — e.g. a Patterson Hood concert recording matched to a James Patterson audiobook on the single shared surname. Conservative: rejects only on explicit signal, fails open otherwise, and manual search keeps the old behavior so users browsing results still see everything.

### Added
- New `RelevanceFilter` (under `Search/Filters/`): stop-word-filtered token overlap between the result title and `audiobook.Title + audiobook.Authors`, rejecting below 30% (`DefaultMinRelevance`). Without an audiobook context it fails open.
- **Author corroboration for generic single-word titles:** a 1-token title with a 2-token author clears the 30% ratio on the title word alone (1/3 ≈ 0.33), so a different book whose title merely *contains* that common word would slip through (e.g. a release whose title contains "Betrayed" matching the wanted "Betrayed" by Lindsay Buroker). For such generic titles the result must also share an author token, else it's rejected; distinctive multi-word titles and author-less books fail open. (This folds in a refinement made since the original approval — see Notes.)
- New `SignificantTokens` helper — the shared stop-word-filtered tokenizer the filter uses.

### Changed
- `ISearchResultFilter.ShouldFilter` gains an optional `Audiobook? audiobook = null` context; `SearchResultFilterPipeline.ApplyFilters` / `WouldFilter` thread it through to each filter. Existing filters take the new parameter but ignore it (only `RelevanceFilter` uses it).
- The two auto-pick flows — `AutomaticSearchService` (background sweep) and `DownloadService.SearchAndDownloadAsync` — invoke the pipeline with the audiobook context after raw search and before scoring.

## Testing
- `SearchRelevanceFilterTests` (9 tests): the bug-report scenario (Patterson Hood vs Dog Diaries), legitimate matches (David Copperfield by Charles Dickens, author-first and title-first naming), case/punctuation insensitivity, fail-open when the audiobook has no significant tokens, and no-audiobook-context backwards-compat.
- Full backend suite passes (1024/1024); `dotnet format --verify-no-changes` clean.

## Notes
- **Rebased onto canary v1.2.1 and re-homed onto the refactored search subsystem** (interface now in `Search/Contracts/`, DI in `ListenarrWorkflowRegistration`, services relocated to `Downloads/Submission/` and `infrastructure/HostedServices/Search/`). The diff shape changed from the originally-approved version, so a fresh look is welcome.
- This now includes the **author-corroboration refinement** developed and battle-tested on the live instance after the original approval (it caught additional single-word-title false positives like "Betrayed" / "The Vision"). Folded in here so upstream gets the version actually in production rather than the earlier draft.
- The old `LibraryController` auto-pick edit is **dropped** — that flow was extracted into `DownloadService.SearchAndDownloadAsync` during the refactor, which this PR already filters. The only other `ScoreSearchResults` caller is a quality-profile preview endpoint, which intentionally isn't filtered.
- **Not included** (dropped from the original draft, unchanged rationale): the Newznab category-3030 restriction (canary no longer applies it; per-indexer category config is the right place) and a hard language filter (canary's `SearchResultScorer` already penalizes language mismatches).
- `CHANGELOG.md` change removed — the project no longer tracks it on canary.

### DIFF
--- a/listenarr.api/Startup/ListenarrWorkflowRegistration.cs
+++ b/listenarr.api/Startup/ListenarrWorkflowRegistration.cs
@@ -61,6 +61,7 @@ public static class ListenarrWorkflowRegistration
         services.AddScoped<ISearchResultFilter, PromotionalTitleFilter>();
         services.AddScoped<ISearchResultFilter, ProductLikeTitleFilter>();
         services.AddScoped<ISearchResultFilter, MissingInformationFilter>();
+        services.AddScoped<ISearchResultFilter, RelevanceFilter>();
         services.AddScoped<SearchResultFilterPipeline>();
         services.AddScoped<AsinCandidateCollector>();
         services.AddScoped<AsinEnricher>();
--- a/listenarr.application/Downloads/Submission/DownloadService.cs
+++ b/listenarr.application/Downloads/Submission/DownloadService.cs
@@ -17,6 +17,7 @@
  */
 
 using Listenarr.Application.Common;
+using Listenarr.Application.Search.Filters;
 using Microsoft.Extensions.Logging;
 
 namespace Listenarr.Application.Downloads.Submission
@@ -28,6 +29,7 @@ namespace Listenarr.Application.Downloads.Submission
         ILogger<DownloadService> logger,
         IQualityProfileService qualityProfileService,
         ISearchService searchService,
+        SearchResultFilterPipeline filterPipeline,
         IDownloadClientGateway clientGateway,
         IDownloadQueueService downloadQueueService,
         INotificationService notificationService,
@@ -158,6 +160,8 @@ namespace Listenarr.Application.Downloads.Submission
                 };
             }
 
+            // Reject results irrelevant to this audiobook before scoring (an empty set falls through below).
+            searchResults = filterPipeline.ApplyFilters(searchResults, audiobook: audiobook);
             // Score results against quality profile
             var scoredResults = await qualityProfileService.ScoreSearchResults(searchResults, audiobook.QualityProfile);
 
--- a/listenarr.application/Search/Contracts/ISearchResultFilter.cs
+++ b/listenarr.application/Search/Contracts/ISearchResultFilter.cs
@@ -28,8 +28,14 @@ public interface ISearchResultFilter
     /// Determines if the result should be filtered out (excluded).
     /// </summary>
     /// <param name="result">The search result to evaluate</param>
+    /// <param name="audiobook">
+    /// The audiobook the search is for, when known. Context-aware filters (e.g.
+    /// relevance) use it to judge a result against a specific target; filters that
+    /// don't need it ignore it. When null (manual search, per-result enrichment),
+    /// context-aware filters must fail open and keep the result.
+    /// </param>
     /// <returns>True if the result should be filtered out, false to keep it</returns>
-    bool ShouldFilter(SearchResult result);
+    bool ShouldFilter(SearchResult result, Audiobook? audiobook = null);
 
     /// <summary>
     /// Reason why the result was filtered (for logging/debugging).
--- a/listenarr.application/Search/Filters/AudiobookOnlyFilter.cs
+++ b/listenarr.application/Search/Filters/AudiobookOnlyFilter.cs
@@ -27,7 +27,7 @@ public class AudiobookOnlyFilter : ISearchResultFilter
 {
     public string FilterReason => "non_audiobook_filtered";
 
-    public bool ShouldFilter(SearchResult result)
+    public bool ShouldFilter(SearchResult result, Audiobook? audiobook = null)
     {
         // If enriched with a metadata source, prefer that metadata only when the
         // metadata source is a trusted audio provider or the enriched metadata
--- a/listenarr.application/Search/Filters/KindleEditionFilter.cs
+++ b/listenarr.application/Search/Filters/KindleEditionFilter.cs
@@ -25,7 +25,7 @@ public class KindleEditionFilter : ISearchResultFilter
 {
     public string FilterReason => "kindle_edition_filtered";
 
-    public bool ShouldFilter(SearchResult result)
+    public bool ShouldFilter(SearchResult result, Audiobook? audiobook = null)
     {
         return SearchValidation.IsKindleEdition(result.Title);
     }
--- a/listenarr.application/Search/Filters/MissingInformationFilter.cs
+++ b/listenarr.application/Search/Filters/MissingInformationFilter.cs
@@ -25,7 +25,7 @@ namespace Listenarr.Application.Search.Filters
     {
         public string FilterReason => "missing_author_or_title";
 
-        public bool ShouldFilter(SearchResult result)
+        public bool ShouldFilter(SearchResult result, Audiobook? audiobook = null)
         {
             return string.IsNullOrWhiteSpace(result.Artist) || string.IsNullOrWhiteSpace(result.Title);
         }
--- a/listenarr.application/Search/Filters/ProductLikeTitleFilter.cs
+++ b/listenarr.application/Search/Filters/ProductLikeTitleFilter.cs
@@ -25,7 +25,7 @@ public class ProductLikeTitleFilter : ISearchResultFilter
 {
     public string FilterReason => "product_like_filtered";
 
-    public bool ShouldFilter(SearchResult result)
+    public bool ShouldFilter(SearchResult result, Audiobook? audiobook = null)
     {
         // If this result was enriched by a metadata source (Amazon/Audible/Audible/Audnexus/OpenLibrary),
         // prefer the enriched metadata and do not treat it as a product-like false positive.
--- a/listenarr.application/Search/Filters/PromotionalTitleFilter.cs
+++ b/listenarr.application/Search/Filters/PromotionalTitleFilter.cs
@@ -25,7 +25,7 @@ public class PromotionalTitleFilter : ISearchResultFilter
 {
     public string FilterReason => "promotional_title_filtered";
 
-    public bool ShouldFilter(SearchResult result)
+    public bool ShouldFilter(SearchResult result, Audiobook? audiobook = null)
     {
         return SearchValidation.IsPromotionalTitle(result.Title) || SearchValidation.IsTitleNoise(result.Title);
     }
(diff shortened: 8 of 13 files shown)


listenarr.api/Services/AutomaticSearchService.cs
```diff
@@ -195,8 +195,10 @@ private async Task<int> ProcessAudiobookAsync(
             var searchQuery = BuildSearchQuery(audiobook);
             _logger.LogInformation("Searching for audiobook '{Title}' with query: {Query}", audiobook.Title, searchQuery);
 
-            // Search for results
-            var searchResults = await searchService.SearchAsync(searchQuery, isAutomaticSearch: true);
+            // Search for results. Restrict to the Newznab "Books > Audiobook" category
+            // (3030) so music-only indexers configured in Prowlarr without a category
+            // restriction don't return concert/album torrents for audiobook queries.
+            var searchResults = await searchService.SearchAsync(searchQuery, category: "3030", isAutomaticSearch: true);
```
T4g1: I don't believe the category is intended to work that way. Idealy, we'd want the search to be performed on a specific indexer (later in the process), and that method should take the specific indexer it runs on as an input, then use the configured categories defined there.

At this point in the process, "3030" is only valid for specific indexers, maybe we could fix the process (which does not seem to be using the configured categories right now) and modify the targeted indexer(s) by making sure 3030 is set as a default when user configure those ?

listenarr.api/Services/QualityProfileService.cs
```diff
@@ -26,8 +26,8 @@ public interface IQualityProfileService
         Task<QualityProfile> CreateAsync(QualityProfile profile);
         Task<QualityProfile> UpdateAsync(QualityProfile profile);
         Task<bool> DeleteAsync(int id);
-        Task<QualityScore> ScoreSearchResult(SearchResult searchResult, QualityProfile profile);
-        Task<List<QualityScore>> ScoreSearchResults(List<SearchResult> searchResults, QualityProfile profile);
+        Task<QualityScore> ScoreSearchResult(SearchResult searchResult, QualityProfile profile, Audiobook? audiobook = null);
```
T4g1: For both those method, it seems redundant to send both an audiobook and the quality profile. Is there a use case to allow to perform a search for no given audiobook ? I would rather do `(SearchResult searchResult, Audiobook audiobook`

T4g1 (changes requested):
Solid work!

listenarr.api/Services/QualityProfileService.cs
kevinheneveld: There's one caller that genuinely has no audiobook: the `ScoreResults` profile-testing endpoint in `QualityProfileController` (`POST /quality-profiles/{id}/score`) scores a result list against a profile with no target book, so the relevance check is just skipped. The other three callers (`AutomaticSearchService`, `DownloadService`, `LibraryController`) always pass one.

Happy to split this into two methods — `ScoreSearchResults` requiring an audiobook for the real search path, plus a separate `ScoreAgainstProfile` for the profile-test endpoint — if you'd prefer that over the optional parameter.

listenarr.api/Services/AutomaticSearchService.cs
kevinheneveld: Agreed — the `3030` hardcode is a stopgap for the music-album bug, not the right long-term shape. The proper fix needs category config threaded per-indexer, which this PR doesn't touch.

Two options:
1. Scope the per-indexer category work into this PR, or
2. Drop the category layer from here entirely and leave this PR as the relevance + language filters — those catch the same music albums anyway (the relevance filter alone eliminated the music grabs in my testing) — then do indexer categories as a separate PR after the foundational changes land.

I'd lean toward option 2 to keep this PR focused, but happy to go either way — your call.

T4g1: I'm a very bad example at keeping PR focused ah ah but from my point of view, both option work as long as we stay agnostic of the indexers specificities. If you go for option 2, I would create an ticket to keep track of that category issue on indexers. There's much to be done on the indexer side so it could be a starting point for that

listenarr.api/Services/QualityProfileService.cs
T4g1: I see, the use case to filter search result on given profile seems burried in the front end, I guess the search results themselves are related to a given Audiobook anyway or at least the profile ID was taken from an audiobook eventualy. We might need to clean that but as far as this PR is concerned, i wouldn't change anything then as having that redundancy is mandatory for now

Edit: So the use case seem to be the following: A user can have default quality profile configured. A user can search (on configured indexers i guess) for anything. The result are filtered based on the default quality profile only. Then, i suppose if the user wants to grab one of the result, the related audiobook get created in the process if it did not existed yet. In conclusion, im convinced your approach is the correct one

listenarr.api/Services/AutomaticSearchService.cs
kevinheneveld: Going with option 2 — dropped the category-3030 piece in af5168e (force-pushed). PR is now relevance + language only. Filed #594 to track the proper per-indexer categories work for when the indexer-side foundations are ready.

kevinheneveld:
@T4g1 — I think both your points are addressed:

- **Category `3030`:** went with option 2 — dropped the category layer from this PR entirely (force-pushed), so it's just the relevance + language filters now (which caught the music-album grabs in testing anyway). Filed #594 to track the proper per-indexer category work for when the indexer-side foundations are ready.
- **`ScoreSearchResults(audiobook, profile)` redundancy:** per your conclusion we're keeping it as-is for now (the profile-test endpoint needs the no-audiobook path), so no change there.

Mind giving it another look when you have a moment? Happy to rebase onto current canary if it's drifted.

listenarr.application/Interfaces/ISearchResultFilter.cs
```diff
@@ -32,6 +32,15 @@ public interface ISearchResultFilter
...
     bool ShouldFilter(SearchResult result);
 
+    /// <summary>
+    /// Context-aware overload used when the filter pipeline knows the audiobook
+    /// the search is being run for (e.g., automatic search). Default implementation
+    /// delegates to the context-free overload so existing filters need no edits.
+    /// Filters that need the audiobook context (e.g., title-relevance) override this
+    /// and may make their context-free overload a no-op (fail open).
+    /// </summary>
+    bool ShouldFilter(SearchResult result, Audiobook? audiobook) => ShouldFilter(result);
```
T4g1: Why not have the ShouldFilter take a null default value for the audiobook instead ? Besides, the "existing filters" part only brings justification for this PR but dos not help understand the code once it gets merged

listenarr.api/Controllers/LibraryController.cs
```diff
@@ -3059,6 +3059,30 @@ private async Task<int> ProcessAudiobookForSearchAsync(
...
+                {
+                    _logger.LogInformation("Filtered {Removed} of {Total} raw results for audiobook '{Title}' via context-aware pipeline", preFilterCount - searchResults.Count, preFilterCount, LogRedaction.SanitizeText(audiobook.Title));
+                }
+                if (!searchResults.Any())
+                {
+                    _logger.LogInformation("All search results filtered for audiobook '{Title}'", LogRedaction.SanitizeText(audiobook.Title));
+                    return 0;
+                }
+            }
+            catch (Exception ex) when (ex is not OperationCanceledException && ex is not OutOfMemoryException && ex is not StackOverflowException)
```
T4g1: Why not: `catch (Exception exception) when (exception is not (OperationCanceledException or OutOfMemoryException or StackOverflowException))`

T4g1 (changes requested):

kevinheneveld:
Thanks @T4g1 — both addressed in `4f53cab4` (additive commit, no force-push):

- **Single method with a null default** — collapsed the two `ShouldFilter` overloads into one `ShouldFilter(SearchResult result, Audiobook? audiobook = null)`. The context-free overload + default-interface method are gone; filters that don't need the audiobook just ignore the param, `RelevanceFilter` keys off it (null ⇒ fail open). The doc comment now describes the null/fail-open contract instead of explaining the migration.
- **Exception filter** — switched the context-aware filter pass to `catch (Exception ex) when (ex is not (OperationCanceledException or OutOfMemoryException or StackOverflowException))`. (Left the ~42 pre-existing identical catches elsewhere in the controller alone to keep this PR scoped — happy to do a sweep separately if you'd like.)

T4g1 (approved):

T4g1:
lgtm! Let's rebase and check pipeline result before merge though

kevinheneveld:
@T4g1 rebased onto canary (now v1.0.11) as requested — the CHANGELOG conflicts are resolved and both the relevance-filter commit and the dropped-3030-category follow-up replayed cleanly. Pipeline is green locally (Release build + 134 search/filter tests pass). Ready for the re-run whenever you are.

therobbiedavis:
> @T4g1 rebased onto canary (now v1.0.11) as requested — the CHANGELOG conflicts are resolved and both the relevance-filter commit and the dropped-3030-category follow-up replayed cleanly. Pipeline is green locally (Release build + 134 search/filter tests pass). Ready for the re-run whenever you are.

@kevinheneveld The changelog is no longer needed at all. Please update the PR description to match the template.

kevinheneveld:
Rebased onto canary v1.2.1. This needed more than a replay — canary refactored the whole search subsystem since approval, so I **re-homed** the change: `ISearchResultFilter` now lives in `Search/Contracts/`, DI moved to `ListenarrWorkflowRegistration`, and the edited services relocated (`Downloads/Submission/DownloadService`, `infrastructure/HostedServices/Search/AutomaticSearchService`). The old `LibraryController` auto-pick edit is **gone** — that flow was extracted into `DownloadService.SearchAndDownloadAsync` during the refactor, which this PR already filters. Net behavior is identical to the approved version (RelevanceFilter applied before scoring in both auto-pick flows); the diff shape just changed, so a re-look is welcome. Full suite 1024/1024, format check clean. CI will need an *Approve and run*.

kevinheneveld:
Updated: folded in the **author-corroboration** refinement (`RequiresAuthorCorroboration` / `MaxGenericTitleTokens`) that was developed on the live instance after the original approval — it rejects generic single-word-title false positives that clear the raw 30% ratio on the title word alone but share no author token (e.g. a release whose title contains "Betrayed" matching the wanted "Betrayed"). 17 filter tests, full suite 1032/1032, format check clean.

PR #596, opened 2026-05-16
[fix] Mark item imported on client adapter was no longer called

T4g1:
## Summary

Hotfix related to the process of downloads: Mark item as improted was no longer being called as it should

## Testing

Added test coverage to shield from future regression on the topic

## Notes

While I was at it, I took the opportunity to rename GetClientDownloadItemId into GetExternalId in an effort to enhance the semantics used around the downloads and related interfaces (GetClientDownloadItemId and GetDownloadClientId are very confusing and did not convey the fact that this was not an ID managed by Listenarr)


therobbiedavis (approved):
lgtm

listenarr.application/Downloads/DownloadClientGateway.cs
```diff
@@ -109,13 +109,32 @@ public async Task<List<QueueItem>> GetQueueAsync(DownloadClientConfiguration cli
 
         public async Task<bool> MarkItemAsImportedAsync(DownloadClientConfiguration client, Download download, CancellationToken ct = default)
         {
-            var adapter = ResolveAdapter(client);
+            if (client == null)
```
T4g1, line 112: Those are tautology as the method signature explicitely force those to not be null (client and download)

listenarr.application/Downloads/DownloadClientGateway.cs
```diff
@@ -109,13 +109,32 @@ public async Task<List<QueueItem>> GetQueueAsync(DownloadClientConfiguration cli
...
             if (string.IsNullOrEmpty(externalId))
             {
                 return true;
             }
 
+            if (!client.IsEnabled)
+            {
+                logger.LogDebug(
+                    "Skipping mark imported for download {DownloadId}: download client {ClientId} is disabled",
+                    LogRedaction.SanitizeText(download.Id),
```
T4g1: Sanitization should be used on user provided values (API/infrastructure) and probably not in the application layer on technical IDs

listenarr.application/Downloads/DownloadClientGateway.cs
T4g1: Spoke too fast, those can be null indeed

Issue #414, opened 2026-03-14
Innacurate internet archive search results


florastiel:
Searching AUTHOR:H. P. Lovecraft returns zero ASIN candidates, then unrelated OpenLibrary fallback results like ’Salem’s Lot, Good wives, and The King in Yellow, even though the Internet Archive indexer is configured and tested successfully and relevant Archive.org items exist.

Example known-valid item:
Collected Public Domain Works of H. P. Lovecraft
Archive.org page:
https://archive.org/details/collected_lovecraft_0810_librivox

Indexer test log:

2026-03-14 09:19:13.022 -04:00 [INF] [Listenarr.Api.Controllers.IndexersController] [IndexerTest] Testing indexer Internet Archive (impl=InternetArchive, url=https://archive.org/)
2026-03-14 09:19:13.029 -04:00 [INF] [Listenarr.Api.Controllers.IndexersController] Testing Internet Archive indexer 'Internet Archive' with collection 'audio_bookspoetry'
2026-03-14 09:19:13.921 -04:00 [INF] [Listenarr.Api.Controllers.IndexersController] Internet Archive indexer 'Internet Archive' test succeeded for collection 'audio_bookspoetry'

Search example 1:
Searching for title:
Collected Public Domain Works of H. P. Lovecraft

Log excerpt:

2026-03-14 09:20:08.989 -04:00 [INF] [Listenarr.Api.Controllers.SearchController] [DBG] Advanced search request: Author='null', Title='Collected Public Domain Works of H. P. Lovecraft', Isbn='null', Asin='null', Query='null', Region='us', Language='null'
2026-03-14 09:20:08.990 -04:00 [INF] [Listenarr.Api.Controllers.SearchController] Advanced search request composed parts=TITLE:Collected Public Domain Works of H. P. Lovecraft -> query='TITLE:Collected Public Domain Works of H. P. Lovecraft'
2026-03-14 09:20:09.013 -04:00 [INF] [Listenarr.Api.Services.SearchService] Starting intelligent search for: TITLE:Collected Public Domain Works of H. P. Lovecraft
2026-03-14 09:20:09.018 -04:00 [INF] [Listenarr.Api.Services.AudimetaService] Searching audimeta.de (advanced title-only db/book) : https://audimeta.de/db/book?title=Collected%20Public%20Domain%20Works%20of%20H.%20P.%20Lovecraft&limit=50&page=1&cache=true&region=us
[...]

PR #599, opened 2026-05-17
[Architecture] Clarify download client and DDL handoffs

### LINKED ISSUE #486: [BUG] Remote path translation adds trailing slash to file paths, causing single-file imports to fail with "Source not found" retry exhaustion
## Symptom

Any completed download resolved to a single file (`.m4b`, `.mp3`, etc.) fails to import when a remote path mapping is configured for the download client. The log shows the resolved source path ends with a file extension followed by `/`, and the finalization service exhausts its retry budget without ever finding the file.

```
INFO  [DownloadMonitorService] Resolved import path for download ef04189d:
      M:/!QBDownloads/Mary Wollstonecraft Shelley - Frankenstein, or The Modern Prometheus [Simon Vance].m4b/
INFO  [DownloadMonitorService] Source not found for download ef04189d.
      Scheduling retry #1 in 30s (resolved path: M:/!QBDownloads/...Frankenstein....m4b/)
ERROR [DownloadMonitorService] Unable to locate source file for download ef04189d after 3 attempts.
      Resolved path: M:/!QBDownloads/...Frankenstein....m4b/, FinalPath=, DownloadPath=X:\
```

The file `M:/!QBDownloads/...Frankenstein....m4b` exists on disk. The path with the added `/` does not.

---

## Root Cause

`RemotePathMappingService.TranslatePathAsync` uses a private `NormalizePath` helper for both **comparison** and **result construction**. `NormalizePath` unconditionally appends `/` to every path so that `StartsWith` prefix matching works consistently across user-supplied values. This is correct for the stored mapping prefixes, but the resulting `localPath` is built by concatenating:

```
NormalizePath(mapping.LocalPath) + relativePath
```

`relativePath` is sliced from `normalizedRemotePath`, which already has the forced trailing slash. A file path like `Frankenstein.m4b` becomes `Frankenstein.m4b/` in the relative segment, and that slash is carried forward into every returned path.

**Step-by-step trace:**

| Step | Value |
|---|---|
| Input to `TranslatePathAsync` | `X:\!QBDownloads\Frankenstein.m4b` |
| After `NormalizePath` (for comparison) | `X:/!QBDownloads/Frankenstein.m4b/` |
| `normalizedMappingPath` (stored mapping `X:\!QBDownloads\`) | `X:/!QBDownloads/` |
| `StartsWith` match | ✅ True |
| `relativePath` (substring after prefix) | `Frankenstein.m4b/` ← carries the `/` |
| `NormalizePath(mapping.LocalPath)` | `M:/!QBDownloads/` |
| Returned `localPath` | `M:/!QBDownloads/Frankenstein.m4b/` ← **wrong** |
| `File.Exists(...)` | ❌ false |
| `Directory.Exists(...)` | ❌ false |
| Result | 3× retry, exhaustion, no import |

The domain model's `RemotePathMapping.NormalizePaths()` is **not** part of this bug — adding trailing slashes to the stored directory prefixes is intentional and correct.

The `ScanBackgroundService` warning (`Audiobook BasePath missing for job`) visible in the same log window is a concurrent, unrelated event: the scanner ran while the library path was inaccessible.

---

## Proposed Fix

**Single change required: `listenarr.api/Services/RemotePathMappingService.cs`, `TranslatePathAsync`.**

After constructing `localPath`, strip the trailing slash that `NormalizePath` forced onto the result — but only when the original caller input did not explicitly indicate a directory (i.e., did not already end with a slash or backslash). This preserves correct handling of directory-style inputs while fixing file-path inputs.

```csharp
// existing lines
var relativePath = normalizedRemotePath.Substring(normalizedMappingPath.Length);
var localPath = NormalizePath(mapping.LocalPath) + relativePath;

// ADD: strip the forced trailing slash for file paths
var originalTrimmed = remotePath.TrimEnd();
if (!originalTrimmed.EndsWith('/') && !originalTrimmed.EndsWith('\\'))
    localPath = localPath.TrimEnd('/');

return localPath;
```

No changes needed to `NormalizePath`, `RequiresTranslationAsync`, `RemotePathMapping.NormalizePaths`, or any adapter.

---

## Regression Tests

New test class: `tests/Listenarr.Api.Tests/RemotePathMappingServiceTranslateTests.cs`

| Test case | Input | Mapping | Expected output |
|---|---|---|---|
| Single-file path | `X:\!QBDownloads\book.m4b` | `X:\!QBDownloads\` → `M:\!QBDownloads\` | `M:/!QBDownloads/book.m4b` — **no trailing slash** |
| Directory path with trailing slash | `X:\!QBDownloads\` | `X:\!QBDownloads\` → `M:\!QBDownloads\` | `M:/!QBDownloads/` — slash preserved |
| Multi-file folder (no trailing slash) | `X:\!QBDownloads\MyBook` | `X:\!QBDownloads\` → `M:\!QBDownloads\` | `M:/!QBDownloads/MyBook` — no trailing slash |
| No matching mapping | `/remote/books/book.m4b` | `X:\!QBDownloads\` → `M:\!QBDownloads\` | `/remote/books/book.m4b` — original returned unchanged |
| Exact directory match | `X:\!QBDownloads\` | `X:\!QBDownloads\` → `M:\!QBDownloads\` | `M:/!QBDownloads/` |
| Mixed slash style | `X:/!QBDownloads/book.m4b` | `X:\!QBDownloads\` → `M:\!QBDownloads\` | `M:/!QBDownloads/book.m4b` |

---

## Acceptance Criteria

- [ ] Translated file paths do not end with `/`
- [ ] Translated directory paths (input ends with `/` or `\`) preserve trailing slash
- [ ] `File.Exists` / `Directory.Exists` check in `DownloadMonitorService` succeeds for valid translated single-file paths
- [ ] No false missing-source retry exhaustion logged for existing single-file imports after fix
- [ ] All six regression tests pass
- [ ] `dotnet build listenarr.slnx` clean — no compile regressions

### LINKED ISSUE #520: Transmission: Activity shows unrelated transfers
### Description

When using Transmission, the Activity tab shows all ongoing transfers from that dowload client.

Image

### Expected behavior

* Activity should only report ongoing download from Listenarr

### LINKED ISSUE #592: [Architecture] Responsability of download client adapters
In the context of download client adapters, the responsibility of these are not clearly identified. As a result, implementing and maintaining those is risky as we do not know which information are required or not and what it should produce precisely.

We should investigate the current features and limitations of the existing adapters. Two existing concerns:
* Adapters interface define two method to get informations about ongoing downloads: Only one of which seems to be used. We should commit to only one of those
* When giving informations about download, adapters produces QueueItem that can support RemotePath, LocalPath, ContentPath and SourceFiles. All of those are nullable fields with no clear orientation of what should be produced. In recent discussions, we suggested that an adapter should have the responsibility of producing source files but we suspect that some of them will be unable to always produce that and can instead only give a remote path.

For this last issue, we want to explore the following solutions:
* Either the remote path is given as ContentPath and SourceFiles is null, the download gateway is then responsible of mapping the path to local path and checking wether or not ContentPath is a file/directory and populate SourceFiles accordingly
* Either we allow Adapter to map path and produce SourceFiles accordingly

### LINKED ISSUE #619: Bug: NZBGet downloads are never imported — `FetchDownloadsAsync` reads numeric `listgroups` fields (`FileSizeMB`/`RemainingSizeMB`) with `.GetString()`
## Summary

## Environment

## Steps to reproduce

## Expected vs. actual

## Root cause

```csharp
```

```csharp
```

## Stack trace
```
```

## Observed log cascade
```
```

## Suggested fix
```csharp
{
    return p.ValueKind switch
    {
        _ => 0d
    };
}

```

## Impact / severity

### LINKED ISSUE #640: Two cleanup paths never wired up: unbounded DownloadProcessingJobs; orphaned downloads never terminalized
Two cleanup paths exist in the codebase but are never invoked, so state accumulates unbounded on long-running instances.

**3a — `DownloadProcessingJobs` grows unbounded.** `IDownloadProcessingJobService.CleanupOldJobsAsync(retentionDays)` is implemented (`EfDownloadProcessingJobRepository.CleanupOldJobsAsync`) but has **no caller**. The table accumulates completed/failed rows indefinitely, and every queue-snapshot reconciliation that queries it pays for the bloat.
→ **PR incoming:** a small hosted `DownloadProcessingJobCleanupService` that calls `CleanupOldJobsAsync` on a schedule (shortly after startup, then daily), with tests. Already implemented and running locally.

**3b — orphaned downloads are never terminalized.** The orphan branch in `DownloadQueueService.GetQueueSnapshotAsync` only logs ("keeping records for resilient monitoring/import handling") and never moves long-dead rows to a terminal state, so they linger in the queue-matching pool indefinitely.

Design care if we act on 3b:
- preserve the existing empty-client guard (a temporarily-unreachable client returning 0 items must **not** mass-terminalize);
- only terminalize an **allowlist** of "should be in the client but isn't" states (Queued/Paused) — never Ready/ImportPending/ImportBlocked, which are legitimately absent and would destroy pending imports;
- key off **continuous absence** (missing-since), not `StartedAt`, so a single partial snapshot can't false-terminalize.

**Open question for maintainers:** should orphans ever auto-terminalize, or stay user-driven? Given the deliberate log-only choice, an opt-in setting (default off) seems safest.

**Related:** #631 (downloads getting Import Blocked) for 3b context. Surfaced during a scale-testing discussion in Discord.

### LINKED ISSUE #652: [BUG] NZBGet imports never run — `GetImportItemAsync` keeps the stale intermediate `ContentPath` instead of resolving the final path from history
## Summary

With NZBGet (separate `InterDir` configured), every completed download fails to import. Files land in `completed/`, but Listenarr keeps scanning the `intermediate/<name>.#NZBID` working dir — which is deleted after completion — so it never imports and the Activity entry stays stuck. No remote path mapping (so not #486).

## Environment

- Listenarr `canary` (reproduced on HEAD / v1.0.6)
- NZBGet (LinuxServer): `DestDir=.../completed`, `InterDir=.../intermediate`
- Listenarr + NZBGet share `/data:/data`; no remote path mapping
- Deterministic, reproduced on 4 titles

## Symptom

```
[WRN] DownloadClientGateway: reported no source files and content path scanning failed
for item Brian.Jacques-Redwall.21-The.Sable.Quean... with path
/data/downloads/nzbget/intermediate/...#125/...
```

File is actually at `/data/downloads/nzbget/completed/.../The Sable Quean.m4b`.

## Root cause

`GetImportItemAsync` (`listenarr.infrastructure/Adapters/NzbgetAdapter.cs`, ~line 1032) returns early on a non-empty `ContentPath` without checking it exists:

```csharp
// If ContentPath is already set and exists, use it
if (!string.IsNullOrEmpty(result.ContentPath))
{
    return result;
}
```

The comment says "and exists," but there's no existence check. `MapGroup` (~line 755) has already set `ContentPath` from the active group's `DestDir`, which for an in-progress NZBGet group is the **intermediate** working dir. So this early return hands back the stale path and the history `FinalDir`/`DestDir` resolution just below (~lines 1063–1065) never runs. After completion the intermediate dir is gone, the gateway's `Directory.EnumerateFiles` throws, and import fails.

## Fix

```csharp
if (!string.IsNullOrEmpty(result.ContentPath) &&
    (File.Exists(result.ContentPath) || Directory.Exists(result.ContentPath)))
{
    return result;
}
```

Now the stale path fails the guard, execution falls through to the history lookup, `ContentPath` resolves to `completed/<name>`, and import proceeds.

Note: clearing `InterDir` is not a workaround — NZBGet still uses a `.#NZBID` temp dir that's renamed on completion, so the early return still returns a stale path.

T4g1:
## Summary

Clarifies the responsibility boundary for download client adapters, the application gateway, monitor-owned reconciliation, deferred cleanup, and Listenarr-owned direct downloads.

This PR keeps full queue snapshots available for queue display and reconciliation, uses targeted ID polling for monitor updates, adds guarded orphan cleanup for active client-backed downloads, fixes remote-path import edge cases, and makes trusted direct downloads a first-class internal transfer pipeline that is shown in Activity and handed to the normal import flow.

Fixes #486.
Fixes #520.
Fixes #592.
Fixes #619.
Fixes #640.
Fixes #652.

Supersedes #634.
Supersedes #641.

Refs #414.
Refs #618.
Refs #631.

## Direction

- Queue display still uses full live snapshots.
- Monitor updates use targeted Listenarr-owned IDs.
- Client-level monitor failures are treated as failures, not successful empty queues.
- Orphan cleanup is outside the gateway.
- Concrete client behavior lives in infrastructure workflows.
- DDL is an internal Listenarr transfer pipeline, not a fake download-client adapter.
- Future DDL sources should be additive through source policies.

### Download clients
- Split queue display and monitor polling contracts.
- Preserved full snapshots for queue display and reconciliation.
- Kept monitor updates targeted to Listenarr-owned IDs.
- Moved concrete adapter behavior into focused infrastructure workflows.
- Added guarded orphan cleanup for trusted missing external IDs.
- Removed the broken NZBGet drone identity fallback.

### Direct downloads
- Added DDL transfer state through `DirectDownloadService` and `DirectDownloadProcessor`.
- Added local DDL staging with per-artifact `.partial` writes and atomic final-file replacement before import enqueue.
- Added persisted multi-artifact DDL plans for single files, archives, or bounded track batches.
- Added source policies so each DDL source validates initial URLs, redirects, and complete artifact plans.
- Added Internet Archive representation planning for M4B, archive, and individual-track artifacts.
- Added Activity visibility for Listenarr-owned DDL rows.
- Added a DDL import-source resolver so staged local files flow into the normal import processor.
- Added an architecture guard so `DirectDownloadProcessor` does not contain provider-specific trust literals.
- Removed the DDL staging path from the info-level transfer log; outbound security logs still intentionally record blocked host/IP diagnostics for internal troubleshooting.

### Search and result display
- Improved Internet Archive query planning for sanitized punctuation and title/author searches.
- Added Internet Archive metadata representation planning for direct-download artifact selection.
- Hardened manual-search result links so generated fallback ids do not become clickable URLs.
- Hardened age display and language badge normalization, including ISO-639 language-code mapping.

### Import and cleanup
- DDL imports skip external-client mark-import calls because no external client owns them.
- Import-owned states are protected from stale live-client telemetry.
- External client imports recover stale or missing download paths through client-specific queue/history resolvers.
- Remote path mapping handles exact roots, separator-bound child paths, and source-file lists without turning single-file imports into trailing-directory paths.
- NZBGet and SABnzbd path resolution remains client-specific and validated before import.
- Deferred cleanup remains separate from import execution.

## Testing

Recent local validation:

- Backend DDL/source-policy/processing slice: passed 44.
- Backend queue/IA search/trusted-candidate slice: passed 40.
- Focused DDL processor slice after logging cleanup: passed 9.
- ActivityView and ManualSearchModal frontend tests: passed 18.
- `npx vue-tsc --build tsconfig.app.json`: passed.
- `git diff --check`: passed.

Commit-time validation also passed staged formatting, frontend checks where applicable, Prettier where applicable, and layering/async-void checks.

## Credit

Thanks to @kevinheneveld for identifying the unbounded processing-job retention problem in #640 and implementing the original cleanup direction in #641. #599 incorporates that cleanup purpose into the broader architecture with a separate cleanup processor so import execution and retention maintenance remain separate.

Thanks to @whacky671 for the detailed NZBGet JSON number bug report in #619/#618 and the original fix work in #628.

Thanks to @NiCADitO for the NZBGet stale intermediate `ContentPath` root-cause report in #652, including the path-existence guard that informed the final import resolver fix.

Thanks to @meister-eraser for the SABnzbd import-blocked report in #631, which helped validate the path-resolution cleanup in this area.

## Notes

The architecture is intentionally conservative. Active client-backed records are removed only after a trusted live snapshot, while full queue snapshots remain available for queue display and reconciliation. DDL downloads are excluded from external-client orphan cleanup because they are internally tracked.

DDL is intentionally modeled as an internal transfer pipeline rather than a download-client adapter. Adding future DDL sources should be additive through source policies and should not require changing `DirectDownloadProcessor`.

I audited nearby open issues and PRs while refreshing this description. This PR now closes #486 in addition to the already listed download-client, import, and cleanup issues. It references #414 because the Internet Archive search/result changes are relevant, but #414 should remain open unless the original reporter confirms the specific inaccurate-result examples are resolved. Open PRs for additional providers or unrelated clients are not superseded by this PR.


listenarr.application/Downloads/Common/DownloadClientGateway.cs
```diff
@@ -131,6 +115,48 @@ public async Task<QueueItem> GetQueueItemAsync(
             return await TranslateQueueItemPathsAsync(client, item);
         }
 
+        public async Task<List<Download>> FetchDownloadsAsync(DownloadClientConfiguration client, List<Download> downloads, CancellationToken ct = default)
+        {
+            var ids = GetExternalIds(downloads);
```
therobbiedavis, line 144: This path now only polls downloads that already have a stored external ID. If a legacy (or hopefully not a new record) is missing ClientDownloadId/TorrentHash, the adapter gets an empty ID list and the download can stay stuck without progress updates. Probably a reasonable tradeoff though, but maybe we should include a view for users to see "orphaned" records? wdyt

listenarr.application/Downloads/Common/DownloadClientGateway.cs
```diff
@@ -90,21 +82,13 @@ public Task<bool> RemoveAsync(DownloadClientConfiguration client, string id, boo
...
-            return translatedResults;
-        }
+            var downloads = await downloadRepository.GetByClientAsync(client.Id);
+            var ids = GetExternalIds(downloads);
 
-        public Task<List<(string Id, string Name)>> GetRecentHistoryAsync(DownloadClientConfiguration client, int limit = 100, CancellationToken ct = default)
-        {
             var adapter = ResolveAdapter(client);
-            return adapter.GetRecentHistoryAsync(client, limit, ct);
+            var items = await adapter.GetQueueAsync(client, ids, ct);
```
therobbiedavis: I think this changes the queue endpoint  fetching the client queue to fetching only downloads already tracked in our DB. That drops untracked queue items and completed external downloads before DownloadQueueService can reconcile or display them.

therobbiedavis (changes requested):
Some logic questions, also this needs to be rebased.

listenarr.application/Downloads/Common/DownloadClientGateway.cs
T4g1: My reasoning is the following: If all adapters have to return an external ID when we add a download: It means that every Listenarr Download has an external ID. So, any Download without an external ID should be dropped and any queue item without external ID should be ignored.

By fetching the whole queue and deciding what to use or not in the application, we gather informations that are not related to Listenarr and should probably never leave the adapter layer (by design, it solves some issues I witnessed where all queue items are shown for an adapter which is not expected)

listenarr.application/Downloads/Common/DownloadClientGateway.cs
T4g1: I believe other *arr silently remove the download in case it becomes orphaned (no external ID or external ID no longer found in the download client), this seems like a good idea as in that case, automatic search will probably pick it up again later or the user can re-start a manual search (audiobook would be shown in the wanted list with no active download on it).
Also, I expect the majority of orphaned download to come from, either migration to this version or later (one time issue and the software is in canary phase anyway) or from download client manualy removing the queue item from their end (probably due to user input or configuration)
So, if we remove them directly (or even after retrying some times to get it from the adapter), by design we no longer have orphaned download and thus the view is probably not useful :p But if we dont go in that direction, then yes, we could bring a new oprhaned status meaning the download can no longer be linked with anything on the adapter side and the user could then remove only or remove and re-search them

I can see an issue with removing orphaned download: If the adapter is no longer reachable or experience technical issues, then it will send an empty list of queue items and we may delete them by error in those case. We can mitigate that by always checking if the adapter is up and running (using the TestAdapter method) first

listenarr.application/Downloads/Common/DownloadClientGateway.cs
therobbiedavis: That makes sense but this is serving two different contracts. For monitor/update polling, targeted IDs are exactly what we want. For `DownloadQueueService`, though, this is still the live queue snapshot source, and the reconciliation/display branches below it still assume they may see unknown client items.

If the desired behavior is that unknown client items are never surfaced, which I believe is the proper way, then I think this is fine as a product boundary. The only thing I’d clean up is the now-stale contract in `DownloadQueueService`: it still has reconciliation/display paths for unknown client items and completed external downloads, but those paths can no longer be reached because the adapter filters before returning.

WDYT? Should we show unknown client items?

listenarr.application/Mapping/QueueItemConverter.cs
```diff
@@ -1,9 +1,52 @@
...
+            }
+
+            var normalizedState = (item.Status ?? string.Empty).ToLowerInvariant();
+            if (normalizedState == "error" || normalizedState == "missingfiles")
+            {
+                download.Failed($"qBittorrent state: {item.Status}");
+                return download;
+            }
+
+            if (item.Progress >= 100)
```
therobbiedavis: I believe qBittorrent can report `amount_left == 0` while progress is slightly under 100, and the old path handled that with `progress >= 1.0 || amountLeft == 0`. Without the fallback, those downloads can probably remain stuck as downloading instead of moving into completed/import handling.

listenarr.application/Downloads/Common/DownloadClientGateway.cs
therobbiedavis: Then I think we should follow suite and silently remove the download if it is orphaned. That seems like the easiest way to handle it and in case client api's change in the future we have something that kind of helps with potential download issues. I beleive we already do this, but we should make sure that orphaned cleanups are logged for debugging.

therobbiedavis (changes requested):

listenarr.application/Downloads/Common/DownloadClientGateway.cs
therobbiedavis: Sorry for taking some liberties with the PR here, but I went ahead and made the orphan behavior explicit because it ended up being part of the adapter contract once monitor polling became ID-based.

The current implementation follows the silent-cleanup direction from this thread, but with guardrails so we do not delete records just because a client had a bad poll. Orphan cleanup now lives outside `DownloadClientGateway`, because the gateway should not own persistence. The monitor path calls `DownloadOrphanCleanupService` after a trusted live snapshot.

It removes active client-backed downloads when either their stored external ID is missing from the live snapshot, or the record is a non-DDL active download with no external ID at all. It skips cached snapshots, unavailable clients, suspicious empty snapshots, recent downloads, terminal/import states, and DDL records. Cleanup is logged and emits metrics so it is still debuggable even though we are not surfacing a separate orphaned status to users.

listenarr.application/Downloads/Common/DownloadClientGateway.cs
therobbiedavis: This ended up landing as the split-contract version rather than filtering the gateway queue down to only tracked records.

`GetQueueAsync(client)` is now documented and used as the full live queue snapshot source for `DownloadQueueService`, so the reconciliation/display paths for unknown queue items and completed external items remain reachable. The monitor/update path uses the ID-filtered overload instead, through `DownloadClientGateway.FetchDownloadsAsync()`, so regular progress polling does not need to process unrelated client items.

That keeps the UI/reconciliation contract and the monitor contract separate, which avoids the contradiction called out here.

listenarr.application/Mapping/QueueItemConverter.cs
therobbiedavis: This is handled now, but with an extra reliability gate to avoid the regression where unknown size telemetry looked like a completed download.

`QueueItemConverter` now treats byte telemetry as reliable only when the client reports a positive total size and a non-negative downloaded value. Completion can still happen from reliable `amountLeft <= 0`, so qBittorrent-style cases where the remaining byte count reaches zero are covered. It can also happen from an explicit terminal client state like `completed` or `success`.

The important part is that `Size = 0` and `Downloaded = 0` no longer means completed by accident.

therobbiedavis (approved):
Code review looks clean now. The adapter/gateway contract split is clear, orphan cleanup has the right guardrails, completion mapping is protected against unreliable telemetry, and the SABnzbd path handling fixes the empty/stale path behavior without inventing import paths. Tests and checks are green.

therobbiedavis:
Implemented the #520 Activity visibility fix in local commit 082d9b04. The queue service still fetches full client snapshots for reconciliation and rebinding, but unmatched active external client items are hidden from Activity for all clients. Matched/rebound Listenarr downloads remain visible, and unmatched completed external items remain controlled by ShowCompletedExternalDownloads. Local validation: queue reconciliation tests passed 26/26; full suite passed 1117/1117.

therobbiedavis:
Updated PR with ed2dc6d0: moved download-client workflow composition into infrastructure DI. Full local suite passed 1122/1122; adapter suites passed 170/170; queue/monitor tests passed 41/41.

PR #604, opened 2026-05-18
fix(dockerfile): normalize wwwroot permissions so the non-root runtime user can read static files

kevinheneveld:
## Summary

Drafted for visibility per the pacing note on #590 — happy to leave this in draft until you have time.

The Docker final stage copies `/app/publish` from the build stage but doesn't normalize permissions on `/app/wwwroot/*`. Depending on the build host's umask and the source files' modes, the non-root runtime user (`listenarr`) can hit `EACCES` reading individual static assets — typically manifests itself as 403/404 on a subset of bundled FE files (e.g., a single chunk fails to load and the SPA boots into a blank screen).

Fix: add a single Dockerfile step right after `COPY --from=build /app/publish .` that walks `/app/wwwroot` and sets directories to `755` and files to `644`. Matches the conventional mode for served static content and is no-op on most build hosts.

```dockerfile
# Normalize frontend asset permissions so the non-root runtime user can read static files.
RUN find /app/wwwroot -type d -exec chmod 755 {} \; \
    && find /app/wwwroot -type f -exec chmod 644 {} \;
```

## Test plan

- [x] Image builds clean on linux/amd64
- [x] Verified on a downstream deploy: a previously-403'd `assets/*.js` chunk now serves correctly under the listenarr user without root override.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -78,6 +78,10 @@ RUN groupadd --system listenarr \
 
 COPY --from=build /app/publish .
 
+# Normalize frontend asset permissions so the non-root runtime user can read static files.
+RUN find /app/wwwroot -type d -exec chmod 755 {} \; \
+	&& find /app/wwwroot -type f -exec chmod 644 {} \;
+
 # Ensure config directory exists
 RUN mkdir -p /app/config/database
 
--- a/listenarr.api/Dockerfile.runtime
+++ b/listenarr.api/Dockerfile.runtime
@@ -52,6 +52,10 @@ RUN groupadd --system listenarr \
 # The workflows publish into listenarr.api/publish, so copy that directory into the image root.
 COPY listenarr.api/publish/ /app/
 
+# Normalize frontend asset permissions so the non-root runtime user can read static files.
+RUN find /app/wwwroot -type d -exec chmod 755 {} \; \
+	&& find /app/wwwroot -type f -exec chmod 644 {} \;
+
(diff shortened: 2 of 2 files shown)


T4g1 (approved):

therobbiedavis (changes requested):
Requesting changes here because this fixes the root `Dockerfile`, but not the Dockerfile used for the images we actually publish.

The release workflow builds GHCR, Docker Hub, and ATCR images with `listenarr.api/Dockerfile.runtime`:

That file still does:

```dockerfile
COPY listenarr.api/publish/ /app/
```
at `listenarr.api/Dockerfile.runtime:53` without normalizing `/app/wwwroot`, so the published images can still hit the same non-root static asset permission issue this PR is trying to fix.

Could you add the same chmod step after that copy in `listenarr.api/Dockerfile.runtime` as well? Keeping the root Dockerfile change makes sense for manual/source builds, but the runtime Dockerfile needs it too for the shipped containers.

kevinheneveld:
Good catch — added the same chmod step to `listenarr.api/Dockerfile.runtime` right after the `COPY listenarr.api/publish/ /app/` at line 53 (`bda11300`). The runtime Dockerfile now mirrors the root one for the `/app/wwwroot` permissions, so the shipped GHCR / Docker Hub / ATCR images get the fix too.

Kept the root Dockerfile change in place for the manual / source-build path, as you noted that path still matters.

therobbiedavis (approved):
lgtm

PR #605, opened 2026-05-18
fix(search): fall back to default quality profile when audiobook has none, and expose bulk-assign in Wanted view

kevinheneveld:
## Summary

Drafted for visibility per the pacing note on #590 — happy to leave this in draft until you have time.

Two related symptoms users hit when an audiobook is monitored but doesn't have a quality profile assigned:

1. **Automatic search silently returns 0:** `AutomaticSearchService.ProcessAudiobookAsync` and `DownloadService.SearchAndDownloadAsync` both early-return when `audiobook.QualityProfile == null`. An unconfigured book is monitored, shows up in Wanted, but never gets searched — the user has no feedback as to why.
2. **No way to fix it in bulk:** Setting a profile requires opening every wanted book's edit modal. With a default profile configured in settings, the user expects it to apply.

## Fix

**Backend:** Both auto-search paths now fall back to `qualityProfileService.GetDefaultAsync()` when the audiobook has no profile assigned. If a default is configured the audiobook gets searched as if that were its profile (in-memory only — no DB write). If there's no default either, log the reason and return 0 as before.

**Frontend:** A new "Assign default profile (N)" button on the Wanted view appears only when at least one wanted book has no profile and a default is configured. Clicking it bulk-assigns the default to every such book via the existing `PUT /library/{id}` endpoint. Per-book failures are surfaced individually but don't stop the batch.

Also: the Wanted row's quality column used to show `item.quality ?? 'Unknown'` for profile-less books; now shows "No profile" so the cause is visible at a glance.

## Test plan

- [x] Backend builds clean
- [x] Frontend builds clean (`npm run build`)
- [ ] Verified in a downstream deploy: a previously-stuck wanted book gets searched as soon as a default profile is configured (no longer requires per-book edit + save round-trip).

### DIFF
--- a/fe/src/views/content/WantedView.vue
+++ b/fe/src/views/content/WantedView.vue
@@ -43,6 +43,20 @@
           <PhRobot />
           Search All
         </button>
+        <button
+          v-if="wantedWithoutProfileCount > 0"
+          class="btn btn-secondary"
+          @click="assignDefaultProfileToWanted"
+          :disabled="assigningQualityProfiles || !defaultQualityProfile"
+          :title="
+            defaultQualityProfile
+              ? `Assign ${defaultQualityProfile.name} to ${wantedWithoutProfileCount} wanted audiobooks`
+              : 'No default quality profile is configured'
+          "
+        >
+          <PhStar />
+          Assign default profile ({{ wantedWithoutProfileCount }})
+        </button>
         <button class="btn btn-secondary" @click="openManualImport">
           <PhFolderPlus />
           Manual Import
@@ -122,7 +136,7 @@
             </div>
             <div class="col-quality">
               <span class="quality-tag">
-                {{ getQualityProfileForAudiobook(item)?.name ?? item.quality ?? 'Unknown' }}
+                {{ getQualityProfileForAudiobook(item)?.name ?? 'No profile' }}
               </span>
             </div>
             <div class="col-status">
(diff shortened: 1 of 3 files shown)


therobbiedavis:
Hey @kevinheneveld Would you mind joining us on https://discord.gg/CwZ2Sqp9NF

We're working on prioritizing PRs and you've opened a good number of them, so I figured you should be involved. 😜

kevinheneveld:
Rebased onto canary v1.2.1 and re-homed onto the refactored search subsystem. Notes for review:
- The two auto-pick flows relocated (`Downloads/Submission/DownloadService`, `infrastructure/HostedServices/Search/AutomaticSearchService`) and the quality-cutoff helpers now live in `AutomaticSearchQualityEvaluator`. Rather than thread a profile parameter through the relocated evaluator, the fallback resolves the default and assigns it to the **in-memory** `audiobook.QualityProfile` (the search flow never persists it), so the evaluator and scorer pick it up transparently — smaller, equivalent change.
- Bulk-assign now goes through `apiService.bulkUpdateAudiobooks` (returns per-row results; partial failures surfaced, batch not aborted).
- `DownloadService` sits at the repo's 500-line architecture limit, so the `DownloadService` fallback is kept terse (no success-log; `AutomaticSearchService` logs the fallback).
- Backend 1015/1015 + `dotnet format` clean; FE type-check, lint, `WantedView.spec` (3/3), prettier all clean. `CHANGELOG.md` not touched (gone on canary). CI will need an *Approve and run*.

PR #607, opened 2026-05-18
Fix/602: root folder deletion

### LINKED ISSUE #602: [Bug] Unable to delete root folder
Trying to delete an unused root folder that is not linked to any current audiobooks and I will receive the following error:

API error: 400 {"message":"Root folder is in use by audiobooks; reassign before deletion or provide reassignRootId."}

I've reassigned all audiobooks to a new directory but it seemingly keeps thinking that there is still a book assigned to the old root folder.

T4g1:
## Summary

* Fix #602 
When removing root folders that are contained within other root folders, audiobooks can be detected as using both of them.
I fixed that issue by making it count the amount of root folders associated with each audiobooks and allowing the deletion only if none of them is orphaned from all root folders 

## Testing

- Reworked the root folders tests and added two tests to prevent regression and validate the behavior

## Notes

- Some other areas of the root folder services assumes root folders are not contained within each other, we might want to create an issue to investigate the behavior regarding those and decide if we want to prevent that use case or not


listenarr.application/Audiobooks/RootFolderService.cs
```diff
@@ -0,0 +1,192 @@
...
+            if (rootFolder == null)
+            {
+                logger.LogWarning($"Root folder with id {id} cannot be found, assuming it's deleted");
+                return;
+            }
+
+            rootFolders = [.. rootFolders.Where(r => r.Id != id)];
+
+            var audiobooks = await audiobookRepository.GetAllAsync();
+            var rootedAudiobooks = audiobooks.Where(a => !string.IsNullOrEmpty(a.BasePath) && !rootFolders.Any(r => a.BasePath.StartsWith(r.Path)));
```
therobbiedavis: A few things here:

- This looks to breaks the reassignRootId path, because we throw as soon as rootedAudiobooks.Any() is true. I think we need to validate/migrate the target root before this orphan check, or evaluate the orphan state after reassignment.

- I think StartsWith(r.Path) will treat sibling prefixes as matches, for example /data/media/book would match an audiobook under /data/media/books/.... so it could delete the actual owning root because the audiobook appears to still be covered by the sibling root. Maybe we use normalized path-boundary matching, similar to exact match or rootPath + separator?

- One edge case with this approach: because this checks all audiobooks against the remaining roots, a pre-existing orphaned audiobook unrelated to the root being deleted will also block deletion.

For example, deleting an unused `/library/old` root would fail if the DB already contains an audiobook at `/imports/orphan/Book`, even though deleting `/library/old` did not orphan it. Is that intentional?

therobbiedavis (changes requested):

listenarr.application/Audiobooks/RootFolderService.cs
```diff
@@ -0,0 +1,204 @@
...
+            var orphanedAudiobooks = audiobooks.Where(a => !string.IsNullOrEmpty(a.BasePath) && !rootFolders.Any(r => a.BasePath.StartsWith(r.Path)));
+            if (orphanedAudiobooks.Any())
+            {
+                var formattedList = string.Join(", ", orphanedAudiobooks.Select(a => a.BasePath));
+
+                logger.LogWarning($"The following audiobooks are orphaned: {formattedList}");
+            }
+
+            var rootedAudiobooks = audiobooks
+                .Where(a => !orphanedAudiobooks.Any(o => o.Id != a.Id)) // Check only audiobooks that are not orphaned
```
therobbiedavis: I think the orphan check is inverted. `!orphanedAudiobooks.Any(o => o.Id != a.Id)` should be checking for the same audiobook, right? So shouldn't this be `o.Id == a.Id`. As written, if there are 2 or more pre-existing orphans, the safety check can become unreliable and allow deletions that would orphan books.

listenarr.application/Audiobooks/RootFolderService.cs
```diff
@@ -0,0 +1,204 @@
...
+
+            if (reassignRootId != null)
+            {
+                var newRoot = await rootFolderRepository.GetByIdAsync(reassignRootId!.Value) ?? throw new KeyNotFoundException("Reassign root not found");
+                await MigrateAudiobookPathsAsync(rootFolder.Path, newRoot.Path);
+            }
+
+            var audiobooks = await audiobookRepository.GetAllAsync();
+
+            var orphanedAudiobooks = audiobooks.Where(a => !string.IsNullOrEmpty(a.BasePath) && !rootFolders.Any(r => a.BasePath.StartsWith(r.Path)));
```
therobbiedavis: The path check here looks to be missing a trailing separator: `.Any(r => a.BasePath.StartsWith(r.Path))`. That means `/data/media/books` could incorrectly match `/data/media/bookstore/a1`. `rootFolderPathsAfterDelete` already uses `EnsureTrailingSeparator`; this check should do the same.

listenarr.application/Audiobooks/RootFolderService.cs
```diff
@@ -0,0 +1,204 @@
...
+                logger.LogWarning($"Root folder with id {id} cannot be found, assuming it's deleted");
+                return;
+            }
+
+            var rootFolderPathsAfterDelete = rootFolders.Where(r => r.Id != id).Select(r => FileUtils.EnsureTrailingSeparator(r.Path)).ToList();
+
+            if (reassignRootId != null)
+            {
+                var newRoot = await rootFolderRepository.GetByIdAsync(reassignRootId!.Value) ?? throw new KeyNotFoundException("Reassign root not found");
+                await MigrateAudiobookPathsAsync(rootFolder.Path, newRoot.Path);
```
therobbiedavis, line 78: This is closer, but I think reassignRootId still has an edge case. If an audiobook at the deleted root is migrated to exactly newRoot.Path, the later check compares it against newRoot.Path + separator, so validation still treats it as uncovered and throws after mutating the audiobook path.

therobbiedavis (changes requested):

listenarr.application/Audiobooks/RootFolderService.cs
therobbiedavis: Splitting this aggregate comment into separate line-specific threads so each issue can be tracked independently. Resolving this one to avoid duplicate review threads.

listenarr.application/Audiobooks/RootFolderService.cs
T4g1: Im not sure about what you mean here but does my "refactor" to have audiobooks and root folder path trailing separator enforced fixes that ?

listenarr.application/Audiobooks/RootFolderService.cs
```diff
@@ -0,0 +1,208 @@
...
+            }
+
+            var audiobooks = await audiobookRepository.GetAllAsync();
+
+            // Audiobooks are considered orphaned if base path is empty or no root folder can be linked to it
+            var orphanedAudiobooks = audiobooks
+                .Where(a => string.IsNullOrEmpty(a.BasePath) || !rootFolders.Any(r => a.BasePath!.StartsWith(r.Path)));
```
therobbiedavis, line 85: The trailing separator fix helps, but I’d still avoid raw `StartsWith` for path ownership. This comparison is case-sensitive, so Windows-style paths with different casing can be treated as unrelated and mark a real child audiobook as orphaned.

listenarr.domain/Models/Audiobook.cs
```diff
@@ -65,10 +65,14 @@ public string? BasePath
...
             get
             {
-                // TODO: Should be put on the set operation with appropriate DB migration to normalize existing data
-                return FileUtils.NormalizeStoredPath(field);
+                field = FileUtils.NormalizeStoredPath(field);
+                return FileUtils.EnsureTrailingSeparator(field);
+            }
+            set
+            {
+                var normalized = FileUtils.NormalizeStoredPath(value);
```
therobbiedavis, line 73: I think we need to guard null/empty before normalizing here. Right now `NormalizeStoredPath(null)` / `NormalizeStoredPath("")` can turn “no BasePath” into the current working directory, which breaks callers checking for an empty BasePath.

listenarr.infrastructure/Persistence/Repositories/AudiobookRepository.cs
```diff
@@ -120,7 +119,6 @@ public async Task<bool> UpdateAsync(Audiobook audiobook)
                 System.Diagnostics.Debug.WriteLine("Suppressed non-fatal exception in catch block.");
             }
 
-            _db.Audiobooks.Update(audiobook);
             await _db.SaveChangesAsync();
```
therobbiedavis, line 122: I think this lost the actual update. `SaveChangesAsync` only persists if the audiobook is already tracked, so a detached entity can return `true` here while saving nothing.

listenarr.api/Program.cs
therobbiedavis, line 558: I don’t see `DisableHostedServices` being read, so `WebApplicationFactory` tests will start the background workers when the test host boots.

therobbiedavis (changes requested):
Thanks for the updates here, the original comments are resolved but sorry for finding more stuff.

listenarr.infrastructure/Persistence/Repositories/AudiobookRepository.cs
T4g1: The update on tracked entities throws errors (you can check the pipelines on the MR) so i either remove the tracking or go full tracking, i went for the default tracking behavior here

So it is assumed the entities are tracked here (which is the case in the repository used currently iirc)

therobbiedavis: Ah ok, I think my primary concern is the method contract more than this specific call path. It now only works for already-tracked instances, but still looks like a general update method and returns `true` for detached entities too. Maybe instead we only call `Update` when the entity is detached? WDYT?

T4g1: I believe its a matter of convention, but we have an issue: we have inconsistent behavior right now (some method do track others dont, and by default, the framework does implicit tracking)
Like I said, I went for the implicit tracking pattern but we could also:
* Disable tracking: Which requires us to explicitely update objects
* Explicit the tracking mode in the method name

I considered the approach to check if it's tracked or not too but it feels a bit like code smell for me as if we have to do that, it's maybe a lack of comprehension about the inner working of the EF framework from the dev that needs it, i can't really think of a use case where we would want to update an untracked object out of thin air)

therobbiedavis:
#717 substantially rewrites root-folder path semantics, deletion/reassignment coordination, active move/relocation guards, and atomic persistence. This PR still covers a distinct legacy case: deleting a parent root when existing nested-root configurations cause books under the child root to be counted against the parent. Please rebase on #717 and adapt that legacy-nested-root logic to the new filesystem semantics and ReassignAudiobooksAndRemoveAsync contract rather than restoring the older service/repository flow.

Issue #438, opened 2026-03-26
[BUG] wild copy of root audiobook folder


julianhille:
## issue

It appeared that the mover somehow multi nested all files in the audio book folder into a specific audio book multiple times.

Structure:
- audio books
  - book 1
  - book 2

New download did create this not sure if manual or auto map did it:

- audio books
  - book 1
  - book 2
    - book 1
    - book 2
      - book 1
      - book 2
      - ...... Like thousands of timesm

Luckily I use zfs which deduplicated but this nested folder was in the end like 2.6tb in size even though I only have some hundreds mbs of audio books at all. 

Had to manually delete them

sellerthorpe:
I had this happen as well.  It occurs when the "book 1" in the series is the same title as the series itself.

Original structure:
Series 1 (name matches book 1, so subfolder isn't created
  Book 2
  Book 3, etc

When you try to move Book 1 into a unique folder, it grabs all the content (Book2, Book 3, etc) and tries to move them. Causes a nasty copy loop...

T4g1:
The following issue is probably related to this too: #577 
Right now, only audio files are tracked as part of Audiobook, we should probably register companion files too (as a simple list of string for example). That way, moving/renaming no longer relies on directory scanning but on concrete informations (if my assumptions are correct)

I believe we have a scanning job that is also actively scanning directories to complete existing audiobooks, that will probably become irrelevant if that's the case

T4g1:
I just checked, MoveBackgroundService is indeed using directory scanning to perform move jobs, which is probably a bad idea in a context where Listenarr cannot enforce one directory = one audiobook only (and probably shouldn't as the user might want to put unrelated data in folders anyway)

T4g1:
Also, i just found an infinite loop for some case in FileMover.cs: there is a recursive method with no guards against infinite recursion on CopyDirRecursive so moving a directory in a subdirectory of that directory is not recommended right now

m4bard:
[...]

PR #608, opened 2026-05-21
[fix] Infinite loop on some file operations

### LINKED ISSUE #598: [Testing] Validate hardlink/copy import behavior
## Context

During discussion around PR #596, Robbie noted that the import path worked, but the hardlink behavior itself was not verified because their environment does not use hardlinks:

> I can’t test if hardlink works as I don’t use it. I just tried coding it up to the best of my knowledge and hoped for the best.

T4g1 suggested tracking this so it can be tested later on a Raspberry Pi/Linux setup.

## What to Validate

- `hardlink/copy` completed-file action creates hardlinks when source and destination support them.
- It falls back to copy when hardlinking is not possible.
- Imported audiobook files are registered against the final library path.
- The source download files are not unexpectedly removed when using hardlink/copy.
- Logs make it clear whether a hardlink or fallback copy occurred.

## Suggested Test Setup

- Linux/Raspberry Pi environment.
- Source and library paths on the same filesystem for the hardlink-success case.
- Source and library paths on different filesystems, or another unsupported hardlink scenario, for fallback-copy behavior.

## Acceptance Criteria

- Hardlink success path is verified manually or with automated coverage.
- Fallback-copy path is verified manually or with automated coverage.
- Any incorrect behavior found during validation is fixed or split into follow-up issues.

T4g1:
## Summary
* Closes #598 
* Help with #438 

### Added
- Tests for hardlink operations
- Tests for recursive directory copy

## Testing

Added automated testsings

## Notes

This should not solve #438 but might help in some cases


listenarr.infrastructure/FileSystem/FileMover.cs
```diff
@@ -109,48 +100,15 @@ public async Task<bool> MoveDirectoryAsync(string sourceDir, string destDir)
                 try { Directory.Delete(sourceDir, true); }
```
therobbiedavis, line 111: I think we need a guard here for the case where destDir is inside sourceDir. As written, the fallback can copy into the source tree, then delete sourceDir, which also deletes the destination it just created, and still return true.

listenarr.infrastructure/FileSystem/FileMover.cs
```diff
@@ -109,48 +100,15 @@ public async Task<bool> MoveDirectoryAsync(string sourceDir, string destDir)
...
-                    }
-                }
-                catch (Exception rex) when (rex is not OperationCanceledException && rex is not OutOfMemoryException && rex is not StackOverflowException)
-                {
-                    _logger.LogWarning(rex, "Robocopy fallback threw an exception");
-                }
+                logger.LogError(ex, "Copy+delete fallback failed for directory {Source} -> {Dest}", sourceDir, destDir);
 
-                return false;
+                return await MoveWithRobocopy(sourceDir, destDir);
```
therobbiedavis: This looks like it accidentally routes directory moves through the file robocopy path. MoveWithRobocopy splits the input into parent directory + file name and uses /MOV, so a directory fallback ends up with the wrong arguments. I think directory moves need their own robocopy path using sourceDir, destDir, /MOVE, and the normal robocopy success range.

listenarr.infrastructure/FileSystem/FileMover.cs
```diff
@@ -346,64 +253,54 @@ public Task<bool> HardlinkFileAsync(string sourceFile, string destFile)
...
             }
 
-            foreach (var file in Directory.GetFiles(src, "*.*", SearchOption.TopDirectoryOnly))
+            Directory.CreateDirectory(dst);
+            foreach (var sourceSubdirectory in Directory.GetDirectories(src, "*", SearchOption.TopDirectoryOnly))
             {
-                var destFile = Path.Join(dst, Path.GetFileName(file));
-                File.Copy(file, destFile, true);
+                var destinationSubdirectory = Path.Join(dst, Path.GetFileName(sourceSubdirectory));
+                if (destinationSubdirectory.StartsWith(src, StringComparison.OrdinalIgnoreCase))
```
therobbiedavis: This prefix check is a little too broad. A destination like source-copy also starts with source, so this can skip valid nested folders even though the destination is not actually inside the source directory.

listenarr.infrastructure/FileSystem/FileMover.cs
```diff
@@ -346,64 +253,54 @@ public Task<bool> HardlinkFileAsync(string sourceFile, string destFile)
...
-                        File.Copy(sourceFile, tempCopyPath, overwrite: true);
-                        File.Move(tempCopyPath, destFile, overwrite: true);
-                        _logger.LogInformation("Copied file (hardlink fallback): {Source} -> {Dest}", sourceFile, destFile);
-                        return Task.FromResult(true);
-                    }
-                    finally
+                        logger.LogWarning(linkEx, "Hardlink failed, falling back to copy: {Source} -> {Dest}", sourceFile, destFile);
+
+                    // Copy fallback
+                    if (await CopyFileAsync(sourceFile, destFile))
```
therobbiedavis, line 275: This seems to lose the safer temp-file behavior from the old fallback. If the hardlink fails, we now copy directly over destFile; if that copy fails halfway through, an existing destination could be left partially overwritten.

tests/Features/Infrastructure/FileSystem/FileMoverTests.cs
```diff
@@ -0,0 +1,356 @@
...
+            var destInfo = new FileInfo(destFile);
+            Assert.Equal(sourceInfo.Length, destInfo.Length);
+        }
+
+        [Fact]
+        public async Task CopyDir_NoInfiniteLoop()
+        {
+            // Arrange
+            var sourceDirectory = FileService.GetTempDirectory("source");
+            var sourceSubDirectory = FileService.GetTempDirectory(FileUtils.GetAbsolutePath("source", "test"));
```
therobbiedavis: This test is building an absolute path and then passing it into GetTempDirectory, which creates an invalid Windows path like ...\Temp\...\C:\source\test. Because of that, the test fails before it actually exercises the recursive copy logic.

therobbiedavis (changes requested):
Outside of the required rebase, I think this needs a few fixes before merging:

- Add a guard for destDir being inside sourceDir before the copy+delete fallback, so we don’t delete the destination and still return success.
- Split the robocopy fallback into separate file and directory paths; directory moves should use sourceDir, destDir, /MOVE, and the normal robocopy success range.
- Fix the recursive-copy test setup on Windows so it uses paths under the temp directory and actually reaches the behavior it’s trying to cover.
- Restore the temp-copy-then-rename behavior for the hardlink fallback, so an existing destination is not partially overwritten if the fallback copy fails.

listenarr.infrastructure/FileSystem/FileMover.cs
T4g1: I removed it because imo, it does not brings any safer fallback: Why would a move tmp to dest be safer than a copy src to dest ? It only brings more IO and therefore more risk of failures I believe

tests/Features/Infrastructure/FileSystem/FileMoverTests.cs
T4g1: Yeah, i fixed that while adding test for your other remarks, I have no idea how it was even working like this ah ah

listenarr.infrastructure/FileSystem/FileMover.cs
therobbiedavis: That's fair, I just like data preservation whenever possible. So if a file fails halfway through we could retry either from tmp to dest, or src to tmp. Im fine without the tmp though.

listenarr.infrastructure/FileSystem/FileMover.cs
therobbiedavis: Your update does address part of the concern: passing `sourceDir` and `destDir` directly avoids the earlier issue where the directory path was split like a file path but I think there’s still one directory-specific piece left, though.

It looks like `MoveWithRobocopy(sourceDir, destDir, "*.*")` still uses the file-style robocopy behavior with `/MOV` and only treats exit code `1` as success. For a directory move fallback, I’d think we should use `/MOVE` with `/E`, and accept robocopy’s normal success range (`0` through `7`).

Maybe the intent was that `/MOV "*.*"` is enough because it moves all files under the tree, but I’m worried it can leave source directories behind and report valid robocopy outcomes as failures. Unless there’s a reason to keep the shared helper here, I think directory moves should have their own robocopy path. wdyt?

listenarr.infrastructure/FileSystem/FileMover.cs
```diff
@@ -43,24 +45,19 @@ public partial class FileMover : IFileMover
...
-        public FileMover(ILogger<FileMover> logger, IProcessRunner? processRunner = null, IOptions<FileMoverOptions>? options = null)
-        {
-            _logger = logger;
-            _processRunner = processRunner;
-            _options = options?.Value ?? new FileMoverOptions();
-        }
-
         public async Task<bool> MoveDirectoryAsync(string sourceDir, string destDir)
         {
+            if (FileUtils.IsPathInsideOf(destDir, sourceDir))
```
therobbiedavis, line 55: Still concerned about one edge case here: `destDir == sourceDir` does not seem to be covered by the new inside-source guard. `IsPathInsideOf(destDir, sourceDir)` returns false for equality, and `Directory.Move(sourceDir, sourceDir)` can throw. That means we can reach the fallback, where `CopyDirRecursive` normalizes the paths, sees they are equal, returns without copying, and then `Directory.Delete(sourceDir, true)` deletes the source while `MoveDirectoryAsync` returns true.

Could we add an explicit same-path guard before the move/fallback path? Either no-op success or a safe failure would avoid that delete path.

tests/Features/Infrastructure/FileSystem/FileMoverTests.cs
```diff
@@ -0,0 +1,382 @@
...
+            var result = await _mover.PerformActionOn(FileAction.HardlinkCopy, sourceFile, destFile);
+
+            // Assert
+            Assert.True(result, "HardlinkFileAsync should succeed");
+            var destContent = await File.ReadAllTextAsync(destFile);
+            Assert.Equal("new content", destContent);
+        }
+
+        [Fact]
+        public async Task HardlinkFileAsync_FallbacksToCopy_WhenHardlinkFails()
```
therobbiedavis: Small test gap: this test name says it covers the copy fallback, but source and destination are in the same temp root, so on filesystems where hardlinks work it never actually exercises the fallback. Could we either force the hardlink operation to fail in a deterministic way or rename/scope this test so it does not imply fallback coverage?

therobbiedavis (changes requested):

tests/Features/Infrastructure/FileSystem/FileMoverTests.cs
T4g1: im not sure we should keep that test, different volume simulation is valid only on Windows anyway and i can't think of a reliable way to simulate hardlink failure (this test was only moved)

therobbiedavis: I trust your judgement.

listenarr.domain/Common/FileUtils.cs
```diff
@@ -377,25 +377,33 @@ public static string AppendSequenceSuffix(string desiredPath, int sequenceNumber
...
-                    + Path.DirectorySeparatorChar;
+            var relative = Path.GetRelativePath(parentPath, childPath);
 
-                return normalizedChild.StartsWith(normalizedRoot, StringComparison.OrdinalIgnoreCase);
-            }
-            catch (Exception caughtEx_3) when (caughtEx_3 is not OperationCanceledException && caughtEx_3 is not OutOfMemoryException && caughtEx_3 is not StackOverflowException)
-            {
-                return false;
-            }
+            return !relative.Equals(".") && !relative.StartsWith("..");
```
therobbiedavis: Won't this treat any relative path that is not "." and does not start with ".." as inside the parent. Also what about Windows paths? `Path.GetRelativePath("C:\\media\\source", "D:\\library\\dest")` would returns the absolute D:\... path, so I think this helper returns true for a destination on a different drive, but then `MoveDirectoryAsync` rejects it. I think we should add root checks and cover different drive paths in the test.

listenarr.infrastructure/FileSystem/FileMover.cs
```diff
@@ -109,48 +111,15 @@ public async Task<bool> MoveDirectoryAsync(string sourceDir, string destDir)
...
-                    }
-                }
-                catch (Exception rex) when (rex is not OperationCanceledException && rex is not OutOfMemoryException && rex is not StackOverflowException)
-                {
-                    _logger.LogWarning(rex, "Robocopy fallback threw an exception");
-                }
+                logger.LogError(ex, "Copy+delete fallback failed for directory {Source} -> {Dest}", sourceDir, destDir);
 
-                return false;
+                return await MoveWithRobocopy(sourceDir, destDir, "*.*");
```
therobbiedavis, line 122: I think this can leave source directories behind or report successful moves as failure because it only accepts exit code 1.

T4g1: I remember I had to change the exit code check because it was succeeding even when nothing was copied, I believe 1 is the only return code we can trust here for full completion. wdyt ?
Also, do you have evidence of this backup method being usefull ? I would expect robocopy to fail too if the initial move fails (why would robocopy succeed where move/library fails ?) I'd rather remove robocopy rather than adding more specific code path for it :x

therobbiedavis: So I dug in and verified this locally and validated that 1 is not quite enough with the current command. With /MOV "*.*" /E, robocopy returned 1 and copied the files, but left the source directory tree behind. So it proves files were copied, not that the directory move completed.

Robocopy can be useful in some Windows-specific cases, but here it only runs after Directory.Move retries and the managed copy/delete fallback have both failed, so I don’t have strong evidence that it’s buying us much at that point, just trying to be proactive instead of responsive because admittedly it's tough to account for multiple OS architectures. I’m fine with removing the directory robocopy fallback instead of adding more logic if you feel confident. My main concern is just not returning true for a partial move. I'd imagine at some point down the line we may consider a helper or something to validate that everything was moved properly instead of relying on returns or exit codes.

therobbiedavis (changes requested):
Once we resolve either keeping and fixing the robocopy with a post-condition check, or remove it altogether, then this can be approved.

therobbiedavis:
Closing as superseded by #717. #717 now covers #598s hardlink success, fallback-copy, source-preservation, and native-validation acceptance criteria, and replaces the underlying FileMover overlap, alias, publication, verified-fallback, and cleanup contracts with broader deterministic regression coverage. #717 references this PR and will close #598 when merged.

PR #610, opened 2026-05-21
Allow qBittorrent to determine hash ahead of time to avoid association errors

### LINKED ISSUE #609: qBittorrent: Trouble to link ongoing downloads with qBittorrent
Currently, the algortihm to link downloads in qBittorrent (QBT) with Listenarr downloads is the following:
* List all torrent's hashes in the QBT
* Add torrent in QBT
* List all torrent's hashes again
* New torrent hash is the diff between the two list of torrent hashes

This is sub-optimal and has some flaws:
* What if another app or the user adds torrent at between our two hash pollings ?
* We have reports of users that shows association issues that might be related to this

Therefore, I suggest to use a torrent library to read/compute the torrent hash ahead of sending it to QBT. This should solve some association issues by design

T4g1:
## Summary
* Closes #609 
Follows two reports of association errors on Discord support channel

### Changed
- Algorithm to match download in QBittorrent and Listennarr now use a more deterministic approach

## Testing
* Added a proper mock for qBittorrent API
* Added test data framework
* Added test on a real torrent file (public domain)

## Notes
As i was not able to reproduce the issue discussed on Discord (association errors when adding torrents to qBittorrent), this might not solve the real issue but it should be an improvement nonetheless


listenarr.api/Program.cs
```diff
@@ -380,11 +380,22 @@ ex is IOException
...
-    .ConfigurePrimaryHttpMessageHandler(() => new HttpClientHandler()
+    .ConfigureHttpClient(client =>
     {
-        AutomaticDecompression = System.Net.DecompressionMethods.All,
-        UseCookies = false
+        client.Timeout = TimeSpan.FromSeconds(30);
+    })
+    .ConfigurePrimaryHttpMessageHandler(sp =>
+    {
+        var cookieJar = sp.GetRequiredService<CookieContainer>();
```
therobbiedavis: This is not referenced, but called again in CookieContainer in line 396

listenarr.infrastructure/Adapters/QbittorrentAdapter.cs
```diff
@@ -240,6 +226,53 @@ async Task<HttpResponseMessage> PostLoginWithAgent(string userAgent)
...
+            [
+                new KeyValuePair<string, string>("username", client.Username ?? string.Empty),
+                new KeyValuePair<string, string>("password", client.Password ?? string.Empty)
+            ]);
+
+            using var loginResponse = await httpClient.PostAsync($"{baseUrl}/api/v2/auth/login", loginData, cancellationToken);
+            if (!loginResponse.IsSuccessStatusCode)
+            {
+                var body = await loginResponse.Content.ReadAsStringAsync(cancellationToken);
+                var redacted = LogRedaction.RedactText(body, LogRedaction.GetSensitiveValuesFromEnvironment().Concat([client.Password ?? string.Empty]));
```
therobbiedavis: This doesn't look be used

listenarr.infrastructure/Adapters/QbittorrentAdapter.cs
```diff
@@ -249,250 +282,162 @@ async Task<HttpResponseMessage> PostLoginWithAgent(string userAgent)
...
+            {
+                using (var stream = new MemoryStream(torrentFileData))
                 {
-                    if (client.Settings.TryGetValue("category", out var categoryObj))
-                        category = categoryObj?.ToString();
-                    if (client.Settings.TryGetValue("tags", out var tagsObj))
-                        tags = tagsObj?.ToString();
+                    var parser = new BencodeParser();
+                    Torrent torrent = parser.Parse<Torrent>(stream);
+                    hash = torrent.GetInfoHash();
```
therobbiedavis, line 314: There seems to be two different cases for the hash, it looks like most have .ToLowerInvariant, but there's one with .ToUpperInvariant. Can you look at the impact is of this discrepency?

listenarr.infrastructure/Adapters/QbittorrentAdapter.cs
```diff
@@ -61,24 +64,7 @@ public QbittorrentAdapter(IHttpClientFactory httpFactory, ITorrentFileDownloader
                 bool disposeHttp = false;
```
therobbiedavis: Isn't this going to make the conditional in line 133 `if (!disposeHttp)` always true?

Directory.Packages.props
```diff
@@ -33,5 +33,6 @@
 		<PackageVersion Include="xunit" Version="2.9.3" />
 		<PackageVersion Include="xunit.runner.visualstudio" Version="3.1.5" />
 		<PackageVersion Include="SQLitePCLRaw.bundle_e_sqlite3" Version="3.0.2" />
+		<PackageVersion Include="BencodeNET" Version="4.0.0" />
```
therobbiedavis, line 36: The latest version is 5.0.0. I see this is unlicensed however it doesn't look to be maintained anymore. Are we just using this to read torrent files? If so do we anticipate issues with v2 bittorrent? (e.g. https://github.com/Krusen/BencodeNET/issues/63#issuecomment-2870179442)

therobbiedavis (changes requested):
Looks good for the most part. We need to clean up dead code, and I had a question about BencodeNET

listenarr.infrastructure/Adapters/QbittorrentAdapter.cs
T4g1: I'll look at that but afaik, the hash is a SHA-1 hash of a particular bit of information from the torrent format. It's an hexadecimal representation so ideally we don't want upper/lower case to have any impact here (AF is the same as af)

T4g1: Ok so the case where it's upper cased is GetItemsAsync for DownloadClientItem which is dead code (we only use the QueueItem one)

The other usages for LoweCaseInvariant are 4:
* 2 unrelated ones
* 1 on DownloadId for GetImportItemAsync on DownloadClientItem (also unused)
* 1 in the MarkItemAsImportedAsync before sending data to the API: I do not believe the case matter for this API (and it shouldn't as per my last comment)

So I wouldn't mess with lower/upper case on this

Directory.Packages.props
T4g1: SHA-1 is an old algortihm that is flawed as there could be collisions for some torrent files. I believe the change with the v2 would be to use SHA-256 or something to avoid those. When it's the case, qBittorrent API will probably change to allow to use those SHA-256 hash instead of the SHA-1 so we will need to update this adapter in any case

T4g1: For Bencode, i didnt saw its stalled but we could fork it if needed to fix eventual security issue i guess. Also decoding only the specific bit required to compute the SHA-1 hash might not be very difficult so we could also use our own partial implementation of it

T4g1: Just tested the 5.0.0 of Bencode, the hash determination is broken, for a specific big-buck-bunny.torrent it produces a hash that correspond to the text "google.com", i leave 4.0.0 because of that :x

listenarr.infrastructure/Adapters/QbittorrentAdapter.cs
T4g1: Indeed, i updated that bit

listenarr.api/Program.cs
```diff
@@ -380,11 +380,20 @@ ex is IOException
 
 // Register named HttpClients for each adapter type so adapter implementations can request the appropriately-configured client.
 // qbittorrent
+builder.Services.AddSingleton<CookieContainer>();
```
therobbiedavis: My suspicion is that this could cause errors for those running multiple instances under the same host, like a docker user would. What about moving this into a qBittorrent session manager keyed client.Id + normalized baseUrl + username?

listenarr.infrastructure/Adapters/QbittorrentAdapter.cs
```diff
@@ -249,250 +254,162 @@ async Task<HttpResponseMessage> PostLoginWithAgent(string userAgent)
...
-                        }
-                    }
-                    else
-                    {
-                        _logger.LogWarning("qBittorrent login failed: {Status} - {Body}", loginResponse.StatusCode, redacted);
-                        throw new Exception($"qBittorrent login failed with status {loginResponse.StatusCode}");
-                    }
-                }
-                else
+            byte[]? torrentFileData = result.TorrentFileContent;
```
therobbiedavis, line 284: I think this regresses URL-only torrent adds. If pre-download returns Empty, we still have a valid httpTorrentUrl, but hash remains empty and AddAsync returns null before reaching the URL add branch. qBittorrent supports adding http/https torrent URLs directly, and canary handled this by adding the URL and then detecting the new hash from torrents/info afterward. I am not sure the best way to handle this with the new direction here though. wdyt?

listenarr.infrastructure/Adapters/QbittorrentAdapter.cs
therobbiedavis: I think there might be some more dead code in the adapter like TryPredownloadTorrentFileAsync(), but it's not a blocker. If you get a chance, we should probably also try to clean up the files as we touch them.

therobbiedavis (changes requested):
Great job, I still have some hesitation around the cookies and url only torrent adds but I think this is looking better.

Sidenote: Let's make sure we don't commit an impermissible torrent to the repo 😆

listenarr.infrastructure/Adapters/QbittorrentAdapter.cs
T4g1: I agree but as I already have a whole PR dedicated to cleaning the adapters (https://github.com/Listenarrs/Listenarr/pull/599) i'd rather not change too much here

listenarr.infrastructure/Adapters/QbittorrentAdapter.cs
T4g1: Indeed, if we can't get the torrent file, we assume qBittorrent will not be able to get it either, hence the branch to add with Magnet/URL will only perform the magnet add
Do you think we should handle the URL case even though we are unable to get the torrent ourselves ? In that case, we re-introduce the issue where we cannot garantee that the hash can be found afterwards

therobbiedavis: Well with the new library, couldn't we:
- download the torrent locally in a temp folder
- parse it for the hash
- store the hash
- send it to qbit 
- delete from temp

Or am I thinking about this incorrectly?

T4g1: We do not use the new library to download the torrent file but using mostly what you already had in place, we already have it in torrentFileData which i use to construct a torrent object that can then be used to get the hash. It's all done in memory so no need to store the file and delete it afterwards

listenarr.infrastructure/Adapters/QbittorrentAdapter.cs
```diff
@@ -249,250 +254,162 @@ async Task<HttpResponseMessage> PostLoginWithAgent(string userAgent)
...
-                    {
-                        // Indexer redirected to a magnet link — use it as the torrent URL instead
-                        magnetLink = DownloadClientUriBuilder.NormalizeMagnetLink(downloadResult.MagnetUri);
-                        _logger.LogInformation("Indexer redirected to magnet link for '{Title}'", LogRedaction.SanitizeText(result.Title));
-                    }
-                }
+            if (string.IsNullOrEmpty(hash))
+            {
+                _logger.LogError($"Unable to compute hash for the given torrent: {result.Title} with torrent URL: {result.TorrentUrl} and magnet link: {result.MagnetLink}");
+                return null;
```
therobbiedavis, line 326: Double check the downloadservice, because since we return null here, we should probably throw in the AddAsync in DownloadService to treat null as a failure.

listenarr.infrastructure/Adapters/QbittorrentAdapter.cs
```diff
@@ -249,250 +254,162 @@ async Task<HttpResponseMessage> PostLoginWithAgent(string userAgent)
...
-                        _logger.LogWarning("qBittorrent login failed: {Status} - {Body}", loginResponse.StatusCode, redacted);
-                        throw new Exception($"qBittorrent login failed with status {loginResponse.StatusCode}");
-                    }
-                }
-                else
+            byte[]? torrentFileData = result.TorrentFileContent;
+            if (torrentFileData == null && !string.IsNullOrEmpty(httpTorrentUrl))
+            {
+                // Try to get torrent file with the torrent URL
+                var downloadResult = await _torrentFileDownloader.DownloadAsync(httpTorrentUrl, ct);
```
therobbiedavis, line 288: I think there's a small edge case here: this calls `_torrentFileDownloader.DownloadAsync(...)` directly. The downloader returns `Empty` for expected HTTP/content failures, but network/timeouts/DNS/read exceptions can still escape before we try the existing `magnetLink`. Since `TryPredownloadTorrentFileAsync(...)` already wraps this and returns `Empty`, can we use that here? Then URL-only failures still fail as intended, while URL+magnet still falls back to the magnet.

therobbiedavis (changes requested):

listenarr.infrastructure/Adapters/QbittorrentAdapter.cs
T4g1: We have this already:
```
// Route to appropriate client handler via adapter and capture client-specific IDs when provided
string? clientSpecificId = await clientGateway.AddAsync(downloadClient, searchResult);
clientSpecificId ??= TryResolveClientSpecificIdFallback(downloadClient, searchResult);

// Update download record with client-specific ID if available
if (!string.IsNullOrEmpty(clientSpecificId))
```

But that whole part should be rewrittent soon as the DownloadService should not perist a download before that call (in fact, it's the download gateway that should handle that)

listenarr.infrastructure/Adapters/QbittorrentAdapter.cs
T4g1: Ill give that responsability to the TorrentFileDownloader instead of re-implementing it in the adapter

tests/Mocks/Api/QbittorrentApiMock.cs
```diff
@@ -0,0 +1,77 @@
...
+            Authenticated = true;
+            var response = MockUtils.GetCannedResponse("Ok");
+            response.Headers.Add("Set-Cookie", "SID=1; HttpOnly; Path=/");
+
+            return response;
+        }
+
+        private async Task<HttpResponseMessage> DoAdd(HttpRequestMessage request, CancellationToken ct)
+        {
+            if (!Authenticated)
```
therobbiedavis, line 42: Optional: The current mock accepts add requests based on a global Authenticated flag, so the tests verify login ordering but not the new cookie-propagation behavior.

therobbiedavis (approved):

listenarr.infrastructure/Adapters/QbittorrentAdapter.cs
therobbiedavis: add a TODO, and we'll handle it later.

PR #634, opened 2026-05-31
Fix NZBGet import completion

### LINKED ISSUE #619: Bug: NZBGet downloads are never imported — `FetchDownloadsAsync` reads numeric `listgroups` fields (`FileSizeMB`/`RemainingSizeMB`) with `.GetString()`

therobbiedavis:
## Summary

Fixes #619 and follows up #628.

This PR keeps the original NZBGet `listgroups` JSON-number fix from #628, then completes the larger import path: NZBGet downloads that have already left the active queue can now be reconciled from history, marked complete, and queued for import with a usable source path.

## Changed

- `NzbgetAdapter.FetchDownloadsAsync` now reads `FileSizeMB` and `RemainingSizeMB` as JSON numbers or numeric strings.
- Active NZBGet queue progress is now reported on the 0-100 scale expected by `AdapterUtils.MapDownloadProgress`.
- NZBGet history matching now uses the stored `ClientDownloadId` / NZBGet `NZBID` instead of the history row `ID`.
- Completed history items now prefer `FinalDir` and fall back to `DestDir` when resolving the import source path.
- NZBGet history reconciliation now targets only tracked NZB IDs instead of materializing every history entry for every poll.
- Active `listgroups` NZB IDs are excluded from the history pass, so normal active progress polling does not also scan history.
- History reconciliation is now best-effort: if NZBGet history is temporarily unavailable, active queue updates are still returned and persisted.
- NZBGet queue items no longer expose the client category as their display quality.
- `.github/CLAUDE.md` no longer instructs AI agents to add changelog entries for meaningful changes.

## Fixed

- Fixed the original `listgroups` JSON type issue where numeric size fields could throw during polling, while preserving support for numeric strings from proxies or mocks.
- Fixed a follow-up robustness gap where string-valued active queue size fields could deserialize as `0` and skip progress updates.
- Fixed active progress being calculated as a 0-1 fraction, which made 50% complete downloads show as `0.5%` instead of `50%`.
- Fixed the broader NZBGet import issue where completed items disappeared from `listgroups` before Listenarr could mark them `Completed` and enqueue import processing.
- Fixed completed NZBGet imports resolving against the history row `ID` instead of the actual NZB download ID.
- Fixed completed imports losing the resolved source path before the processing job is created.
- Fixed a follow-up polling regression where a history endpoint failure could have caused the whole NZBGet poll to fail even after active queue updates were read successfully.
- Fixed the Activity view showing the NZBGet category, such as `audiobooks`, in the Quality column.

## Added

- Added shared NZBGet history parsing/resolution helpers for `NZBID`, `Status`, `NZBName`, `FinalDir`, and `DestDir`.
- Added a JSON double helper for active queue size fields that accepts both JSON numbers and numeric strings.
- Added a monitor-level regression test proving a completed NZBGet history item queues a processing job with the resolved path.
- Added coverage for numeric and string-valued `listgroups` progress mapping, including the expected `50M` progress value.
- Added coverage proving active downloads and empty/no-eligible polls skip NZBGet history lookup.
- Added coverage proving active queue updates still return when the history endpoint fails.
- Added coverage proving NZBGet category filtering still works while queue-item quality stays empty.
- Added NZBGet mock support for distinct history `ID` vs `NZBID`, active queue toggling, completed content paths, history request counts, simulated history failures, and string-valued active queue sizes.

## Removed

- Removed the old assumption that NZBGet history `ID` is the same identifier stored on tracked downloads.
- Removed the old single-shape parsing assumption for NZBGet `FileSizeMB` / `RemainingSizeMB` in `FetchDownloadsAsync`.
- Removed unnecessary history requests when there are no eligible tracked downloads to reconcile.
- Removed unnecessary history reconciliation for downloads already present in the active NZBGet queue.
- Removed the unreleased `0.2.72` changelog entry for this PR.
- Removed the AI-agent changelog-maintenance instructions from `.github/CLAUDE.md`.
- No user-facing features were removed.

## Testing

- `dotnet test tests/Listenarr.Tests.csproj --filter FullyQualifiedName~NzbgetAdapterTests` - 8 passed
- `dotnet test tests/Listenarr.Tests.csproj --filter FullyQualifiedName~DownloadClientAdapterTests` - 7 passed
- `dotnet test tests/Listenarr.Tests.csproj --filter "FullyQualifiedName~UsenetAdapterFilteringTests|FullyQualifiedName~NzbgetAdapterTests"` - 10 passed
- `dotnet test tests/Listenarr.Tests.csproj --no-build` - 677 passed
- pre-push hook: backend format/layering checks passed
- pre-push hook: frontend type check passed
- pre-push hook: frontend tests passed - 361 passed, 1 skipped

Note: backend restore/build still reports the pre-existing `NU1902` advisory for `SharpCompress 0.47.4`.


.github/CLAUDE.md
```diff
@@ -56,28 +56,4 @@ As a security-aware developer, generate secure .NET code using ASP.NET Core that
 - Use provide/inject for deep component communication
 - Use async components for code-splitting
 
----
```
T4g1, line 59: fyi: If we are more rigorous regarding the commit messages, we could parse them to generate the changelog automatically. In Jira, I also did some proof of concept to manage changelog based on ticket/issue metadata that may be done too with Github issue/PR

tests/Mocks/Api/NzbgetApiMock.cs
```diff
@@ -7,14 +7,37 @@ public class NzbgetApiMock : BaseApiMock
...
+
+            if (FailHistoryRequests)
+            {
+                return new HttpResponseMessage(System.Net.HttpStatusCode.InternalServerError)
+                {
+                    Content = new StringContent("history unavailable")
+                };
+            }
+
             var response = """
```
T4g1, line 41: Maybe we can simplify this a bit with the $$""" notation (i was not aware of it's existence before the previous PR)

tests/Mocks/Api/NzbgetApiMock.cs
```diff
@@ -56,9 +131,53 @@ private async Task<HttpResponseMessage> GetHistory(HttpRequestMessage request, C
...
+            response = response.Replace("{{COMPLETED_FILE_NZBGET}}", COMPLETED_FILE_NZBGET);
+            response = response.Replace("{{COMPLETED_FILE_HISTORY_ID}}", COMPLETED_FILE_HISTORY_ID);
             response = response.Replace("{{ARBITRARY_PATH_1}}", FileUtils.GetAbsolutePath("nzbget", "completed", "Book.m4b"));
             response = response.Replace("{{ARBITRARY_PATH_2}}", FileUtils.GetAbsolutePath("nzbget", "completed", "Book Folder"));
+            response = response.Replace("{{COMPLETED_CONTENT_PATH}}", CompletedContentPath);
+            response = response.Replace("{{COMPLETED_DEST_DIR}}", Path.GetDirectoryName(CompletedContentPath) ?? string.Empty);
             return MockUtils.GetCannedResponse(response, "text/xml");
         }
+
+        private async Task<HttpResponseMessage> GetJsonrpc(HttpRequestMessage request, CancellationToken ct)
```
T4g1, line 143: Another suggestion but if NZBGet allows to do the same thing with XML and JSON RPC, we probably want to stick to one unless we have good reasons to keep both

tests/Features/Infrastructure/Adapters/NzbgetAdapterTests.cs
```diff
@@ -16,18 +16,39 @@
...
 using Microsoft.Extensions.Logging.Abstractions;
 using Moq;
 using Xunit;
 
 namespace Listenarr.Tests.Features.Infrastructure.Adapters
 {
-    public class NzbgetAdapterTests
+    public class NzbgetAdapterTests : BaseTests
     {
+        private DownloadClientConfiguration? _client;
```
T4g1, line 35: i believe you can also use the null! notation here if you want

listenarr.infrastructure/Adapters/NzbgetAdapter.cs
```diff
@@ -1014,6 +994,198 @@ private async Task<byte[]> DownloadNzbAsync(string nzbUrl, string? indexerApiKey
             return null;
         }
 
+        private sealed record NzbgetHistoryItem(string NzbId, string Name, string Status, string ContentPath);
```
T4g1, line 997: I would put this at the start or at the end of the class declaration (depending on which standard you want to follow) but not in the middle of it

listenarr.infrastructure/Adapters/NzbgetAdapter.cs
```diff
@@ -1014,6 +994,198 @@ private async Task<byte[]> DownloadNzbAsync(string nzbUrl, string? indexerApiKey
...
+
+        private static IEnumerable<IReadOnlyDictionary<string, string>> EnumerateXmlRpcStructMembers(XElement? arrayData)
```
T4g1, line 999: This should be in a XML utils file probably (it will probably be handy for Sabnzbd too)

listenarr.infrastructure/Adapters/NzbgetAdapter.cs
```diff
@@ -1014,6 +994,198 @@ private async Task<byte[]> DownloadNzbAsync(string nzbUrl, string? indexerApiKey
...
+            return status.StartsWith("SUCCESS", StringComparison.OrdinalIgnoreCase);
+        }
+
+        private static bool IsFailedHistoryStatus(string status)
+        {
+            return status.StartsWith("FAILURE", StringComparison.OrdinalIgnoreCase) ||
+                status.StartsWith("FAILED", StringComparison.OrdinalIgnoreCase);
+        }
+
+        private static double GetJsonDoubleOrDefault(JsonElement element, string propertyName, double defaultValue = 0d)
```
T4g1, line 1099: Why not reuse GetPropertyOrDefault in JsonExtensions ?

listenarr.infrastructure/Adapters/NzbgetAdapter.cs
```diff
@@ -1181,10 +1330,15 @@ public async Task<List<Download>> FetchDownloadsAsync(
                                     try
                                     {
                                         var nzbId = group.TryGetProperty("NZBID", out var nzbIdProp) ? nzbIdProp.GetInt32() : 0;
+                                        if (nzbId > 0)
```
T4g1, line 1333: Are you sure 0 is an invalid NZBID ? (i dont see anything about that here: https://nzbget.com/documentation/api/listgroups/)

listenarr.infrastructure/Adapters/NzbgetAdapter.cs
```diff
@@ -1218,6 +1370,15 @@ public async Task<List<Download>> FetchDownloadsAsync(
                     }
                 }
 
+                try
+                {
+                    await ApplyHistoryUpdatesAsync(client, downloads, activeNzbIds);
+                }
+                catch (Exception ex) when (ex is not OperationCanceledException && ex is not OutOfMemoryException && ex is not StackOverflowException)
```
T4g1, line 1377: What do you think of `catch (Exception exception) when (exception is not (OperationCanceledException or OutOfMemoryException or StackOverflowException))` instead ?

listenarr.infrastructure/Adapters/NzbgetAdapter.cs
```diff
@@ -1014,6 +994,198 @@ private async Task<byte[]> DownloadNzbAsync(string nzbUrl, string? indexerApiKey
...
+                .Where(structElement => structElement != null))
+            {
+                yield return structElement!.Elements("member").ToDictionary(
+                    m => m.Element("name")?.Value ?? string.Empty,
+                    m => m.Element("value")?.Elements().FirstOrDefault()?.Value ?? string.Empty,
+                    StringComparer.Ordinal);
+            }
+        }
+
+        private static string ResolveHistoryContentPath(IReadOnlyDictionary<string, string> members)
```
T4g1, line 1017: Its not obvious what this method does with only the prototype, maybe we want to document what is the members parameter or what it does

T4g1 (changes requested):
Looks good! Only some minor change request and suggestions

kevinheneveld:
**Adjacent gap: `DELETED/*` history statuses still strand downloads in Queued**

Not asking to expand this PR's scope — flagging for a possible follow-up, since it lives in the exact reconciliation path this PR adds. (Independently verified the SUCCESS path here works: deployed an equivalent fix and watched 6 stranded grabs go Queued→Moved with real files imported — so this isn't a critique, just an adjacent gap found in the same path on a production instance.)

NZBGet's built-in DupeCheck removes a redundant grab from the active queue and records it in history as `DELETED/COPY` (also `DELETED/DUPE`, `DELETED/GOOD`); manual/scan removals are `DELETED/MANUAL`/`DELETED/SCAN`; corrupt/unhealthy removals are `DELETED/HEALTH`/`DELETED/BAD`. None of these start with `SUCCESS` or `FAILURE`, so with this PR's `IsSuccessfulHistoryStatus` / `IsFailedHistoryStatus` branches a `DELETED/*` item hits neither branch and the tracked download stays stuck in `Queued` indefinitely — the same orphaned-in-Queued symptom this PR fixes for the success/fail cases, still present for the deleted case.

This is not rare in practice. Any title grabbed more than once (re-search, manual + auto, etc.) trips DupeCheck: on one instance **147 of 501** history items were `DELETED/COPY`, leaving **419** tracked downloads stuck. (Ironically, fixing the SUCCESS path here *reduces* future dupes, because books finally import instead of being re-grabbed — but the existing backlog and any concurrent re-grabs still produce `DELETED/COPY`.)

**Suggested handling** (the nuance is *block* vs *fail*):
- `DELETED/COPY` | `DELETED/DUPE` | `DELETED/GOOD` → mark **ImportBlocked** (terminal), **not Failed**. With `FailedDownloadHandlingEnabled` + `FailedDownloadAutoSearch` on, failing a dupe fires an auto-search that re-grabs the same release, which NZBGet deletes again → a churn loop. Blocking is terminal and doesn't auto-search.
- `DELETED/HEALTH` | `DELETED/BAD` → **Fail** (genuinely bad; normal retry/auto-search should apply).
- `DELETED/MANUAL` | `DELETED/SCAN` | other `DELETED/*` → **block** (ambiguous removal; terminal, don't re-search).

Two implementation notes from doing this elsewhere:
1. `Queued → ImportBlocked` is an invalid state-machine transition (only reachable from Downloading/Processing/ImportPending/Completed), so a still-queued dupe needs a nudge `Queued → Downloading → ImportBlocked`.
2. Blocking alone doesn't fully stop the churn: `AutomaticSearchService` treats `Queued` as "active → skip re-search" but **not** `ImportBlocked`, so moving a dupe Queued→ImportBlocked actually *lifts* the accidental suppression the stuck-Queued state was giving it. To close the loop, the background auto-search cycle should also treat `ImportBlocked` as blocking (while leaving manual "Search now" able to override).

Happy to open a separate PR for this if useful, or it can fold into this one — your call on scope.

therobbiedavis:
Closing this as superseded by #599.

The NZBGet import-completion fixes from this PR have been folded into the broader download-client architecture work in #599, including targeted polling, history/final-path resolution, and adapter workflow slicing. Keeping #599 as the single merge path avoids maintaining two overlapping NZBGet fixes.

PR #641, opened 2026-06-01
fix(downloads): purge old processing jobs via a hosted cleanup worker

kevinheneveld:
Addresses the **3a (job retention)** portion of #640.

`IDownloadProcessingJobService.CleanupOldJobsAsync(retentionDays)` already existed (and is implemented in `EfDownloadProcessingJobRepository`) but had **no caller**, so `DownloadProcessingJobs` grew unbounded on long-running instances — and every queue-snapshot reconciliation that queries this table pays for the bloat.

This adds `DownloadProcessingJobCleanupService`, a hosted `BackgroundService` that purges terminal (Completed/Failed) jobs older than a 7-day retention window shortly after startup and then daily.

**Notes**
- Retention is a constant (mirrors the existing `CleanupOldJobsAsync` default) to keep this change migration-free; making it configurable is an easy follow-up.
- Cleanup logic is isolated in an internal `RunCleanupAsync` so it can be exercised directly in tests without driving the background loop.
- Registered alongside the other hosted services in `HostedServiceRegistrationExtensions`.

**Tests**
- Terminal jobs past the retention window are removed while recent and non-terminal jobs are retained.
- No-op when nothing is eligible.
- Service is registered as a hosted service.

Does **not** touch the 3b (orphan terminalization) part of #640 — that's a separate, more behavioral change and is left for discussion on the issue.


listenarr.application/Downloads/DownloadProcessingJobCleanupService.cs
T4g1, line 1: Why not handle the cleaning in the DownloadProcessingJobProcessor ?

listenarr.application/Downloads/DownloadProcessingJobCleanupService.cs
```diff
@@ -0,0 +1,108 @@
...
+        /// Runs a single cleanup pass. Internal so it can be exercised directly in tests
+        /// without driving the background loop.
+        /// </summary>
+        internal async Task RunCleanupAsync(CancellationToken cancellationToken)
+        {
+            try
+            {
+                using var scope = _scopeFactory.CreateScope();
+                var jobService = scope.ServiceProvider.GetRequiredService<IDownloadProcessingJobService>();
+                await jobService.CleanupOldJobsAsync(RetentionDays);
```
T4g1: That method should be update so the actual cleaning logic does not stays in the infrastructure layer

T4g1 (changes requested):

listenarr.application/Downloads/DownloadProcessingJobCleanupService.cs
kevinheneveld: Good call — done in a1fbb946. Dropped the standalone `DownloadProcessingJobCleanupService` and folded the daily purge into `DownloadProcessingJobProcessor`, which already owns job-table maintenance (the startup `ResetStuckJobsAsync`). One worker instead of two.

listenarr.application/Downloads/DownloadProcessingJobCleanupService.cs
kevinheneveld: Agreed — also in a1fbb946. The retention policy (terminal statuses + age cutoff) now lives in the application `DownloadProcessingJobService`; the repository method is reduced to a thin `DeleteCompletedBeforeAsync(statuses, cutoffUtc)` that just runs the delete it's handed.

kevinheneveld:
Thanks for the review @T4g1 — both points addressed in a1fbb946 (added on top, no force-push):

- **Folded cleanup into `DownloadProcessingJobProcessor`** instead of a separate hosted service, since that worker already owns job-table maintenance (`ResetStuckJobsAsync` on startup). Removed `DownloadProcessingJobCleanupService` and its registration.
- **Moved the retention policy out of the infrastructure layer** — terminal statuses + cutoff now live in the application `DownloadProcessingJobService`; the repo method is a thin `DeleteCompletedBeforeAsync(statuses, cutoffUtc)`.

Behavior is unchanged (terminal jobs older than 7 days purged shortly after startup, then daily). Cleanup tests rehomed to drive the processor end-to-end through the moved policy.

therobbiedavis:
Closing this as superseded by #599.

Thank you for identifying and implementing the processing-job retention cleanup direction here. #599 incorporates that cleanup purpose into the broader download-client architecture, but uses the final structure chosen for that PR: a separate hosted cleanup processor so import execution and retention maintenance remain separate.

I also added explicit credit in #599 so the contribution is visible there.

Issue #647, opened 2026-06-04
Split PreferredFormats into PreferredContainers (container/extension) vs codec/bitrate in Qualities


timk75:
## Summary

Split the `QualityProfile.PreferredFormats` field into a container/extension concern (`PreferredContainers`) and let codec/bitrate/lossless live solely in `Qualities`. Today `PreferredFormats` defaults to `{ "m4b", "mp3", "m4a", "flac", "opus" }`, which mixes **containers** (m4b, m4a) with **codecs** (mp3, flac, opus). This conflation makes format filtering and quality matching ambiguous.

This is the agreed follow-up to the profile-driven quality matcher (the M4B re-grab loop fix). That PR deliberately left `PreferredFormats` untouched because this split carries an EF migration and a frontend change that warrant their own review.

## Proposed shape

- `PreferredContainers`: container/extension tokens only — `m4b, m4a, mp3, flac, ogg, opus, …` — used strictly for format/extension filtering.
- `Qualities`: continue to carry codec + bitrate + lossless (`AAC 128kbps`, `MP3 320kbps`, `FLAC`, `OGG Vorbis 192kbps`, …), which `QualityMatcher` already consumes.

## Scope / touch points

**Backend**
- `listenarr.domain/Models/QualityProfile.cs` — rename/replace `PreferredFormats` → `PreferredContainers` (treat strictly as container/extension tokens).
- `listenarr.infrastructure/Persistence/Configurations/QualityProfileConfiguration.cs` — EF mapping + a migration that **copies existing `PreferredFormats` data** into the new column (don't drop user data).
- `listenarr.application` — status evaluator candidate-filter, `SearchResultScorer`, `QualityProfileService`, and the automatic-search format detection that read `PreferredFormats`.

**Frontend**
- `fe/src/types/index.ts` (`QualityProfile.preferredFormats`), `QualityProfileFormModal.vue`, `QualityProfilesTab.vue`, `audiobookStatus.ts`, plus the Vitest specs.

## Notes
- Migration must be reviewed on its own (data copy + backfill).
- `QualityMatcher` already separates codec/bitrate matching from container handling, so the matcher itself needs no change — this issue is about the container field and its data migration.

timk75:
I'm happy to implement this. I'd prefer to wait until #648 (the profile-driven matcher) is reviewed/approved first, so the container/codec split + EF migration build on an agreed foundation rather than stacking unreviewed changes. Will pick it up once #648 lands.

PR #648, opened 2026-06-04
Fix M4B re-grab loop with profile-driven quality matching

timk75:
## Summary

Fixes the **M4B re-grab loop**: a monitored audiobook whose file is already at/above cutoff is searched and re-downloaded on every cycle. Replaces #581 with the profile-driven matching approach we agreed on.

**Root cause.** The cutoff check classified an on-disk file into an ad-hoc label — the container `"M4B"`, or a bare `"256kbps"` — and looked that string up in the profile's quality rungs. Real rungs are named by codec + bitrate (`"AAC 256kbps"`), so the label never matched, `cutoffMet` stayed false, and the audiobook was re-grabbed forever. The same string-label logic existed in **three** places with **three different vocabularies**: `AudiobookStatusEvaluator` (status display), `AutomaticSearchService` (the actual re-grab driver), and a near-duplicate in `LibraryController`.

### Added
- **`QualityMatcher`** (`listenarr.domain/Common`): a pure, dependency-free matcher. Given a file's codec/bitrate/lossless facts and a profile, it walks `profile.Qualities` using the structured `Codec` / `Bitrate` / `IsLossless` fields — falling back to parsing the rung label for seed/legacy profiles that only set `Quality` + `Priority` — and: **round-down** (picks the highest rung the file meets or exceeds); returns a defined **`CodecMismatch`** when the file's codec isn't in the profile (instead of the old `int.MaxValue` silent fallback); matches OGG Vorbis by codec, maps AAC-in-M4B/M4A (incl. the legacy `M4B` codec group), and normalizes bits/sec → kbps in one place.
- 26 new unit tests (`QualityMatcherTests`), incl. the exact M4B-loop regression; `QualityProfileBuilder` extended.

### Changed
- The status evaluator, automatic search, and library controller now share a single **`QualityCutoffEvaluator`** that calls the matcher.
- **Inverted priority comparison corrected.** `AutomaticSearchService`'s cutoff check (`Priority >= cutoff`) and `IsQualityBetter` (`cand.Priority > exist.Priority`) compared rungs the wrong way round given the model's "lower number = higher quality" (matching the seed profiles, FE, and status evaluator). They now use the canonical direction.

### Fixed
- The M4B re-grab loop (see Summary).
- **Path-only / metadata-less files** are no longer dropped before matching. `AudiobookStatusEvaluator` now forwards `Path` into `AudioQualityInput` (as `QualityCutoffEvaluator` does), and its candidate-format pre-filter falls back to the path extension when `Format`/`Container` are empty — so e.g. a tagless `book.m4b` with `PreferredFormats = ["m4b"]` reaches the matcher instead of returning quality-mismatch.

### Removed
- ~380 lines of duplicated, drifting string-label detection logic across the three call sites.
- The frontend's client-side status recomputation in `audiobookStatus.ts`; the FE now trusts the server-computed status (overriding only for the transient "download in flight" case).

## Testing

- Full backend suite green (**711 tests**), including the M4B-loop regression and the path-only fallback cases (FLAC-by-extension and a second path-only `m4b` case proving the fallback is format-agnostic).
- Frontend: 364 tests + type-check + eslint green.

## Notes

**Behaviour changes to confirm:**
1. **Inverted priority comparison corrected** — this changes grab/skip decisions: a result is grabbed only when genuinely better than the existing file, and cutoff is met only when the file is at least cutoff quality.
2. **Codec absent from profile ⇒ cutoff not met** ⇒ such files keep being searched (the logical inverse of the old accidental "always at max priority" behaviour).
3. **Round-down floor:** a file below the lowest rung (or unknown bitrate with no VBR rung) maps to the worst rung for its codec — conservative, never over-claims.

**Out of scope (deliberate):** the `PreferredFormats → PreferredContainers` split + EF migration + frontend is tracked separately in #647. `DownloadImportService.IsQualityBetter` is left untouched — it's a different, profile-independent "acceptable downgrade?" gate, not the inverted-priority bug.

Closes #581.
Follow-up: #647.

### DIFF
--- a/fe/src/utils/audiobookStatus.ts
+++ b/fe/src/utils/audiobookStatus.ts
@@ -15,7 +15,7 @@
  * You should have received a copy of the GNU Affero General Public License
  * along with this program. If not, see <https://www.gnu.org/licenses/>.
  */
-import type { Audiobook, AudiobookStatus, QualityProfile } from '@/types'
+import type { Audiobook, AudiobookStatus } from '@/types'
 
 const AUDIOBOOK_STATUSES: AudiobookStatus[] = [
   'downloading',
@@ -23,8 +23,6 @@ const AUDIOBOOK_STATUSES: AudiobookStatus[] = [
   'quality-mismatch',
   'quality-match',
 ]
-
-const normalize = (value?: string): string => (value || '').toString().trim().toLowerCase()
 
 function isAudiobookStatus(value: unknown): value is AudiobookStatus {
   return typeof value === 'string' && AUDIOBOOK_STATUSES.includes(value as AudiobookStatus)
@@ -45,69 +43,24 @@ export function formatAudiobookStatus(status: AudiobookStatus): string {
   }
 }
 
+/**
+ * Resolve the display status for an audiobook.
+ *
+ * The backend is the single source of truth for quality matching: the `/library`
+ * list endpoint runs {@link AudiobookStatusEvaluator.ComputeStatus} server-side and
+ * returns the result as `audiobook.status`. The frontend trusts that value and only
+ * applies a client-side override for transient state the server's snapshot can't see —
+ * an in-flight download for this audiobook.
+ *
+ * This deliberately no longer mirrors the backend QualityMatcher client-side; that
+ * duplicated business logic and could disagree with the automatic-search cutoff.
+ */
 export function computeAudiobookStatus(
-  audiobook: Audiobook,
+  audiobook: Pick<Audiobook, 'id' | 'status'>,
   activeDownloadAudiobookIds: ReadonlySet<number>,
-  qualityProfiles: QualityProfile[],
 ): AudiobookStatus {
   if (activeDownloadAudiobookIds.has(audiobook.id)) {
     return 'downloading'
-  }
-
-  const hasFiles = Array.isArray(audiobook.files) && audiobook.files.length > 0
-  if (hasFiles) {
-    const profile = qualityProfiles.find((item) => item.id === audiobook.qualityProfileId)
-
-    if (!profile) {
-      const hasFileSummary =
-        !!(audiobook.filePath && audiobook.fileSize && audiobook.fileSize > 0) || hasFiles
-      return hasFileSummary ? 'quality-match' : 'no-file'
-    }
-
-    const preferredFormats = (profile.preferredFormats || []).map((item) => normalize(item))
-    const candidateFiles = audiobook.files!.filter((file) => {
-      if (!file) return false
-      const fileFormat = normalize(file.format) || normalize(file.container) || ''
-      if (preferredFormats.length === 0) return true
-      return (
-        preferredFormats.includes(fileFormat) ||
-        preferredFormats.some((preferredFormat) => fileFormat.includes(preferredFormat))
-      )
-    })
-
-    if (candidateFiles.length === 0) {
-      return 'quality-mismatch'
-    }
-
-    if (!profile.cutoffQuality || !profile.qualities || profile.qualities.length === 0) {
-      return 'quality-match'
-    }
-
-    const qualityPriority = new Map<string, number>()
-    for (const quality of profile.qualities) {
-      if (!quality || !quality.quality) continue
-      qualityPriority.set(normalize(quality.quality), quality.priority)
-    }
-
-    const cutoff = normalize(profile.cutoffQuality)
-    const cutoffPriority = qualityPriority.has(cutoff)
-      ? qualityPriority.get(cutoff)!
-      : Number.POSITIVE_INFINITY
-
-    for (const file of candidateFiles) {
-      const derivedQuality = deriveQualityLabel(audiobook, file)
-      if (!derivedQuality) continue
-
-      const priority = qualityPriority.has(derivedQuality)
-        ? qualityPriority.get(derivedQuality)!
-        : Number.POSITIVE_INFINITY
-
-      if (priority <= cutoffPriority) {
-        return 'quality-match'
-      }
-    }
-
-    return 'quality-mismatch'
   }
 
   if (isAudiobookStatus(audiobook.status)) {
@@ -115,54 +68,4 @@ export function computeAudiobookStatus(
   }
 
   return 'no-file'
-}
-
-function deriveQualityLabel(
-  audiobook: Audiobook,
-  file:
-    | {
-        bitrate?: number
-        container?: string
-        codec?: string
-        format?: string
-      }
-    | undefined,
-): string {
-  if (audiobook.quality) return normalize(audiobook.quality)
-
-  if (file && file.bitrate) {
-    const bitrate = Number(file.bitrate)
-    if (!Number.isNaN(bitrate)) {
-      const bitrateKbps = bitrate >= 1000 ? bitrate / 1000 : bitrate
-      if (bitrateKbps >= 320) return '320kbps'
-      if (bitrateKbps >= 256) return '256kbps'
-      if (bitrateKbps >= 192) return '192kbps'
-      return `${Math.round(bitrateKbps)}kbps`
-    }
-  }
-
-  const container = normalize(file?.container)
-  const codec = normalize(file?.codec)
-  if (
-    container.includes('flac') ||
-    codec.includes('flac') ||
-    container.includes('alac') ||
-    codec.includes('alac') ||
-    container.includes('aiff') ||
-    codec.includes('aiff') ||
-    container.includes('ape') ||
-    codec.includes('ape') ||
-    container.includes('dsd') ||
-    codec.includes('dsd') ||
-    container.includes('wv') ||
-    codec.includes('wv') ||
-    container.includes('wav') ||
-    codec.includes('wav')
-  ) {
-    return 'lossless'
-  }
-
-  if (file?.format) return normalize(file.format)
-
-  return ''
 }
--- a/listenarr.domain/Common/QualityMatcher.cs
+++ b/listenarr.domain/Common/QualityMatcher.cs
@@ -1,0 +1,409 @@
+/*
+ * Listenarr - Audiobook Management System
+ * Copyright (C) 2024-2026 Listenarr Contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+using System.Text.RegularExpressions;
+
+namespace Listenarr.Domain.Common
+{
+    /// <summary>
+    /// Normalised, unit-explicit view of an audio file's quality-relevant facts.
+    /// All three call-sites (status evaluator, automatic search, library controller)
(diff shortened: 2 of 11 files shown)


timk75:
Note for maintainers: I can't self-apply a version label (fork contributor). This is a backward-compatible bug fix, so it should take the **`patch`** label for the `Validate PR version label` check. Happy to adjust if you'd classify it differently.

listenarr.domain/Common/QualityMatcher.cs
```diff
@@ -0,0 +1,412 @@
...
+                groups.Add("AAC");
+                groups.Add("M4B");
+            }
+
+            return groups;
+        }
+
+        private static bool IsLosslessFile(AudioQualityInput file)
+        {
+            foreach (var value in new[] { file.Codec, file.Container, file.Format })
```
therobbiedavis: AudiobookFile records can have no quality metadata when metadata processing is disabled, ffprobe is unavailable, or extraction fails; in those cases Path is the only quality signal. MapCodec already uses the path extension, but IsLosslessFile does not, so book.flac maps to the FLAC codec group while still being treated as lossy and filtered away from the FLAC rung. Please include path extension in the lossless check too, or derive lossless-ness from the mapped codec groups.

therobbiedavis (changes requested):
One more thing I noticed while reviewing the status behavior: the backend now fixes the M4B/AAC cutoff path, but the frontend still has a separate client-side status recomputation in `fe/src/utils/audiobookStatus.ts` that can disagree with it.

For a detailed audiobook payload with `format=m4b`, `codec=aac`, `bitrate=256000`, and a profile cutoff/rung of `AAC 256kbps`, the backend `AudiobookStatusEvaluator` resolves this as `quality-match`, but the frontend utility derives the bare label `256kbps` and then fails to find that in structured rungs like `AAC 256kbps`, so it returns `quality-mismatch`.

Since that file is outside this PR’s diff, this may be follow-up work, but otherwise users can still see “Below Cutoff” in richer library views even though the automatic search/backend cutoff logic is fixed.

listenarr.domain/Common/QualityMatcher.cs
timk75: Good catch — fixed in 3496cd3b. `IsLosslessFile` now derives lossless-ness from the mapped codec groups (`MapCodec`), which already fold in the path extension, instead of only checking Codec/Container/Format. So a `book.flac` with no probe metadata maps to the FLAC group *and* is treated as lossless, landing on the FLAC rung. Added a regression test: `Match_FlacFile_WithOnlyPathExtension_MatchesFlacRung`.

timk75:
Both review points addressed in 3496cd3b:

1. **Backend lossless detection** — replied on the inline thread; `IsLosslessFile` now derives lossless-ness from the mapped codec groups (path-extension aware).

2. **Frontend status recomputation** (`fe/src/utils/audiobookStatus.ts`) — the client-side utility now mirrors the backend `QualityMatcher` (codec-group mapping + bitrate round-down, structured-rung aware) instead of comparing a bare derived label by string equality. An `m4b`/`aac`/256kbps file now resolves to `quality-match` against an `AAC 256kbps` rung, matching the backend evaluator, so richer library views no longer show "Below Cutoff" while the search/cutoff logic considers it satisfied. I kept the port deliberately parallel to the C# matcher (with a comment to that effect) so the two stay in lockstep.

Tests added on both sides; backend (`dotnet test`, 27/27 QualityMatcher) and frontend (`vitest`, 8/8) pass, plus `vue-tsc` and ESLint clean.

@therobbiedavis would you mind taking another look when you get a chance?

timk75:
Heads-up on the force-push: it changed the **commit message only** — my local tooling had appended a stray trailer to the message, so I amended it out and re-pushed. No code changes; the tree is identical to the previous push (`3496cd3b` → `555024c3`). Sorry for the noise on your notifications.

fe/src/utils/audiobookStatus.ts
```diff
@@ -83,26 +83,34 @@ export function computeAudiobookStatus(
...
-    const cutoffPriority = qualityPriority.has(cutoff)
-      ? qualityPriority.get(cutoff)!
-      : Number.POSITIVE_INFINITY
+    const cutoffRung = rungs.find((rung) => rung.label === cutoff)
+    if (!cutoffRung) {
+      // Cutoff label is not a known rung — the backend treats this as "never met".
+      return 'quality-mismatch'
+    }
 
     for (const file of candidateFiles) {
```
therobbiedavis: I was going to ask if we can move the audiobook.quality check before the file loop so the frontend matches the backend behavior, but after stepping back and questioning this a bit, why is the FE handling this business logic. A better shape would be for hte BE to be the source of truth for quality-match/quality-mismatch/no-file. Could we change the FE to use the backend-provided status for quality/no-file states and only apply client-side overrides for transient state like active downloads? This will help us prevent duplicating logic everywhere.

therobbiedavis (changes requested):
Getting much closer! Outside of the inline comment I left on audiobookStatus.ts, there's another single-source-of-truth concern on the backend status path: 

QualityCutoffEvaluator now passes Path into QualityMatcher, but AudiobookStatusEvaluator has its own projection/filtering logic and drops AudiobookFormatSummary.Path. That lets automatic search and library status disagree for path-only files. Can we reuse the shared quality matching/projection logic here, or at least pass Path = file.Path, so status and cutoff evaluation stay aligned?

Sorry for the back and forth. I just had a brain blast for a cleaner method to handle this stuff this review round. Ideally the proper way to do this is we shouldn't be duplicating any logic and making sure the appropriate files have the authority and responsibility over the business logic, then we just import where necessary.

fe/src/utils/audiobookStatus.ts
timk75: Done — went with your cleaner shape. The frontend no longer mirrors the matcher at all.

`computeAudiobookStatus` is now: in-flight download → `downloading`; otherwise trust the server-computed `audiobook.status` (the `/library` list endpoint already runs `AudiobookStatusEvaluator.ComputeStatus` and returns it on every item); fallback `no-file`. Deleted the entire client-side `QualityMatcher` port (~160 lines) and the `qualityProfiles` argument, and updated the two call sites + the spec to match. So the backend is the single source of truth for quality/no-file, and the FE only owns the transient download override.

timk75:
Pushed 86d12804 addressing both points from this round.

**Backend `AudiobookStatusEvaluator` dropping `Path`** — good catch. It now forwards `Path = file.Path` into `AudioQualityInput`, exactly the way `QualityCutoffEvaluator.ToInput` does, so automatic-search and library status agree for path-only files (metadata disabled / ffprobe missing / extraction failed). Added a regression test: `ComputeStatus_ReturnsQualityMatch_ForPathOnlyFile_WhenProbeMetadataMissing` (a metadata-less `book.flac` now resolves to `quality-match`).

**Frontend** — replied inline; the client-side matcher is gone and the FE now trusts the server-computed status.

Both `AudiobookFormatSummary` and `AudiobookFile` are still distinct carriers projecting into the shared `AudioQualityInput` (as the record's own doc-comment intends), so the projection itself stays per-carrier — but both now carry `Path`, and the actual matching/cutoff logic lives only in `QualityMatcher`.

Backend 710 tests green, frontend 364 tests + type-check + eslint green.

therobbiedavis (changes requested):
This is closer, but AudiobookStatusEvaluator can still reject path-only files before Path reaches QualityMatcher. The candidate filter only checks Format/Container, so a metadata-less book.flac with PreferredFormats = ["flac"] gets filtered out and returns quality-mismatch before the matcher can use the extension. The new regression test uses empty PreferredFormats, so it does not cover that path. Can we include the path extension in this preferred-format check too?

timk75:
Fixed in 281f7bee — good catch, you were right the filter rejected path-only files before `Path` reached the matcher.

**Fix:** `AudiobookStatusEvaluator`'s candidate-format pre-filter now falls back to the **path extension** when `Format`/`Container` are empty (new `ExtensionFromPath` helper), mirroring what `QualityMatcher.MapCodec` already does. So a metadata-less `book.flac` with `PreferredFormats = ["flac"]` matches by extension instead of being dropped.

**Tests:** the regression test now uses `PreferredFormats = ["flac"]` (covering the filter path your previous note flagged), **plus** a second-format case — a path-only `book.m4b` with `PreferredFormats = ["m4b"]` — to prove the fallback is format-agnostic, not FLAC-specific. Full suite green (711).

**Proactively audited the rest of the codebase for this same bug class** (path-only / metadata-less files mishandled because code reads `Codec`/`Container`/`Format` but ignores the path extension), so we're not back here next round:
- Both `AudioQualityInput` construction sites set `Path` (`QualityCutoffEvaluator.ToInput`, `AudiobookStatusEvaluator`).
- All file-based cutoff/status/upgrade decisions converge on `QualityMatcher` via `AudioQualityInput`; `EfAudiobookFileRepository.GetFormatSummariesAsync` preserves `Path` into its `(Format, Codec, Container)` grouping, so the evaluator still receives it.
- The remaining `"M4B"`/`"FLAC"` string classifiers (`SearchService`, the `*SearchProvider`s) operate on **indexer release strings**, not on-disk files — not this bug class.
- Frontend trusts server-computed status; no client-side codec/extension derivation remains.
- One non-bug note: `DownloadImportService.DetermineQualityFromMetadata` is a separate pre-#648 string classifier used only for import-dedup; it already has a path-extension fallback and fails open (allows import), so no file is mishandled. Could optionally be routed through `QualityMatcher.IsLabelBetter` later for consistency, but it's not a correctness issue.

Net: no other instances of the path-only bug remain.

therobbiedavis (approved):
LGTM. A couple small updates: The changelog is no longer needed. Please update your PR description to match the template. 

We have some other PR's ahead in the queue to get merged so we may ask you to rebase before we merge.

timk75:
Thanks! Both small updates done:

- **CHANGELOG** entry removed (`42d120a9`).
- **PR description** rewritten to match the template (Summary / Changes / Testing / Notes).

I've left the branch as-is rather than rebasing — just say the word once the PRs ahead are merged and I'll rebase onto `canary` right away.

timk75:
Thanks for the review, @therobbiedavis! This one shows as approved but still mergeable=BLOCKED on `canary` — looks like branch protection is holding it (a required check or maintainer merge). Anything outstanding on my end, or is it good to merge when you get a moment?

therobbiedavis:
> Thanks for the review, @therobbiedavis! This one shows as approved but still mergeable=BLOCKED on `canary` — looks like branch protection is holding it (a required check or maintainer merge). Anything outstanding on my end, or is it good to merge when you get a moment?

hey sorry for the delay @timk75 I was working on a large reorg of the backend to help further stabilize some import regressions. Can you resolve the conflicts and rebase so i can retest this?

timk75:
Rebased onto `canary` and retested, @therobbiedavis — conflicts resolved.

Since the backend reorg had moved the quality-decision code, this is more than a textual rebase, so a quick note on how it was re-applied:

- The container-label classifier (`DetermineFileQuality` emitting `"M4B"`, the root of the re-grab loop) still lived on `canary`, just split across `AudiobookQualityCutoffEvaluator`, `AutomaticSearchQualityEvaluator`, and `AudiobookStatusEvaluator`. All three now route through the new `QualityMatcher` instead.
- `AudiobookQualityCutoffEvaluator` is the shared cutoff evaluator now; `AutomaticSearchQualityEvaluator` delegates to it and uses `QualityMatcher.IsLabelBetter` / `ResolveFileQualityLabel` (its duplicate classifier is gone).
- Path-only handling is preserved: `AudiobookStatusEvaluator.ComputeStatus` forwards the file `Path` so metadata-less files resolve by extension through the same matcher, with the FLAC and M4B regression tests carried over.
- Frontend `computeAudiobookStatus` trusts the server-computed `audiobook.status` rather than re-deriving it client-side.

The branch is squashed to two commits (backend matcher + frontend trust-server-status). Full backend suite green (994 passed / 0 failed); `vue-tsc` type-check and eslint clean on the changed frontend files. Ready for your retest.

PR #649, opened 2026-06-04
fix(metadata): restore author/series catalog cache on SQLite

kevinheneveld:
## Summary

The persisted author/series metadata cache never returned a hit on SQLite, so every author and series page re-fetched the full catalog from Audible on each load. The cache-read query ordered by `Nullable<DateTime>.GetValueOrDefault(...)`, which the SQLite EF Core provider can't translate — the query throws, each caller's best-effort `catch` swallows it as a cache miss, and the cache is effectively dead. This switches the ordering to a translatable `COALESCE` (`??`) and adds regression coverage that runs against the real SQLite provider.

### Fixed
- The four catalog-cache reads in `AudiobookRepository` (`GetCachedAuthorByNameAsync`, `GetCachedAuthorByAsinAsync`, `GetCachedSeriesByNameAsync`, `GetCachedSeriesByAsinAsync`) now order by `entry.LastFetchedAt ?? entry.UpdatedAt` (maps to SQL `COALESCE`) instead of `LastFetchedAt.GetValueOrDefault(UpdatedAt)`, which the SQLite EF provider cannot translate and throws on at execution time. Identical freshest-first ordering, but it no longer throws — so the author/series cache actually returns hits instead of re-fetching from Audible on every page load. An inline comment at each site records why, to prevent the pattern regressing.

## Testing
- New `AudiobookRepository_CatalogCacheReadTests` exercises the four reads against the **real** SQLite provider (`Microsoft.Data.Sqlite` `:memory:`) rather than the EF InMemory provider, which tolerates the untranslatable LINQ and so cannot reproduce the failure. The tests fail on the pre-fix code with `Translation of method 'System.DateTime?.GetValueOrDefault' failed` and pass with the fix.
- Full suite green: `dotnet test` (1015/1015).

## Notes
- No changes to the cache *write* paths or schema — behavior is identical apart from the cache now actually being read.
- Rebased onto current `canary`; the `CHANGELOG.md` entry was removed per review (the project no longer tracks `CHANGELOG.md`).

### DIFF
--- a/listenarr.infrastructure/Persistence/Repositories/AudiobookRepository.cs
+++ b/listenarr.infrastructure/Persistence/Repositories/AudiobookRepository.cs
@@ -223,7 +223,9 @@ namespace Listenarr.Infrastructure.Persistence.Repositories
 
             return await _db.AuthorCacheEntries
                 .AsNoTracking()
-                .OrderByDescending(entry => entry.LastFetchedAt.GetValueOrDefault(entry.UpdatedAt))
+                // COALESCE form — SQLite EF can't translate Nullable.GetValueOrDefault (it throws,
+                // and the caller's best-effort catch then silently disables this cache).
+                .OrderByDescending(entry => entry.LastFetchedAt ?? entry.UpdatedAt)
                 .FirstOrDefaultAsync(entry =>
                     entry.AuthorNameNormalized == normalizedName &&
                     entry.Region == normalizedRegion);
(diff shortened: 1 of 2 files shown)


therobbiedavis (approved):
@kevinheneveld This looks good to me. Please delete the changelog and update the PR description to match the template.

therobbiedavis:
@kevinheneveld Can you rebase this and delete the changelog. I want this one to get merged next.

kevinheneveld:
@therobbiedavis addressed your review — should be ready to merge:

- **Removed the `CHANGELOG.md` change** (and the file is gone on `canary` now anyway).
- **Updated the PR description** to match the template.
- **Rebased onto current `canary`** to clear the merge conflict. `AudiobookRepository.cs` and the new test rebased cleanly against the domain refactor; full suite is green (1015/1015). Condensed the inline comments slightly so the file stays under the 500-line architecture limit.

kevinheneveld:
Heads up: the CI runs on the new push are sitting in **`action_required`** (this repo gates fork-PR workflows behind maintainer approval), so they need an *Approve and run* before they'll go green. The `patch` version label is still on the PR, so `check-version-label` will pass as before. Ready to merge once checks run. 🙏

kevinheneveld:
Pushed a fix for the `unit-tests` failure — it was the **Backend format check** (`dotnet format --verify-no-changes`), not a test failure: the new test file carried `using` directives (`Xunit`, `Listenarr.Domain.Models`, `Listenarr.Infrastructure.Persistence`) that are already covered by `tests/GlobalUsings.cs`, so IDE0005 flagged them. Ran `dotnet format` to drop the redundant usings; format check is now clean (exit 0) and the full suite passes locally (1015/1015). The CI runs will need another *Approve and run* to execute on the new commit.

PR #650, opened 2026-06-04
fix(search): normalize bare-array response from POST /search

kevinheneveld:
## Summary

`searchAudibleByTitleAndAuthor` silently dropped every result when the search endpoint returned a bare array. The unified `POST /search` endpoint returns a flat array of Audible-shaped result objects (`SearchController` does `return Ok(flatMapped)`), but the FE service method typed the response as the wrapped `{ totalResults, results }` envelope and read `.results` off it. When the array shape arrived it had no `.results`, so the method returned zero hits and callers saw "no matches" even though IntelligentSearch had actually found one. The method now normalizes both response shapes.

## Why this matters

Any feature that searches Audible by title + author is affected — the call can find a match server-side and the UI still reports nothing. It's shape-dependent, so it surfaces inconsistently and is easy to misread as "Audible just didn't have it."

## Implementation notes

- **Surface:** `fe/src/services/api.ts` — `searchAudibleByTitleAndAuthor` only.
- **Normalization:**
  - bare array → `{ totalResults: resp.length, results: resp }`
  - wrapped `{ totalResults, results }` envelope → returned as-is
  - `null` / `undefined` → empty envelope (`{ totalResults: 0, results: [] }`)
- No change to the endpoint or to callers; purely makes the client tolerant of both shapes the API can return.

## Test plan

- [x] `npx vitest run src/__tests__/api.advancedSearch.spec.ts` — 5/5 passing (new; covers array, wrapped, and null cases).
- [x] `npm run type-check` — clean.
- [x] Full FE unit suite — 367 passing / 1 skipped, no regressions.


kevinheneveld:

PR #654, opened 2026-06-06
fix(downloads): keep DownloadMonitorService alive when a poll cycle hits a transient SQLite lock

kevinheneveld:
## Summary

`DownloadMonitorService`'s poll loop caught only `OperationCanceledException`, so any other exception thrown during a cycle escaped `ExecuteAsync` and tripped the .NET host's default `BackgroundServiceExceptionBehavior.StopHost` — shutting the whole app down. Under a container restart policy that manifests as a crash-loop. This makes the monitor loop log-and-continue on a failed cycle, matching every sibling background service.

## Why this matters

A transient `Microsoft.Data.Sqlite.SqliteException` ("database is locked") is a normal, recoverable event under concurrent writes — it's raised while persisting a download (e.g. `EfDownloadRepository.UpdateAsync`/`GetByIdAsync`). With the old loop, a single such exception didn't just fail one cycle; it took the entire host down. On an instance running a container restart policy this became a crash-loop, restarting roughly every poll cycle (observed ~18 restarts in 9 hours), and each restart re-ran startup work — so a recoverable lock turned into sustained downtime and wasted CPU.

`DownloadMonitorService` was the lone background service with this gap: `QueueMonitorService`, `MovedDownloadProcessor`, and `AutomaticSearchService` already log-and-continue.

## Implementation notes

- Extracted the poll loop into `internal async Task RunMonitorLoopAsync(CancellationToken)` so the resilience behavior is isolated and unit-testable (the public `ExecuteAsync` keeps its 5s polling interval, which a test can't wait on).
- The loop now distinguishes three cases per cycle:
  - `OperationCanceledException` **with** cancellation requested → break (graceful shutdown, unchanged).
  - `OperationCanceledException` **without** cancellation → a transient timeout within a cycle; logged and the loop continues.
  - Any other exception → logged ("Error in Download Monitor Service") and the loop continues. `OutOfMemoryException`/`StackOverflowException` are deliberately not swallowed.
- Surface area: one application-layer service file (`DownloadMonitorService.cs`) plus a new test. No API, schema, or behavioral change to what the monitor *does* on a successful cycle.
- Backwards compatible: identical behavior on the happy path and on cancellation; the only change is that a failed cycle no longer terminates the host.

## What this PR does NOT include

- The underlying SQLite write contention. The connection string sets no `busy_timeout`, so concurrent writes can still raise transient locks — this PR stops a single transient lock from being *fatal*, but reducing how often it occurs is a separate hardening change.

## Test plan

- [x] `cd tests && dotnet test` — 683 / 683 passing (1 new regression test: `RunMonitorLoopAsync_CycleThrows_KeepsLoopingInsteadOfStoppingHost` drives the extracted loop, forces a cycle to throw, and asserts the loop task is still running rather than completed).
- No frontend changes (backend-only).


kevinheneveld:
Closing this — #686's `WorkerCycleRunner.RunPeriodicAsync` already centralizes this exact log-and-continue behavior (`WorkerExceptionClassifier.IsNonFatal` excludes `OperationCanceledException`/`OutOfMemoryException`/`StackOverflowException`) across all background services, which is a cleaner and broader fix than this single-service patch. Thanks — superseded by #686.

PR #656, opened 2026-06-07
System Storage: show available space per root folder, app data and system disk (#508)

### LINKED ISSUE #508: [bug] Naive implementation for "System: Storage" stat readout
`docker:latest`

Tells me I have 4.06gb left, but that is for a USB DOM that the system is installed on, the configured media folder is not on a USB DOM.

Image

s3ntin3l8:
Fixes #508

## Summary

The System page's Storage card previously reported only the disk hosting the application — in Docker/NAS setups that's a small boot/app device (e.g. a 4 GB USB DOM), while the audiobook library lives on a separate mount that was never shown. This adds per-disk reporting so the System page shows available space for the System disk, the App Data (config) volume, and every configured root folder.

### Added
- `GET /api/v1/system/storage` now returns a `disks[]` array with an independently-measured entry for the **System** disk (container root, e.g. `docker.img` on Unraid), the **App Data** disk (config volume holding the database/logs/cache), and **each configured root folder** (one row per folder, no dedupe). Missing or inaccessible paths yield an `unavailable` entry instead of failing the request.
- System page Storage card renders these as a stacked per-disk list: label + mount point, a full-width usage bar, and **free of total** per row.
- `IDiskSpaceProbe` seam abstracting disk measurement — Windows `GetDiskFreeSpaceEx` (handles drive paths, mounted-folder junctions, and UNC/NAS shares) with a `DriveInfo`/statvfs fallback on Unix — so NAS roots are measured and the logic is unit-testable.
- Defensive clamping of used bytes/percentage for filesystems that report free space exceeding total (reserved blocks, over-provisioning, ZFS/Btrfs, network shares).

### Changed
- Legacy top-level `StorageInfo` fields are retained for compatibility but now describe the **App Data disk (config volume)**; `driveName` carries that path (was `DriveInfo.Name`, e.g. `/`). Per-disk detail is in `disks[]`. **API consumers relying on these fields should note the changed meaning.**
- `ProgressBar`'s size formatter gains a `TB` unit (multi-terabyte mounts previously displayed as thousands of GB).

### Fixed
- System page Storage no longer reports only the application/boot device — it reflects the disk(s) where the library actually lives, resolving the misleading free-space readout in #508.

### Removed
- The Storage card's summary badge — it could only headline a single disk; the per-disk list now shows each disk's free/total concisely.

## Screenshots

image

## Testing

- Backend (`SystemServiceStorageTests`, hermetic via temp dirs): no root folders, config-root measurement and content-root fallback, multiple folders on one filesystem, missing path → `unavailable`, UNC path measured via a fake probe (available + unavailable), and an over-provisioned disk → usage clamped to zero.
- Backend (`DiskSpaceProbeTests`): the real `DriveInfo` branch on an existing temp dir and a missing path. The Windows `GetDiskFreeSpaceEx` call can't run on Linux CI, so that path is review-covered.
- Frontend: `StorageDisksList.spec.ts` (per-disk rows, free-of-total, unique keys, unavailable) and `ProgressBar.spec.ts` (TB / sub-TB formatting).
- Full suites green: **691 backend / 368 frontend**, plus `dotnet format`, `vue-tsc`, ESLint, Prettier.
- Verified end-to-end in Docker and on an Unraid deployment with the library on a separate mount.

## Notes

- This PR needs a version label for the `Validate PR version label` check — I don't have triage rights to add it.
- No `CHANGELOG.md` entry, per the no-changelog-on-upstream-PRs policy.


listenarr.infrastructure/Platform/SystemService.cs
```diff
@@ -90,44 +93,113 @@ public SystemInfo GetSystemInfo()
...
+                if (!Directory.Exists(path))
+                {
+                    return new DiskStorageInfo { Label = label, Path = path, Status = "unavailable" };
+                }
 
+                // DriveInfo on the path itself (not Path.GetPathRoot): on Linux this
+                // stats the filesystem containing the path, which is what makes Docker
+                // volume mounts like /audiobooks report their own free space instead
+                // of the container root's. Windows UNC paths throw -> unavailable.
+                var driveInfo = new DriveInfo(path);
```
therobbiedavis: As stated in the comment this won't work for Windows UNC/NAS roots. Can we add a Windows path branch before falling back to unavailabled. Maybe try GetDiskFreeSpaceEx?

fe/src/components/system/StorageDisksList.vue
```diff
@@ -0,0 +1,91 @@
...
+import type { DiskStorageInfo } from '@/types'
+
+defineProps<{
+  disks: DiskStorageInfo[]
+}>()
+</script>
+
+<template>
+  <div class="storage-disks">
+    <div v-for="disk in disks" :key="disk.path" class="disk-entry">
```
therobbiedavis: There is a low likelihood edge case where a user may configure `/` as root and `/app/config/` as root causing a duplicate key warning. Could we include an index in the key? Something like:
`<div v-for="(disk, index) in disks" :key="`${disk.label}:${disk.path}:${index}`" class="disk-entry">`

fe/src/components/system/StorageDisksList.vue
```diff
@@ -0,0 +1,91 @@
...
+      <div class="disk-header">
+        <span class="disk-label">{{ disk.label }}</span>
+        <span class="disk-path">{{ disk.path }}</span>
+      </div>
+      <ProgressBar
```
therobbiedavis, line 41: This is technically derivable from used/total, but #508 is specifically about available space and knowing how much room is left seems like the more useful value. Could we show disk.freeFormatted per row instead of disk.usedBytes so users can immediately see remaining capacity for each root folder without doing the subtraction themselves?

therobbiedavis (changes requested):
Thanks for the fix here. I left a couple of small review notes around edge cases/UX polish; nothing huge, but worth tightening before merge.

listenarr.infrastructure/Platform/SystemService.cs
s3ntin3l8: Added a Windows branch using `GetDiskFreeSpaceExW` (commit 40aec149). It accepts any directory — drive paths, mounted-folder junctions, and UNC shares alike — so NAS roots now measure instead of reporting `unavailable`, and it reports quota-aware free space. Applied to all Windows paths since it's also more accurate than drive-root stats for mounted folders; Unix keeps `DriveInfo`/statvfs. Heads up: CI runs on Linux, so this branch is covered by review rather than an automated test.

fe/src/components/system/StorageDisksList.vue
s3ntin3l8: Fixed in 44da2738 — adopted the suggested composite key `${disk.label}:${disk.path}:${index}`, so a configured root folder sharing a path with the System/App Data entry no longer warns.

fe/src/components/system/StorageDisksList.vue
s3ntin3l8: Changed in 44da2738 — each row's header now shows "«free» free of «total»" (e.g. "6.18 TB free of 7.28 TB"), and the bar's used/total readout is dropped; the percentage under the bar stays. Available space is now visible per row without subtraction.
image

therobbiedavis (approved):
LGTM, but there is one non-blocking request: Could we pull the Windows path/native-call handling behind a tiny testable seam and add coverage for UNC inputs? I’m not asking because I reproduced a failure: I tested GetDiskFreeSpaceExW against a temporary elevated local SMB share and it succeeded with and without a trailing slash. This is more about keeping the NAS-support behavior from regressing later.

s3ntin3l8:
Done in `1bbdc329`. Pulled the Windows `GetDiskFreeSpaceEx` handling behind a small `IDiskSpaceProbe` seam (interface in `application`, impl in `infrastructure`) so `SystemService` no longer branches on OS. Added tests that drive a UNC path through the measurement pipeline (available + unavailable) plus probe tests for the Unix/`DriveInfo` branch — the native call itself still can't run on Linux CI, so that one line stays review-covered.

Same commit also tightens a few small things: clamp used bytes/percentage so filesystems reporting free > total (over-provisioning, ZFS/Btrfs, network shares) can't go negative; XML-doc the legacy top-level `StorageInfo` fields as describing the App Data disk; drop the now-redundant Storage card summary badge (the per-disk list covers it); and add ProgressBar TB-formatting tests.

therobbiedavis:
@s3ntin3l8 LGTM. Can you update the PR description to match the template? We should be good to merge after that.

s3ntin3l8:
> @s3ntin3l8 LGTM. Can you update the PR description to match the template? We should be good to merge after that.

@therobbiedavis: Done and ready to merge

PR #657, opened 2026-06-07
fix: apply naming patterns exactly and unify all naming flows (#571)

### LINKED ISSUE #571: Implicit patterns added to configured naming patterns

s3ntin3l8:
Fixes #571

## Summary

Naming patterns are now applied exactly as configured — no implicit `{Series}` folder level and no implicit `"Title: Subtitle"` combining ("apply given naming patterns exactly", as suggested in the issue). Previously, `LibraryController.ComputeAudiobookBaseDirectoryFromPattern` injected `{Series}` into folder patterns that omitted it (and the result was persisted as the audiobook's `BasePath`, so imported files really landed in the unconfigured series folder), while `RenameService` and `ManualImportController` rewrote `{Title}` to `"Title: Subtitle"` when the patterns didn't use `{Subtitle}` (rendered as `Title - Subtitle`, the originally reported symptom).

As discussed in the issue, the fix expanded into consolidating the previously divergent naming flows behind a single orchestrator, so the per-flow drift that caused this can't recur.

### Added
- `NamingContext` (domain record): normalized, source-agnostic naming metadata with `From(Audiobook)` / `From(AudioMetadata)` / `From(AudibleBookMetadata)` factories. Source-specific quirks live in the factories; processing lives in the naming service.
- `IFileNamingService.BuildDirectory` / `BuildPath`: the single naming orchestrator — folder/file/legacy/custom-base pattern application, multi-file sequence suffixes, sanitization, and path-length limits in one place.

### Changed
- All naming flows route through the orchestrator: `LibraryController` (`POST /library/preview-path`, bulk root-folder change), `ManualImportController`, `RenameService`, `LibraryAddService`, `DownloadImportService`. Five hand-rolled variable dictionaries and four copies of the pattern-branching logic are gone.
- `LibraryController`'s no-pattern fallback is now `{Author}/{Series}/{Title}`, aligned with the orchestrator's legacy default (was `{Author}/{Title}`). Only reachable when no folder pattern is configured at all; an empty `{Series}` collapses away, so non-series books are unaffected.

### Fixed
- Removed the `{Series}` injection in `ComputeAudiobookBaseDirectoryFromPattern` — e.g. `{Author}/{Title} ({Year}) [{Asin}]` no longer previews/persists `Tom Clancy/A Jack Ryan Novel (chronological order)/Executive Orders (2010) [B004ESTSSO]`.
- Removed the implicit `"Title: Subtitle"` combining in `RenameService` and `ManualImportController`. The explicit `{Subtitle}` token keeps working.
- A deliberately slash-less folder pattern (e.g. `{Title}`) is applied exactly instead of being replaced with `{Author}/{Title}` — matching how the import pipeline already applied such patterns.
- Multi-file downloads whose pattern has no `{DiskNumber}`/`{ChapterNumber}` token no longer resolve every file to the same name (overwriting each other) — `BuildPath` appends the stable `-NN` sequence suffix, matching rename/manual-import behavior.
- Author-narrated imports (Artist == Narrator in file tags) keep the author instead of collapsing to `Unknown Author`.
- UNC (NAS share) destinations survive Windows MAX_PATH enforcement. On Windows, `Path.GetPathRoot(@"\\server\share\dir")` has no trailing separator and `EnsurePathWithinLimits` always returns the rebuilt `root + join(parts)` string, so every UNC path it touched came back as `\\server\sharedir\...` — even under the limit. Dormant while the method had no production callers, but reachable once `BuildPath` (which the import/rename flows now use) started calling it. The MAX_PATH logic now sits behind a testable seam with explicit Windows root parsing, so the previously self-skipping path-length tests (and new UNC cases) run on the ubuntu CI runners too.

### Removed
- Controller-local `SanitizeDirectoryName`, the `ApplyNamingPattern(AudioMetadata)` / `ApplyNamingPattern(AudibleBookMetadata)` overloads, and `FileNamingService`'s `ChooseAuthor` heuristic (the noisy-tag author heuristic remains where it's used, in `DownloadImportService.BuildNamingMetadata`).

## Testing

- New tests cover: pattern-without-`{Series}` (incl. the literal reported pattern), slash-less pattern, `{SeriesNumber}` preserved for series-less books, subtitle not combined in rename + manual import, explicit `{Subtitle}` still renders, multi-file no-token suffix collision fix, 13 orchestrator tests locking in canonical `BuildDirectory`/`BuildPath` behavior, the no-pattern `{Author}/{Series}/{Title}` fallback (controller + add-to-library service level), and Windows MAX_PATH/UNC handling via the new seam (drive, UNC, and root-parsing cases that previously self-skipped on non-Windows). Full backend suite: 735/735 green.
- Verified live on a dev instance and a real-library instance: a pattern without `{Series}` previews `Tom Clancy/Executive Orders (2010) [B004ESTSSO]`; switching to `{Author}/{Series}/{Title}` still produces the series folder.

## Notes

Behavior changes to be aware of:

- Users who relied on the implicit additions must add `{Series}`/`{Subtitle}` to their patterns. The default folder pattern `{Author}/{Series}/{Title}` already contains `{Series}`, so default configurations are unaffected.
- Organize/rename previews will propose moves for books sitting in previously implicitly-created paths (correct, but visible). Nothing moves without executing the rename.
- Already-persisted `BasePath` values are not rewritten retroactively.
- Manual import now creates an `Unknown Author` folder for author-less books: the unified builder includes the `Author` token with the standard `"Unknown Author"` fallback (matching every other flow); manual import previously omitted blank variables, so author-less batches landed directly under the base path.
- Colon (`:`) in titles now renders as ` - ` in the Add-book destination preview / persisted `BasePath` (the consolidated `SanitizePathComponent` replaces the removed controller-local sanitizer, which mapped `:` → `_`).


listenarr.api/Controllers/LibraryController.cs
```diff
@@ -3491,20 +3492,6 @@ private string ComputeAudiobookBaseDirectoryFromPattern(Audiobook audiobook, str
...
-                }
-                else if (directoryPattern.Contains("{Author}/"))
-                {
-                    directoryPattern = directoryPattern.Replace("{Author}/", "{Author}/{Series}/");
-                }
-            }
-
             // If the audiobook has no Series, remove any {Series} tokens from the directory pattern
             // Tests expect the controller to strip the Series token when series metadata is missing.
             if (string.IsNullOrWhiteSpace(audiobook.Series))
```
therobbiedavis: This cleanup still removes more than the {Series} token. The regex \{Series[^}]*\} also matches explicit tokens like {SeriesNumber} or {SeriesNumber:00}, so a configured pattern such as {Author}/{SeriesNumber}/{Title} gets collapsed when the audiobook has no series title. Since this PR is making patterns apply exactly as configured, can we narrow this to only remove {Series} with an optional format, or let ApplyNamingPattern handle the empty {Series} value cleanup?

therobbiedavis (changes requested):
Overall this looks good. I found one edge case in the path cleanup logic though. Once that's buttoned up this should be good to merge. Just FYI we're reviewing multiple PRs at once and haven't put this on our roadmap just yet so you may need to rebase while we review and merge other PRs before yours. Thanks!

listenarr.api/Controllers/LibraryController.cs
s3ntin3l8: Good catch — fixed in 719d9d7c by removing that strip block entirely and letting `ApplyNamingPattern` handle the empty `{Series}` cleanup (it already substitutes empty tokens and collapses the surrounding separators, which is the single cleanup path the rest of this PR relies on). `{SeriesNumber}`/`{SeriesNumber:00}` are now preserved when a book has no series title; the existing non-series test (`{Author}/{Series}/{Title}` → `Author/Title`) still passes, and I added a regression test for `{Author}/{SeriesNumber}/{Title}`.

T4g1 (changes requested):
I believe we need more that that, the original issue arise from the fact that each service kind of handle the pattern naming in its own way, the fix should focus first on reconciliating the pattern handling to a single service or from a centralized place at least

s3ntin3l8:
@T4g1: You're right that the path handling is duplicated. 
The token-substitution step itself is already centralized in `FileNamingService.ApplyNamingPattern`, every flow calls it. What's re-implemented per service is the path-assembly orchestration and variable-building around it (build the variable dict -> folder/file/legacy branch -> multi-file suffix -> extension). That lives independently in `RenameService.BuildExpectedPath`, `ManualImportController`, `DownloadImportService`, and `LibraryController.ComputeAudiobookBaseDirectoryFromPattern` / `LibraryAddService.AddToLibraryAsync`.

I agree consolidating onto one orchestrator is the right move. Should we fold it into this PR (which is just a small bug fix), or adress in a seperate PR?

therobbiedavis (commented):
@s3ntin3l8, I am more prone to expanding the scope slightly to go ahead and do things the proper way than I am separating out things into subtasks/prs.

s3ntin3l8:
@T4g1 @therobbiedavis The consolidation discussed above is implemented and pushed.

**What changed:**
- New `NamingContext` (domain record) with `From(Audiobook / AudioMetadata / AudibleBookMetadata)` factories, and a single orchestrator in `FileNamingService`: `BuildDirectory` (folder-only, for BasePath) and `BuildPath` (folder + file patterns, multi-file suffixes, sanitization, MAX_PATH).
- All five flows now route through it - `RenameService`, `ManualImportController`, `DownloadImportService`, `LibraryController` (preview-path / bulk root-folder), `LibraryAddService`. The per-service variable dicts, branch logic, and the duplicate sanitizer are deleted, so the per-flow drift that caused #571 can't recur.

**Consolidation flushed out two latent bugs, both fixed with regression tests:**
- Multi-file downloads whose pattern has no `{DiskNumber}`/`{ChapterNumber}` token resolved every file to the same name (silent overwrite); `BuildPath` now appends the stable `-NN` suffix like rename/manual import already did.
- Windows MAX_PATH enforcement mangled UNC destinations (`\\server\share\dir` -> `\\server\sharedir\…`) because `Path.GetPathRoot`'s UNC root has no trailing separator. The limit logic now sits behind a small testable seam with explicit Windows root parsing, so the path-length tests (incl. new UNC cases) run on the ubuntu CI runners instead of self-skipping.

Full backend suite is 735/735 green. The PR description is rewritten to the template and lists the (opt-in) behavior changes - happy to adjust anything that doesn't match what you had in mind.

Issue #626, opened 2026-05-28
[Feature Request] Sort Series by the order of the book in the series


scadthings:
When viewing books within a Series, the only available sort is alphabetical by title, by author or by date added.  There is no way to sort by series reading order/position number.

Currently:
Only title, author and date added sort is available, which for many series bears no relation to reading order.

Requested feature:
The ability to sort books within a series by their series position number (e.g. Book 1, Book 2, Book 3) as this is the primary way audiobook listeners navigate a series.

s3ntin3l8:
Implemented in #660 — series collections now have a "Series Position" sort that orders books by their position number, and it's the default for series views.

PR #659, opened 2026-06-07
feat: show audiobooks under all their series and fix the Primary-series toggle (#658)

### LINKED ISSUE #658: Books in multiple series: only shown under their primary series, and the "Primary" toggle doesn't
  persist (also affects {Series} folder naming)
## Summary

The backend fully models books-in-multiple-series (`AudiobookSeriesMembership` join with `IsPrimary`/`SortOrder`, populated by Audible/Audnexus), but the UI and one list query still treat a book as single-series. Two user-visible problems result, plus a knock-on effect on file/folder naming.

## Why this matters (real-world example)

Many series exist in multiple *ordering variants*, and a single book belongs to all of them. Classic case: Tom Clancy's **Jack Ryan** universe is cataloged both in **publication order** and **chronological order** — most books appear in both. With the current single-series behavior:

- **No congruent series view.** Each added book lands in only *one* of the two orderings (whichever the metadata provider marks primary), so the library splits the same universe across two half-empty series — you can never see one complete, correctly ordered list.
- **Add intent is silently ignored.** If you deliberately add a book *from within* a specific series (say you're browsing the chronological-order collection) but the provider marks the other ordering as primary, the book does **not** show up in the series you added it from. It lands in the "wrong" series with no obvious way to fix it (and, per Problem 2, the Primary toggle that should fix it doesn't stick).
- **Files get foldered under the wrong ordering.** Because `{Series}` uses the primary, the on-disk layout (`{Author}/{Series}/{Title}`) follows the provider's pick, not the ordering the user actually wants.

Showing a book under **all** its series (Problem 1) restores one coherent view, and a **persistent Primary toggle** (Problem 2) lets the user choose the canonical ordering — which also drives folder naming. (Optional follow-up: adding a book from within a series view could mark that series as the chosen primary on add.)

## Problem 1 — A book in 2+ series only appears under one series

Library grouping (`fe/src/views/library/AudiobooksView.vue` `groupedCollections`, ~L1371) and the series collection filter (`fe/src/views/library/CollectionView.vue` `matchesCurrentCollection`, L964–966) key off the legacy single `book.series` (the primary), ignoring `seriesMemberships`. Compounding this, the library list endpoint (`LibraryListService.GetAllAsync` → `LibraryAudiobookListItem`) doesn't even include `SeriesMemberships`, so the frontend is blind to them.

## Problem 2 — The "Primary" series toggle doesn't persist

In the Edit Audiobook modal, toggling which series is Primary appears to do nothing — the book reverts to the provider's original primary. Root cause is a frontend payload bug: `fe/src/components/domain/audiobook/EditAudiobookModal.vue:1681` builds `isPrimary: Boolean(membership.isPrimary || index === 0)`, which re-flags index 0 as primary alongside the user's pick (e.g. `[F,T]` → `[T,T]`). The backend `AudiobookSeriesMembershipHelper.Normalize` (`listenarr.domain/Common/AudiobookSeriesMembershipHelper.cs:79`, `FindIndex(m => m.IsPrimary)`) then takes the *first* primary = index 0 = the provider default, discarding the user's choice. (Same buggy pattern at L1004.)

## Related — manual rescan wipes the chosen primary

`LibraryController.ApplyMetadataRescanPatch` (~L3997) → `ApplyToAudiobook` fully replaces memberships with provider data and re-derives primary as the provider's first, so a manual "Rescan Metadata" resets the user's choice again. There's no field-locking.

## Knock-on — folder/file naming follows the (wrong) primary

`{Series}`/`{SeriesNumber}` resolve to `audiobook.Series`/`SeriesNumber` (the primary), and the default folder pattern is `{Author}/{Series}/{Title}`. So while the toggle bug exists, books get foldered under the provider's series rather than the user's chosen one. Fixing Problem 2 fixes naming for free.

## Suggested fixes

1. **Toggle (Problem 2):** drop `|| index === 0` at `EditAudiobookModal.vue:1681` (and L1004) → `isPrimary: Boolean(membership.isPrimary)`. Frontend-only; the backend can't recover intent once two primaries are sent.
2. **Show in all series (Problem 1):** add `SeriesMemberships` to `LibraryAudiobookListItem` and batch-load them in `LibraryListService` (mirroring its existing file-summary batching — not a blanket `.Include` on the shared `GetAllAsync`). Then fan books across all memberships in `AudiobooksView` grouping and `CollectionView` filtering, resolving the correct per-series number within each collection.
3. **Cards/modals:** list all series (fallback to primary).
4. **Rescan (Related):** preserve the user's primary across manual rescan when the provider still returns that series; fall back to the provider default otherwise.
5. **Naming:** no code change — covered by (1)/(4); add a regression test asserting `{Series}` foldering uses the chosen primary.

s3ntin3l8:
## Summary

The backend already models books that belong to multiple series (`AudiobookSeriesMembership` join with `IsPrimary`/`SortOrder`, populated by Audible/Audnexus), but the UI and one list query still treated each book as single-series. As a result a book that's in two or more series only showed under its provider-chosen primary, and the per-book "Primary" series toggle silently reverted on save.

This PR makes a book appear under **every** series it belongs to and makes the **Primary** selection actually stick — which also makes `{Series}` folder naming follow the chosen series.

Motivating example: a series catalogued in both *publication order* and *chronological order* (e.g. Jack Ryan) previously split across two half-empty series with no way to pick the canonical one.

Closes #658.

### Added
- **Books appear under all their series:** the library "Series" grouping and the series collection view list a book under each of its series memberships (not just the primary), showing the correct position number for that series. Book cards and detail modals list every series the book belongs to.
- **Series memberships in the library list payload:** `LibraryAudiobookListItem` now carries `SeriesMemberships`, batch-loaded in `LibraryListService` via a new `GetAllSeriesMembershipsGroupedByAudiobookIdAsync` repository query (mirrors the existing file-summary batching — no per-row `Include`, no change to the shared `GetAllAsync`).

### Fixed
- **"Primary" series toggle now persists:** the edit dialog's save payload built `isPrimary: Boolean(membership.isPrimary || index === 0)`, which re-flagged the first series as primary alongside the user's pick; the backend keeps the first primary it finds, so the choice reverted to the provider default. The payload now sends only the user's selection.
- **Metadata rescan keeps the chosen primary:** a manual "Rescan Metadata" replaced memberships with provider data and re-derived primary as the provider's first. A new `AudiobookSeriesMembershipHelper.ApplyToAudiobookPreservingPrimary` re-applies the user's chosen primary when the provider still returns that series (falling back to the provider default otherwise).
- **Latent DbContext concurrency hardening (independent of the feature):** `LibraryListService` started the file-summary, file-count, and series-membership reads concurrently on the shared scoped `ListenArrDbContext`; under real database latency that risks EF Core's "a second operation was started on this context instance" error (SQLite masks it today by running the async path synchronously). Those same-context reads now run sequentially, while the download read — which uses its own `IDbContextFactory` context — stays parallel.

## Testing
- Backend `dotnet test`: full suite green, including new unit tests for `ApplyToAudiobookPreservingPrimary` and the `Normalize` non-first-primary case, a library-list payload test asserting `seriesMemberships`, and a `RenameService` regression asserting `{Series}` folders under the chosen primary.
- Frontend `vitest`: full suite green, including a new `EditAudiobookModal` test asserting the save payload keeps a non-first primary (index 0 = `isPrimary:false`) and `seriesUtils` unit tests.
- `vue-tsc` type-check, ESLint, Vue template-handler check, Prettier, and `dotnet format` all clean.

## Notes
- No DB migration — the `AudiobookSeriesMemberships` table already exists; this is display + persistence wiring only.
- Related PR #626 ("sort a series by reading order") is **independent** and also based on `canary` — the two compose (this PR gives #626 correct per-series numbers for multi-series books) but neither depends on the other, and they can merge in any order.
- **Overlap with #571 (PR #657):** this PR's new `RenameServiceTests` test is placed at the end of the class specifically so it no longer collides with #571's additions — the two now **auto-merge** in `RenameServiceTests.cs` regardless of merge order.
- Needs a version label.


listenarr.application/Audiobooks/LibraryListService.cs
```diff
@@ -66,7 +66,9 @@ public async Task<IReadOnlyList<LibraryAudiobookListItem>> GetAllAsync()
             var fileSummaryTask = _audiobookFileRepository.GetFormatSummariesAsync();
             var fileCountTask = _audiobookFileRepository.GetCountsByAudiobookIdAsync();
             var activeDownloadTask = _downloadRepository.GetActiveAudiobookIdsAsync(ActiveLibraryDownloadStatuses);
+            var seriesMembershipTask = _audiobookRepository.GetSeriesMembershipsByAudiobookIdsAsync();
```
therobbiedavis: This starts another EF query before the shared-DbContext file queries are awaited. Under real DB latency this can hit EF’s “second operation was started” error. Please await same-context reads sequentially or use independent contexts.

fe/src/views/library/CollectionView.vue
```diff
@@ -985,14 +992,36 @@ function matchesCurrentCollection(book: Audiobook): boolean {
 }
 
 function mapLibraryItem(book: Audiobook): CollectionDisplayItem {
+  // In a series collection a book may be matched via a non-primary membership, so show the
+  // series name/number for THIS collection rather than the book's primary series.
+  const seriesContext = type.value === 'series' ? resolveSeriesForCollection(book) : null
   return {
     ...book,
+    ...(seriesContext
+      ? { series: seriesContext.seriesName, seriesNumber: seriesContext.seriesNumber }
```
therobbiedavis, line 1013: This resolves the collection-specific seriesNumber, but the templates never render it. We should probably show the series+position in series collection list/grid details somehow.

therobbiedavis (changes requested):
Thanks, this is a nicely focused fix and the regression coverage is good. I left two comments: one backend concurrency risk around starting multiple EF queries on the same scoped DbContext, and one UI gap where the series collection view resolves the per-series number but never renders it. Also the changelog is no longer needed.

s3ntin3l8:
Fixed the points from the review, also pushed two follow-ups that I noticed while testing:

- "Not Added" items now show the viewed series' position, not the book's primary-series number. The series catalog was mapping each book via book.Series.FirstOrDefault() (its primary); it now selects the entry for the series being viewed, so a book that's #3 in this series no longer shows its primary's #5. Library items were already correct. (Covered by a new test.)

- Renamed GetSeriesMembershipsByAudiobookIdsAsync → GetAllSeriesMembershipsGroupedByAudiobookIdAsync — it takes no IDs and returns the whole AudiobookSeriesMemberships table grouped by audiobook id, so the name now matches the behavior (no more "expects IDs" footgun). Pure rename, no behavior change.

I also called out the LibraryListService sequential-await change in the description as the latent DbContext "second operation" fix, per the review note.

These are separate commits for easier review — can be squashed before merge.

therobbiedavis (approved):
This one LGTM as well!

PR #660, opened 2026-06-07
feat: sort series collections by reading order (#626)

### LINKED ISSUE #626: [Feature Request] Sort Series by the order of the book in the series

s3ntin3l8:
## Summary

Within a series, the only sorts were title / author / date added — none reflected reading order. This adds a **"Series Position"** sort and makes it the default for series collections, so books are ordered by their position number (the way listeners actually navigate a series).

Ordering is applied **within** the existing "In Library" and "Not Added" groupings (owned books in position order, then not-added in position order). The series view already renders those two sections and paginates the sorted list before splitting it, so grouping-then-position keeps each section whole and in order rather than scattering owned books across pages.

Closes #626.

### Added
- **"Series Position" sort for series collections**, shown only inside a single series and selected by default; other collection types keep their existing default (title).
- Numeric-aware position ordering (handles `1`, `2.5`, `10`; books without a position sort last), applied within the owned / not-added grouping.

## Testing
- Frontend `vitest`: full suite green, including a new `CollectionView` test that owns books #1 and #3 of a 4-book series and asserts the order is `#1, #3` (owned, by position) then `#2, #4` (not added, by position).
- `vue-tsc` type-check, ESLint, Vue template-handler check, and Prettier all clean.

## Notes
- **Independent of #658** — based directly on `canary` and merges in any order. #658 ("show books under all their series") is a related PR that *enhances* this one (it supplies the correct per-series position number for books that belong to multiple series), but this PR doesn't require it: on its own it sorts each series collection by `seriesNumber`. The only overlap between the two is the routine `CHANGELOG.md` `[Unreleased]` conflict (keep both entries).
- Frontend-only; no backend or DB changes.
- Needs a version label.


fe/src/views/library/CollectionView.vue
```diff
@@ -1101,6 +1103,18 @@ function getSortValue(book: CollectionDisplayItem): string {
   }
 }
 
+// Build a lexicographically-comparable key from a series position number ("1", "2.5", "10")
+// so a string sort yields reading order. Non-numeric positions sort by their text; books
+// with no position sort last.
+function seriesPositionSortKey(value: string | null | undefined): string {
+  const raw = (value || '').trim()
+  if (!raw) return '~~~~~~~~'
```
therobbiedavis: This doesn’t actually put missing positions last once the key reaches the comparator below. `seriesPositionSortKey()` returns `~~~~~~~~`, but the sort uses `localeCompare`, and in JS `"~~~~~~~~".localeCompare("00000001")` is negative. So books with no `seriesNumber` appear before `#1`, despite the PR description/comment saying they sort last. Can we handle missing positions explicitly in the comparator or use a safer prefixed key, and add a test with blank/null `seriesNumber`?

fe/src/views/library/CollectionView.vue
```diff
@@ -1101,6 +1103,18 @@ function getSortValue(book: CollectionDisplayItem): string {
...
+  const parsed = Number.parseFloat(raw)
```
therobbiedavis: `Number.parseFloat` accepts partial numeric prefixes, so values like `1a` or `1-2` are treated as numeric `1` and collapse to the same sort key as book `#1`. That doesn’t match the comment that non-numeric positions sort by their text. Could we use a stricter numeric check before parsing, so only fully numeric positions enter the numeric ordering path?

therobbiedavis (changes requested):
Thanks, I left some comments. Once those cases are handled and covered with tests, this looks good to me. Oh and changelog is no longer needed. Just make sure the PR description matches the template (which it looks like it does)

s3ntin3l8:
Pushed the fixes for the two `seriesPositionSortKey` review comments:

- **Missing positions now sort last.** The blank sentinel (`'~~~~~~~~'`) actually sorted *before* digits under `localeCompare`, so un-numbered books jumped ahead of #1. The key is now digit-led, so blanks reliably land last.
- **Strict numeric gate.** `Number.parseFloat` accepted partial-numeric values, so `1a` / `1-2` collapsed onto `#1`. Only fully-numeric positions (`/^\d+(\.\d+)?$/`) take the numeric ordering path now; anything else sorts as text, after the numbers.

Reworked `seriesPositionSortKey` into a 3-tier key (numeric → other text → missing). Added tests for blank/null positions, multi-digit ordering (#2 before #10), and the `1-2`-isn't-#1 case — frontend suite green. Also dropped the CHANGELOG entry per the no-longer-needed policy.

Single commit for review — happy to **squash**/fixup into the feature commit before merge.

therobbiedavis (approved):
LGTM

PR #661, opened 2026-06-08
fix(frontend): load root folders on mount so cold first add isn't falsely rejected

timk75:
## Summary

Fixes the **cold-start "Root folder not configured" false negative**. After a fresh page load (or after the app has been idle long enough that the SPA is re-initialised), the *first* attempt to add an audiobook reports **"Root folder not configured"** and redirects to Settings — even though a root folder *is* configured. Going **browser-back and trying again works**, which is the tell-tale symptom.

**Root cause.** `rootFoldersStore` is only ever populated by an explicit `load()` call. Several views read `rootFoldersStore.folders` *synchronously* to gate UI, but nothing loads the store on a cold SPA start — not `App.vue`, not the router, not these views' `onMounted`. So the guard sees an empty list and assumes no folder is configured. The redirect to `/settings` happens to load the store, which is why the second attempt succeeds.

The same un-loaded-store read affects two more views (found while tracing the root cause).

### Changed
- **`AddNewView.vue`** — load root folders in `onMounted` before `addToLibrary()`'s guard can read them. This is the primary cold-add fix.
- **`AudiobooksView.vue`** — `hasRootFolderConfigured` drives a "Root Folder Not Configured" empty-state CTA; load the store on mount so an empty library doesn't falsely show it.
- **`AudiobookDetailView.vue`** — the path-preview computed falls back to `rootFoldersStore.defaultFolder`; load the store on mount so a cold load previews the correct destination instead of the legacy `outputPath`.
- All three mirror the existing pattern already used in `LibraryImportView.vue` and `UnmatchedFilesModal.vue` (`if (folders.length === 0) await load()`).

### Fixed
- **`stores/rootFolders.ts`** — `load()` no longer resets `folders` to `[]` inside its `catch`. Previously a transient API failure on *reload* wiped already-loaded folders, making a failed load indistinguishable from "no folders configured". `folders` still starts as `[]`, so a first-ever failure naturally yields an empty list.

### Added
- Regression test in `AddNewView.spec.ts` asserting the view fetches root folders on mount.

## Testing

- Full frontend suite green (**363 passed, 1 skipped**), including the new regression test.
- `vue-tsc --noEmit` and `eslint` clean on all changed files.

## Notes

- The guard at `AddNewView.vue` also accepts a configured `applicationSettings.outputPath` as an alternative to a root folder; that path was already loaded on mount, so only the root-folder half was failing.
- `load()` has no in-flight de-dupe; the `if (folders.length === 0)` caller-side guard keeps this to a single fetch per cold load, consistent with the existing call sites. A store-level loading guard would be a reasonable follow-up but is out of scope here.

### DIFF
--- a/fe/src/__tests__/AddNewView.spec.ts
+++ b/fe/src/__tests__/AddNewView.spec.ts
@@ -768,4 +768,21 @@ describe('AddNewView pagination', () => {
     expect(addBtn.text()).toContain('Added')
     expect(addBtn.attributes('disabled')).toBeDefined()
   })
+
+  // Regression: cold-add "root folder not configured" false negative.
+  // The addToLibrary() guard reads rootFoldersStore.folders synchronously, so
+  // the store must be populated on mount; otherwise the first add after a fresh
+  // page load wrongly redirects to /settings.
+  it('loads root folders on mount so the cold first add is not falsely rejected', async () => {
+    const apiModule = await import('@/services/api')
+    const getRootFolders = apiModule.apiService.getRootFolders as unknown as Mock
+    getRootFolders.mockClear()
+    getRootFolders.mockResolvedValue([{ id: 1, name: 'Books', path: '/books', isDefault: true }])
+
+    const router = createTestRouter()
+    mount(AddNewView, { global: { plugins: [createPinia(), router] } })
+    await flushPromises()
+
+    expect(getRootFolders).toHaveBeenCalled()
+  })
 })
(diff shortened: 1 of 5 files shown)


therobbiedavis:
#717 changes the root-folder store and destination/relocation editing flow, so this will need a rebase after #717. The cold-load bug remains distinct: #717 does not add the AddNewView/AudiobooksView/AudiobookDetailView initialization from this PR, and it does not preserve previously loaded folders on a transient load failure. Please keep those behaviors when resolving the rebase.

PR #673, opened 2026-06-10
fix: normalise author group keys to prevent duplicate author cards (#672)

### LINKED ISSUE #672: [Bug] Author overview creates duplicate cards when metadata has inconsistent author name spacing (e.g. "R.R." vs "R. R.")
When multiple audiobooks by the same author are added from sources that return slightly different name formatting — e.g. `"George R.R. Martin"` (no spaces around initials) vs `"George R. R. Martin"` (with spaces) — the Authors overview creates two separate author cards instead of one.

Image

**Environment:** canary v1.11

**Steps to reproduce:**
1. Add several audiobooks by the same author where at least one book has a slightly different author name format in its metadata (different spacing or punctuation around initials)
2. Go to Library → switch to **Authors** view
3. Two author cards appear for the same person with split book counts — one with most books and an author photo, one with the remaining book(s) and no photo

**Expected behaviour:** A single author card showing all books for that author, regardless of minor formatting differences in the underlying metadata.

**Actual behaviour:** Two separate cards are shown with the books split across the name variants. Clicking *either* card correctly shows all books in the detail view — only the overview grouping is affected.

**Root cause:**

`AudiobooksView.vue`'s `groupedCollections` computed uses the exact raw string `book.authors[0]` as the Map key:

```ts
const keys =
  groupBy.value === 'authors'
    ? book.authors?.[0] ? [book.authors[0]] : []
    : getBookSeriesNames(book)
```

Since different metadata sources can return subtly different formatting for the same author, `"George R.R. Martin"` and `"George R. R. Martin"` land in separate groups.

Interestingly, `CollectionView.vue` already solves this with `normalizeCollectionText` (line 934), which collapses all non-alphanumeric characters to single spaces before comparing — that's why clicking either duplicate card shows all books. The fix already exists in the codebase, just not applied to the grouping step.

**Suggested fix:** Apply the same `normalizeCollectionText` normalisation as the Map key in `groupedCollections`, while keeping the raw name as the display value.

**Secondary effect:** The inconsistent name also creates a duplicate `AuthorCacheEntry` in the database, since the backend normalisation treats `"R. R."` and `"R.R."` as different strings. The secondary entry ends up with no image URL because the alternate author ASIN returned by Audible for that name variant has no image in Audnexus.

s3ntin3l8:
## Summary

When metadata returns slightly different formatting for the same author or series name (e.g. `"George R.R. Martin"` vs `"George R. R. Martin"`, or `"Wheel of Time"` vs `"wheel of time"`), the Library overview created duplicate cards instead of merging them. This extracts the existing normalisation helper from `CollectionView.vue` into shared utils and applies it to the grouping step in `AudiobooksView.vue`.

### Added
- `normalizeCollectionText` exported from `fe/src/utils/textUtils.ts` (extracted from the local copy in `CollectionView.vue`)
- Regression tests for the duplicate author-card and series-card failure modes

### Changed
- `CollectionView.vue` now imports `normalizeCollectionText` from `textUtils` instead of defining it locally
- `AudiobooksView.vue` `groupedCollections` normalises the Map key for both author and series grouping, so name variants differing only in case/spacing/punctuation merge into one card; the raw first-seen name is preserved for display

### Fixed
- Author and series overviews no longer create separate cards for the same author/series when the name differs only in punctuation/whitespace/case across books

## Testing

- `cd fe && npm run test:unit` — 380 tests, 0 failures
- `cd fe && npm run type-check` — clean
- Manually verified against a library where `"George R.R. Martin"` and `"George R. R. Martin"` appeared as separate author cards; after the fix a single card shows all books

## Notes

Fixes #672. `normalizeCollectionText` was already used in `CollectionView.vue`'s `matchesCurrentCollection` (for both author and series names), which is why clicking either duplicate card already showed all books — the fix just brings the grouping step in sync.

### DIFF
--- a/fe/src/__tests__/AudiobooksView.spec.ts
+++ b/fe/src/__tests__/AudiobooksView.spec.ts
@@ -795,4 +795,127 @@ describe('AudiobooksView Grouping', () => {
     await wrapper.vm.$nextTick()
     expect(wrapper.find('.series-bottom-placard').exists()).toBe(true)
   })
+
+  it('merges author cards when names differ only in spacing or punctuation around initials', async () => {
+    if (
+      typeof (globalThis as unknown as { ResizeObserver?: unknown }).ResizeObserver === 'undefined'
+    ) {
+      ;(globalThis as unknown as Record<string, unknown>).ResizeObserver = class {
+        observe() {}
+        disconnect() {}
+      }
+    }
+    if (typeof (globalThis as unknown as { WebSocket?: unknown }).WebSocket === 'undefined') {
+      ;(globalThis as unknown as Record<string, unknown>).WebSocket = function () {
+        /* noop */
+      }
+    }
+
+    const pinia = createPinia()
+    setActivePinia(pinia)
+    const router = createRouter({
+      history: createMemoryHistory(),
+      routes: [
+        { path: '/', name: 'home', component: { template: '<div />' } },
+        { path: '/audiobooks', name: 'audiobooks', component: AudiobooksView },
+      ],
+    })
+    await router.push('/audiobooks')
+    await router.isReady().catch(() => {})
+
+    const store = useLibraryStore()
+    // "George R.R. Martin" (no spaces) vs "George R. R. Martin" (spaced) — same person
+    store.audiobooks = [
+      { id: 1, title: 'A Clash of Kings', authors: ['George R. R. Martin'], files: [] },
+      { id: 2, title: 'A Dance with Dragons', authors: ['George R.R. Martin'], files: [] },
+      { id: 3, title: 'A Feast for Crows', authors: ['George R. R. Martin'], files: [] },
+    ] as unknown as import('@/types').Audiobook[]
+
+    store.fetchLibrary = vi.fn(async () => undefined)
+    const wrapper = mount(AudiobooksView, {
+      global: {
+        plugins: [pinia, router],
+        stubs: [
+          'BulkEditModal',
+          'EditAudiobookModal',
+          'CustomFilterModal',
+          'FiltersDropdown',
+          'CustomSelect',
+        ],
+      },
+    })
+    await new Promise((r) => setTimeout(r, 0))
+
+    const vm = getVm(wrapper)
+    await vm.setGroupBy?.('authors')
+    await wrapper.vm.$nextTick()
+
+    const groupedCollections = vm.groupedCollections ?? []
+    // Both name variants must collapse into a single author card
+    expect(groupedCollections).toHaveLength(1)
+    expect(groupedCollections[0].count).toBe(3)
+    // The surviving card keeps the first-seen raw display name (not the normalized key)
+    expect(groupedCollections[0].name).toBe('George R. R. Martin')
+  })
+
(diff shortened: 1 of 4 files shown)


kevinheneveld:
Heads-up on a small overlap: in #680 (unified library search) I independently added a `normalizeCollectionText` to `fe/src/utils/textUtils.ts` that's **byte-for-byte identical** to the one this PR introduces — same NFKD → strip-diacritics → lowercase → collapse-non-alphanumerics helper. So whichever of these two lands second will hit a duplicate-export conflict.

Easy to resolve either direction: happy to drop mine and have #680 consume yours, or vice-versa — whatever's simplest for the maintainers. No change needed on your end unless #673 lands first, in which case I'll rebase #680 onto it. Just flagging so it's visible before either merges.

m4bard:
This fixes the reported case. Before the part that needs changing, one thing worth saying in its favour: normalising and then matching exactly is the conservative choice. It doesn't drop single-letter tokens and it doesn't do subset matching, so `M. R. James` and `Henry James` stay separate. Author normalisation that reaches further than this starts merging people who are not the same person, and nothing in the UI would show that it had happened.

The trouble is the character class. `[^a-z0-9]+` treats every non-Latin script as punctuation, so a name written in Cyrillic or Japanese normalises to the empty string and every such author lands on the same key.

I ran `normalizeCollectionText` over a corpus of 60 distinct public-domain authors. It produces 58 keys, and one of them is `""`, holding three different people:

```
key ""  <-  Лев Толстой | Федор Достоевский | 芥川 龍之介
```

Those three render as one author card. Series names go through the same function, since `rawKeys` feeds both branches into it, so a series called `Война и мир` also keys to `""`, and a library with two non-Latin series would merge them.

Making the class Unicode-aware fixes that without changing the approach:

```js
export function normalizeCollectionText(value: string | undefined | null): string {
  if (!value) return ''
  return value
    .normalize('NFKD')
    .replace(/\p{M}+/gu, '')
    .toLowerCase()
    .replace(/[^\p{L}\p{N}]+/gu, ' ')
    .trim()
}
```

`\p{L}` and `\p{N}` keep letters and numbers from any script. `\p{M}` covers combining marks from any script, whereas `\u0300-\u036f` is one Unicode block. Same corpus against that version: 60 keys for 60 authors, nothing merged. `George R.R. Martin` and `George R. R. Martin` still collapse together, and `Heinrich Böll` still folds to match `Heinrich Boll`.

Smaller thing while you're in there. The group key is normalised now, but `authorCoverOverrides[rawKey]` still looks up the raw name, so a card built from whichever variant was seen first will miss a cover override stored against the other spelling.

Last thing, and #672 was scoped to spacing and punctuation so this isn't a gap in the fix: the variants that still split are the ones where a particle or a full given name differs. `Johann Wolfgang von Goethe` against `Johann Wolfgang Goethe`, `J.M. Barrie` against `James M. Barrie`, `L. M. Montgomery` against `Lucy Maud Montgomery`. All real pairs from the same corpus if they're useful as test cases later.

That corpus is public-domain audiobook metadata I keep for reproducing Listenarr issues. Happy to hand over the author list if you want the non-Latin names as fixtures.

PR #704, opened 2026-06-27
feat: built-in Audiobookshelf-style player

shuff57:
Adds an in-app audiobook player so Listenarr can search, download, **and listen** in one app (no external player needed).

## Backend
- `ListenController`: range-enabled streaming endpoint (`206 Partial Content`, seeking) + playback-state endpoints; path-traversal guard via `FileUtils.IsPathSameOrInside` against configured root folders.
- `PlaybackService`: natural-sorted file ordering, resume position, finished flag, optional auto-unmonitor on finish.
- Chapters: lazy `ffprobe -show_chapters` extraction cached per file; aggregated into playback state.
- Bookmarks: entity + CRUD endpoints. Continue-Listening endpoint.
- EF migrations for the new playback/chapter/bookmark columns + tables.

## Frontend (Vue 3 + Pinia)
- Centered floating mini-player: play/resume, variable speed, skip ±, prev/next file & chapter, scrub, volume, OS/lock-screen media controls (`mediaSession`), keyboard shortcuts, sleep timer, bookmarks, mark-finished, minimize-to-pill + close.
- Library "Books" view split into collapsible **Continue Listening / Finished / Library** sections.
- WCAG-AA contrast pass using existing theme tokens.

## Tests
461 frontend unit tests, ~21 backend unit tests. Verified live against a real ~10.7h GraphicAudio m4b (stream/seek/resume, chapter extraction, progress persistence).


therobbiedavis:
@shuff57 Hey, thanks for contributing! I am very torn on this. I love the idea, it's neat and different than the other *arrs, however I just don't know if this is a feature that people would use and it does expand the scope of Listenarr from just a downloader/organizational tool. 

If we choose to implement this, I could see the player allowing for external integrations, like ABS, Hardcover, Plex(?), where the progress is synced, and that is also exciting.

shuff57:
Hi! I’m open to adjusting what I have. I was looking for an all-in-one audiobook player that didn’t involve searching/downloading on one then opening another app.

For the integration, are you wanting to bring ABS etc into listenarr or are you wanting to build something that uses Listenarr as a search/dowload in other apps? Like jellyfin-enhanced with jellyfin and jellyseer?

Let me know?



________________________________

therobbiedavis:
I understand the desire for an AIO app, my concern stems around fracturing the focus of the app when it comes to maintenance. I will be adding an audio player flow on top of all the other flow. I'm not against that this juncture I'm just still weighing it.

As far as the integration, it would be sending api calls out to ABS/HC/etc to sync the progress at certain events. Events would be like start/stop/pause and then a polling event with a configurable time in seconds to send the same progress sync.

Of course this also means setting up connections to each of these external applications. All of that would need to be a separate pr though. Too many considerations for it to be considered in this pr.

shuff57:
Totally understand. I’ll patiently await your decision.

I’m just trying to give back to the community where I can.

Cheers!

________________________________

shuff57:
Closing this — the built-in player is being reworked as a runtime-installable plugin instead of a core feature. A generic plugin-system PR will follow, with the player available as an installable plugin rather than baked into core.

PR #708, opened 2026-06-27
fix(downloads): fall back to default quality profile in auto-search

shuff57:
## Problem
On the **Wanted** tab, clicking **auto search** returned nothing for any audiobook that had no quality profile assigned, while **manual search** worked. `DownloadService.SearchAndDownloadAsync` bails out *before searching any indexer* when `audiobook.QualityProfile == null`:

```
Audiobook '<title>' has no quality profile assigned
```

Manual search (the `ManualSearchModal`) has no such requirement, so it worked — making the failure look inconsistent and silent (the Wanted view only flashes the message for ~5s).

## Fix
Fall back to the **default** quality profile when an audiobook has none, instead of bailing. Only fail when *no* profile exists at all. Quality scoring then proceeds against the resolved profile.

## Behavior
- Audiobook with a profile → unchanged.
- Audiobook with **no** profile → uses the default profile (logged at info level) and searches/grabs normally.
- No profiles configured at all → clear failure (`No quality profile is configured`).

Small, self-contained change in `SearchAndDownloadAsync`.

### DIFF
--- a/listenarr.application/Downloads/Submission/DownloadService.cs
+++ b/listenarr.application/Downloads/Submission/DownloadService.cs
@@ -131,8 +131,12 @@ namespace Listenarr.Application.Downloads.Submission
             }
 
             // Fall back to the default profile when an audiobook has none, so auto-search
-            // doesn't silently bail. Only give up if no profile exists at all.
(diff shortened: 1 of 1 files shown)


shuff57:
Closing — no longer pursuing this as an upstream contribution.

PR #711, opened 2026-06-29
fix(scan): clear stale legacy file columns when a book's folder is gone

kevinheneveld:
## Problem

Audiobooks can show **"Downloaded" status with zero files** and silently stop being re-searched.

`LibraryListService` derives file presence from a backward-compat fallback:

```csharp
var hasTrackedFiles      = files != null && files.Count > 0;          // real AudiobookFiles rows
var hasLegacyFileSummary = !string.IsNullOrWhiteSpace(a.FilePath);   // legacy single-file column
var hasAnyFile = hasTrackedFiles || hasLegacyFileSummary;
```

So a book with a stale legacy `FilePath` (and no tracked files) reads as `hasAnyFile == true` → "Downloaded", while the UI's FileCount (from the real `AudiobookFiles` rows) is 0. It's also treated as complete by `AudiobookWantedEvaluator`, so it never re-searches.

## Root cause

`ScanJobProcessor.ProcessJobAsync` has a dedicated branch for when a book's `BasePath` directory has vanished: it deletes the tracked `AudiobookFile` rows and sets `audiobook.BasePath = null`, then returns — but it **never clears the legacy `FilePath`/`FileSize` columns**. The later "legacy filePath migration" block that *would* clear them is downstream of this early `return`, so it's skipped.

Result: a book whose whole folder disappeared keeps its legacy columns set → permanent phantom "Downloaded".

## Fix

Clear `FilePath`/`FileSize` alongside `BasePath` in the missing-folder branch:

```csharp
audiobook.BasePath = null;
audiobook.FilePath = null;
audiobook.FileSize = null;
await audiobookRepository.UpdateAsync(audiobook);
```

The synchronous/manual scan path (`LibraryManualScanWorkflow.MigrateLegacyFilePathAsync`) already clears these for a missing file; this brings the background worker in line.

## Test

Added `ProcessJobAsync_BasePathMissing_ClearsLegacyFileColumns`: a book with a non-existent `BasePath` and populated legacy columns is scanned, and `FilePath`/`FileSize` are asserted null afterward (read through a fresh DI scope to bypass the test context's identity map). Full suite green (1016 tests).

### DIFF
--- a/listenarr.infrastructure/Library/Scanning/ScanJobProcessor.cs
+++ b/listenarr.infrastructure/Library/Scanning/ScanJobProcessor.cs
@@ -147,6 +147,12 @@ namespace Listenarr.Infrastructure.Library.Scanning
                         }
 
                         audiobook.BasePath = null;
+                        // The whole folder is gone, so the legacy single-file columns are stale too.
+                        // Leaving them set makes hasAnyFile/Wanted treat the book as "Downloaded"
+                        // with zero files forever (it never re-searches). Clear them alongside
+                        // BasePath so the book correctly reverts to "no file" / wanted.
+                        audiobook.FilePath = null;
+                        audiobook.FileSize = null;
                         await audiobookRepository.UpdateAsync(audiobook);
 
                         if (removedFilesDto.Count > 0)
--- a/tests/Features/Infrastructure/Library/Scanning/ScanJobProcessorTests.cs
+++ b/tests/Features/Infrastructure/Library/Scanning/ScanJobProcessorTests.cs
@@ -124,6 +124,44 @@ namespace Listenarr.Tests.Features.Infrastructure.Library.Scanning
             Assert.Single(files);
         }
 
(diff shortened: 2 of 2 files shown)


therobbiedavis:
#717 rewrites scan authorization/reconciliation and file ownership, so this needs a rebase rather than a direct merge. The bug here remains distinct: in #717, a missing current BasePath is rejected by ScanJobProcessor before AudiobookScanService reconciliation can clear a verified-missing legacy FilePath/FileSize. Please adapt the fix to the new authorization/ownership path and preserve the fail-closed scan semantics.

kevinheneveld:
Good to hear the fail-closed scan semantics survive in #717's authorization path. I'll re-cut this as an adaptation on top of #717: same regression — clearing a verified-missing legacy FilePath/FileSize — expressed against the new `AudiobookScanService` reconciliation, keeping the missing-BasePath rejection fail-closed.

PR #719, opened 2026-07-02
feat(library): move files to another audiobook record

kevinheneveld:
## What

Adds a "Move Files to Another Book" action to the audiobook detail page: move audio files from one library record to another existing record, for when files actually belong to a different book the library already tracks — two books' tracks imported onto one record, or a collection that needs splitting into its real books.

- **`POST /library/{id}/files/transfer`** (new `LibraryTransferFilesWorkflow`): DB ownership reassigns always; the physical file moves into the destination folder best-effort — a failed disk move leaves the file in place with a warning (Organize can relocate it later) and never aborts the transfer.
- **Collision safety**: if the target already owns a row at the path a file would occupy (duplicate from a prior partial transfer, or a dual-referenced file), the file is skipped with a clear warning *before* any disk move — moving first would orphan the source row. Any other per-file DB failure also degrades to a warning instead of a 500.
- **Bookkeeping**: a source left without audio gets its legacy single-file columns (`FilePath`/`FileSize`) reset, gated on what actually reassigned; history entries land on both records ("Files Transferred" / "Files Received").
- **`IAudiobookFileRepository.ReassignAsync`**: targeted column update via a detached stub (only `AudiobookId`/`Path` marked modified, nothing tracked) — a full-entity `Update` on rows sharing navigation graphs trips EF's identity map on the second file of a bulk transfer. Implemented without `ExecuteUpdate` so the InMemory test provider exercises the same code path production runs.
- **UI**: transfer modal with per-file checkboxes and a library search ranked by token overlap (so a filename-ish query still surfaces the right record), a merge warning + confirm when the destination already has files, and a detail-page action next to Organize.
- **Toasts**: warning/error toasts are now sticky by default (dismissed via the close button) — transfer warnings carry information the user needs to read and act on. Info/success keep the 5s auto-dismiss; callers can still pass an explicit timeout to opt back in.

## Why

Live case: a dramatized adaptation's parts were sitting on the wrong record while the book's own entry sat wanted with zero files. Today the only remedies are delete-and-rescan or manual filesystem surgery; this makes the fix a two-click operation that keeps history and never loses a file.

## Tests

9 new tests in `LibraryController_TransferFilesTests` (validation, full + subset transfer, legacy-column gating, missing-on-disk, DB-row collision skip, on-disk collision reassign-in-place, history). Full suite: **1024/1024 passing**. `vue-tsc` + eslint + prettier clean.

### DIFF
--- a/fe/src/components/domain/audiobook/TransferFilesModal.vue
+++ b/fe/src/components/domain/audiobook/TransferFilesModal.vue
@@ -1,0 +1,355 @@
+<!--
+  Listenarr - Audiobook Management System
+  Copyright (C) 2024-2026 Listenarr Contributors
+
+  This program is free software: you can redistribute it and/or modify
+  it under the terms of the GNU Affero General Public License as published
+  by the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  GNU Affero General Public License for more details.
+
+  You should have received a copy of the GNU Affero General Public License
+  along with this program. If not, see <https://www.gnu.org/licenses/>.
+-->
+<!--
+  Move audio files from this record to ANOTHER EXISTING library record — the
+  remediation when files actually belong to a different book the library
+  already tracks (often still "wanted"): two books' tracks imported onto one
+  record, or a collection being split into its real books.
+-->
+<template>
+  <Modal :visible="visible" size="lg" title="Move files to another audiobook" @close="onClose">
+    <ModalBody>
+      <div v-if="!audiobook" class="transfer-empty">No audiobook selected.</div>
+      <template v-else>
+        <div class="transfer-section">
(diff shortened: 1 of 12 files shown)


therobbiedavis:
#717 introduces database-enforced AudiobookFile ownership/identity, a global filesystem mutation coordinator, and ordered per-audiobook operation locks. This transfer feature remains distinct, but its current detached ReassignAsync + best-effort physical move path should be reworked before merge so reassignment cannot bypass ClaimAsync/ownership uniqueness or race concurrent scan/import/move work. Please rebase this stack root on #717 first; #720/#722/#729 should then build on the updated transfer contract.

kevinheneveld:
Makes sense, and agreed on the ordering. Plan: once #717 merges I'll rebase this first — rework `ReassignAsync` to go through `ClaimAsync`/ownership uniqueness and take the per-audiobook operation lock around the physical move, so a reassignment can't race concurrent scan/import/move work. #720/#722/#729 then rebuild on the updated transfer contract in stack order.

One field note from running this live for a month: transfers into a record with no folder of its own need a canonical destination computed *before* any file moves — I got bitten by files landing loose at the output root. I'll make sure the reworked path handles that case under the new mutation coordinator.

PR #720, opened 2026-07-02
feat(files): multi-select files with bulk move/delete on the audiobook detail page

kevinheneveld:
> **Stacked on #719** (transfer-files) — this branch contains that PR's commit because bulk move reuses its `TransferFilesModal`. Only the top commit (`feat(files): multi-select…`) is new here; the diff will collapse to just it once #719 merges.

## What

Adds multi-select to the Files list on the audiobook detail page: a selection checkbox per file row (shift-click extends a range) and a toolbar with select-all, **Move selected…** and **Delete selected**.

- **Bulk move** hands the selection to `TransferFilesModal` (from #719) via its `initialFileIds` preselection — pick a destination record and only the ticked files transfer.
- **Bulk delete** loops a per-file delete behind a single confirm listing the files, with a shared "also delete from disk" option (unchecked: DB rows only; a rescan re-imports them) and a summary toast that reports partial failures.
- **New endpoint** `DELETE /library/{id}/files/{fileId}?deleteFromDisk=` (`LibraryFileDeleteWorkflow` + `IAudiobookFileService.DeleteAudiobookFileAsync`): ownership-guarded (the file must belong to the addressed audiobook), disk-delete failures surface as warnings while the DB row is still removed, and a "File Removed" history entry records each deletion.
- Selection clears on reload/navigation so a stale selection can't act on the wrong rows.

## Why

Deleting duplicates one at a time doesn't scale — the motivating live case was a duplicated book: the same audio under two different filename schemes sitting on one record, which a split-by-content flow can't separate (it's the same book twice). Select the redundant copies → delete once.

## Tests

6 new tests in `LibraryController_FileDeleteTests` (missing audiobook/file, wrong-owner guard, DB-only delete keeps the file on disk, disk delete, row-only ghost cleanup). Full suite: **1030/1030 passing**. `vue-tsc` + eslint + prettier clean.

### DIFF
--- a/fe/src/components/domain/audiobook/TransferFilesModal.vue
+++ b/fe/src/components/domain/audiobook/TransferFilesModal.vue
@@ -1,0 +1,355 @@
+<!--
+  Listenarr - Audiobook Management System
+  Copyright (C) 2024-2026 Listenarr Contributors
+
+  This program is free software: you can redistribute it and/or modify
+  it under the terms of the GNU Affero General Public License as published
+  by the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  GNU Affero General Public License for more details.
+
+  You should have received a copy of the GNU Affero General Public License
+  along with this program. If not, see <https://www.gnu.org/licenses/>.
+-->
+<!--
+  Move audio files from this record to ANOTHER EXISTING library record — the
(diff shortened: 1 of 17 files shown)


therobbiedavis:
#717 changes file ownership and destructive filesystem mutation contracts underneath this stacked PR. Please rebase after #719 is adapted to #717, and route per-file delete/move through the new ownership-safe mutation/delete primitives rather than deleting or reassigning by path/row alone. The multi-select UI itself remains distinct.

kevinheneveld:
Will do — after #719 is adapted, per-file delete/move here get routed through the new ownership-safe primitives instead of operating by path/row. The multi-select UI layer is untouched by that, so this should shrink to a small rebase once the root settles.

PR #722, opened 2026-07-02
feat(library): "Split collection" — break a multi-book record into its real books

kevinheneveld:
> **Stacked on #720** (which stacks on #719) — split's apply phase is a sequence of those PRs' file-transfer and per-file-delete calls. Only the top commit (`feat(library): "Split collection"…`) is new here; the diff collapses to just it as the base PRs merge.

## What

A **Split Collection** action on the audiobook detail page for multi-book dumps imported onto one record (the motivating live case: 773 files spanning 34 books on a single record).

- **`GET /library/{id}/split/preview`** (`LibrarySplitPreviewWorkflow`, read-only) clusters the record's files into per-book groups:
  1. **Subdirectory** — the strongest signal, wins even over embedded tags (trusting a shared mis-tag across folders merged genuinely different books).
  2. **Embedded Album/Title tag** for flat files — a collection bulk-renamed to the parent record's name has useless filenames, but each file's tag still names the real book.
  3. **Filename stem** with track/list numbering stripped (`friday-01_77.mp3 … friday-44_77.mp3` → one `friday` group). The numbering *style* stays part of a group's identity so two copies of one book (`Title-NN` vs `Title (N)`) yield two groups; a trailing volume designator's own number is preserved so `Vol. 1`/`Vol. 2` don't collapse; an unnumbered file joins its numbered siblings only when duration/size says it plausibly is one track.
- Each group gets a **suggested destination**: same-author records first, longest normalized-title containment wins (`SplitDestinationSuggester` + a new punctuation-tolerant `TitleMatcher.Normalize`).
- **Applying is client-driven**: per group, move (existing transfer endpoint), delete (existing per-file delete, behind a danger confirm), or leave — with per-group progress and a summary toast.
- **Determinism**: only the raw ffprobe runs in parallel (bounded ×4); higher-level metadata reads share a request-scoped, non-thread-safe DbContext, and probing those concurrently made the same file read its tag on one run and blank on the next — i.e. non-deterministic clusters. ffprobe unavailability degrades gracefully to path/stem clustering.

## Tests

`FileClustering` is pure and deterministic; its unit tests use real filename shapes from the live collection that motivated the feature (subdirectory-wins, stacked markers, volume preservation, duplicate-copy separation, unnumbered-first-track merge, embedded-tag clustering). Full suite: **1056/1056 passing**. `vue-tsc` + eslint + prettier clean.

### DIFF
--- a/fe/src/components/domain/audiobook/SplitCollectionModal.vue
+++ b/fe/src/components/domain/audiobook/SplitCollectionModal.vue
@@ -1,0 +1,497 @@
+<!--
+  Listenarr - Audiobook Management System
+  Copyright (C) 2024-2026 Listenarr Contributors
+
+  This program is free software: you can redistribute it and/or modify
+  it under the terms of the GNU Affero General Public License as published
+  by the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  GNU Affero General Public License for more details.
+
+  You should have received a copy of the GNU Affero General Public License
+  along with this program. If not, see <https://www.gnu.org/licenses/>.
+-->
+<!--
+  Break a multi-book record apart: server-side clustering (subdirectory /
+  embedded tag / filename stem) with a suggested existing library record per
+  group. Each group can be MOVED to a record (files relocate to its folder),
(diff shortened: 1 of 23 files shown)


therobbiedavis:
#717 does not supersede Split Collection, but this is stacked on #720/#719 and its apply phase inherits their file-transfer/delete contracts. Please rebase after the lower stack is adapted to #717s database-enforced file ownership and operation locking. The clustering/preview behavior remains independently useful.

kevinheneveld:
Understood — clustering/preview stay as they are; the apply phase inherits whatever #719/#720 look like post-#717. Will follow the stack order.

PR #726, opened 2026-07-02
feat(organize): library-wide organize sweep with metadata-stub target replacement

kevinheneveld:
## What

A library-wide **Organize** flow (Settings → General): preview every book whose on-disk location disagrees with the configured naming pattern, then apply the moves as queued jobs.

- **Preview** (`LibraryOrganizeSweepWorkflow`): computes each book's expected path via the same `LibraryPathPlanner` that add/move use (so organize can never disagree with them), and groups results — `will_move`, `already_organized`, and per-reason invalid buckets (missing base path, unresolvable pattern, flatten-infeasible, target occupied…).
- **Apply**: enqueues standard move jobs (existing `IMoveQueueService` machinery — nothing new runs in-request), with progress via the existing MoveJob update channel. Returns 503 when the worker isn't available rather than silently dropping.
- **Metadata-stub replacement**: a `will_move` whose target folder exists but contains only metadata (cover/nfo — no audio at any depth, nothing in the DB referencing it) is flagged `ReplaceStubTarget` (new MoveJobs column, hand-written migration) instead of `invalid`. At execution time `MoveJobProcessor` re-verifies the no-audio invariant on disk and deletes the stub so the move can replace it; any audio found since enqueue → the job fails instead. Populated non-stub targets keep failing exactly as before.
- **UI**: `OrganizeLibraryModal` (grouped preview, per-group selection, apply progress) + a pending-moves banner, in a new maintenance section on the General settings tab.

## Notes for review

- Filesystem probing lives behind `IOrganizeFilesystem` (infrastructure); the workflow stays IO-free per the architecture tests.
- `MoveJobProcessor`'s target gate moved to a `TargetGate` partial (the main file was at the 500-line cap).
- Migration `AddReplaceStubTargetToMoveJobs` is hand-written (INTEGER, default false) with Designer + snapshot in agreement; boots and migrates clean in the e2e host.

## Tests

31 new: organize preview/apply workflow + controller tests (ported from the original suite), processor stub-gate tests, and pure filesystem-classifier tests. Full suite **1046/1046**; `vue-tsc` (pinned 3.3.4) + eslint + prettier + vitest clean.

### DIFF
--- a/fe/src/components/domain/organize/MoveQueueStatusBanner.vue
+++ b/fe/src/components/domain/organize/MoveQueueStatusBanner.vue
@@ -1,0 +1,414 @@
+<!--
+  Listenarr - Audiobook Management System
+  Copyright (C) 2024-2026 Listenarr Contributors
+
+  This program is free software: you can redistribute it and/or modify
+  it under the terms of the GNU Affero General Public License as published
+  by the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  GNU Affero General Public License for more details.
+
+  You should have received a copy of the GNU Affero General Public License
+  along with this program. If not, see <https://www.gnu.org/licenses/>.
+-->
+<!--
+  Polls GET /library/move/summary on a 5s interval while mounted (paused
+  automatically when the browser tab is hidden — Page Visibility API).
+  Displays counts + the most recent completed/failed entries so the user
+  can see at-a-glance whether the move queue is draining, stuck, or
+  failing.
+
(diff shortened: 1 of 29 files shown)


therobbiedavis:
#717 replaces the underlying move-job model with a durable manifest/lease/recovery/ownership workflow, so this organize sweep remains useful but needs a substantive rebase. Please enqueue through the new move contract and integrate metadata-stub target replacement with the new target-ownership/pinned-mutation rules. Also, the AddReplaceStubTargetToMoveJobs migration described here is hand-written; repository rules after #717 require future migrations to be scaffolded with dotnet ef migrations add, so that migration must be regenerated with EF tooling.

kevinheneveld:
Fair on both counts. Post-#717 plan: enqueue through the new move contract, integrate the metadata-stub target replacement with the target-ownership/pinned-mutation rules, and regenerate `AddReplaceStubTargetToMoveJobs` with `dotnet ef migrations add` per the new tooling rule. This one will trail the #719 chain since it shares the move surface.

PR #727, opened 2026-07-02
fix(persistence): two migrations were never discovered by EF — schema drifts from model on every install

kevinheneveld:
## The bug

`20251124102000_AddMoveJobSourcePath.cs` shipped **without a `[Migration]` attribute or Designer file**, so EF's migration discovery has never seen it. Every SQLite install is missing `MoveJobs.SourcePath` while the model maps it — verifiable on any instance:

```
sqlite3 config/database/listenarr.db "PRAGMA table_info(MoveJobs);"   # no SourcePath
sqlite3 config/database/listenarr.db "SELECT MigrationId FROM __EFMigrationsHistory WHERE MigrationId LIKE '%SourcePath%';"   # empty
```

Any full-entity MoveJobs query or insert dies with `SQLite Error 1: 'no such column: m.SourcePath'` — I hit it live on a new endpoint, but move-job persistence itself is affected.

## The fix

- **Restore the Designer** for `AddMoveJobSourcePath` (attributes + model as of that migration). EF applies not-yet-recorded migrations on next startup regardless of later ones already being in history, so existing installs heal themselves on upgrade.
- The audit found a **second identical drift**: `ProcessExecutionLog` has been in the model and every snapshot since late 2025, but no discovered migration ever created its table (`SELECT name FROM sqlite_master WHERE name='ProcessExecutionLogs'` is empty on real installs). New migration `AddProcessExecutionLogs` creates it.
- **Regression net**: `SqliteMigrationSchemaTests` migrates a real (shared in-memory) SQLite database and asserts every mapped column of every entity exists in the migrated schema. Hand-written migrations + an InMemory-provider test suite is exactly the combination that let this ship silently; this test makes the whole class of drift fail CI. It found the second bug above on its first run.

## Heads-up

There are **16 more migration files without a `[Migration]` attribute/Designer** (mostly `AddXxxToApplicationSettings` and pre-org-migration files) — equally invisible to EF. The schema test passing means the *current model* doesn't depend on any of them (their changes were superseded or re-rooted), but they're dead code worth cleaning up separately. List available on request.

## Tests

Full suite passing (1017/1017), including the 2 new schema tests.


therobbiedavis:
Closing as superseded by #717. #717 incorporates/adapts the missing AddMoveJobSourcePath designer, the ProcessExecutionLogs migration, and real-SQLite migration/schema drift regression coverage, then extends the migration contract further. The relationship and credit are preserved in #717s description. Thanks for identifying the original EF discovery gap.

kevinheneveld:
Sounds right — glad the discovery gap and the SQLite drift regression coverage found a home in #717, and thanks for preserving the attribution. I'll drop my branch once #717 lands.

PR #729, opened 2026-07-03
feat(library): duplicate record + duplicate copy detection

kevinheneveld:
> **Stacked on #722** (split-collection) — reuses its `TitleMatcher.Normalize` and `FileClustering`. Diff collapses when the stack merges.

## What

`GET /library/duplicates` (read-only sweep, no ffprobe) plus a Settings → General section:

- **Duplicate records**: grouped by identical ASIN, else normalized title+author. Deliberately conservative — groups with ≥2 distinct ASINs (different editions) or ≥2 distinct subtitles AND years are dropped; false positives would erode trust. Each group nominates a suggested keeper (most files, oldest id) and offers record-only delete (files stay on disk) behind a danger confirm.
- **Duplicate copies within a record**: `FileClustering` signatures — same filename stem with a different numbering style, or two multi-file clusters with totals within 25%, flags a record as likely holding the same book twice (resolved on the book page via multi-select / split). Subdirectory clusters are excluded so CD1/CD2 layouts don't false-positive.

Per-book rows are returned (not just counts) so an upcoming dashboard can surface these as clickable health categories.

## Tests

7 new workflow tests (group formation, both conservative exclusions, keeper choice, two-scheme copy detection, single-scheme non-flag). Full suite 1063/1063; vue-tsc (3.3.4) + vite build + eslint + prettier + vitest clean.

### DIFF
--- a/fe/src/components/domain/audiobook/SplitCollectionModal.vue
+++ b/fe/src/components/domain/audiobook/SplitCollectionModal.vue
@@ -1,0 +1,497 @@
+<!--
+  Listenarr - Audiobook Management System
+  Copyright (C) 2024-2026 Listenarr Contributors
+
+  This program is free software: you can redistribute it and/or modify
+  it under the terms of the GNU Affero General Public License as published
+  by the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  GNU Affero General Public License for more details.
+
+  You should have received a copy of the GNU Affero General Public License
+  along with this program. If not, see <https://www.gnu.org/licenses/>.
+-->
+<!--
+  Break a multi-book record apart: server-side clustering (subdirectory /
+  embedded tag / filename stem) with a suggested existing library record per
+  group. Each group can be MOVED to a record (files relocate to its folder),
+  DELETED (a redundant duplicate copy — files removed from disk), or left
+  alone.
+-->
+<template>
+  <Modal :visible="visible" size="lg" title="Split collection" @close="onClose">
+    <ModalBody>
+      <div v-if="loading" class="split-empty">Analyzing files…</div>
+      <div v-else-if="error" class="split-error">{{ error }}</div>
+      <template v-else>
+        <p class="split-intro">
+          {{ clusters.length }} group{{ clusters.length === 1 ? '' : 's' }} detected. Move a group
+          to the record it belongs to (files relocate into its folder), delete a redundant copy, or
+          leave it alone.
+        </p>
+        <div class="split-clusters">
+          <div v-for="c in clusters" :key="c.key" class="split-cluster">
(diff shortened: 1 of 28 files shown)


therobbiedavis:
#717 does not supersede duplicate detection, but this is stacked on #722/#720/#719. Please rebase after that stack is adapted to #717s file identity/ownership contracts; record-only cleanup and any suggested remediation should not bypass the new ownership/mutation boundaries.

kevinheneveld:
Will rebase after the stack settles on #717. Record-only cleanup and the suggested remediations will go through the ownership/mutation boundaries rather than direct row work.

m4bard:
Ran this against a generated library to see what the sweep picks up, and the exclusion for two distinct ASINs drops a case that is worth a second look.

Eight records, four pairs, each pair being two ASINs that Audible's own metadata places at the same series slot:

```
id=1 B008DFUGCQ  'A Princess of Mars'              Edgar Rice Burroughs
id=2 B071YLS9YL  'A Princess of Mars'              Edgar Rice Burroughs
id=5 B007BR5KZA  'The Wonderful Wizard of Oz'      L. Frank Baum
id=6 B002V5CJM4  'The Wonderful Wizard of Oz'      L. Frank Baum
id=7 B002UZJF4U  'The Three Musketeers'            Alexandre Dumas
id=8 B002V0RG8G  'The Three Musketeers'            Alexandre Dumas
id=3 B01FKWL15A  'Twenty Thousand Leagues Under the Sea'  Jules Verne
id=4 B076HSP1FT  '20,000 Leagues Under the Sea'           Jules Verne
```

`GET /library/duplicates` returns `{"duplicateGroups": [], "duplicateCopyBooks": []}`.

Three of those pairs have byte-identical title and author, so they group on title+author and are then dropped by:

```csharp
// Two distinct non-empty ASINs = two catalog editions, not a dupe.
if (distinctAsins >= 2) continue;
```

For all four pairs the catalogue disagrees with the assumption in that comment: both ASINs carry the same `seriesPrimary.asin` and the same position.

```
B008DFUGCQ  Barsoom #1  series asin B007D0J4H0
B071YLS9YL  Barsoom #1  series asin B007D0J4H0
```

So a second ASIN is not reliable evidence of a second edition. It is often the same work re-listed, and when a user has added a book twice that is exactly the shape it takes.

The endpoint is doing what the PR describes, your own tests cover the same-ASIN and no-ASIN groupings, and being conservative about false positives is the right instinct for a screen that offers deletion. The question is only whether an identical title and author under two ASINs should stay excluded, or whether the series identifier gives a safer signal to group on than the ASIN does.

If you want to run it against your own build:

```
git clone https://github.com/m4bard/listenarr-testdata
cd listenarr-testdata
python3 -m venv .venv && .venv/bin/pip install -e ".[dev]"
./tools/check_duplicate_detection.sh <your image>
```

It generates the library, adds both sides of all four pairs, and prints what `GET /library/duplicates` returns. It needs podman or docker, ffmpeg for the one-second silent audio, and an image; `tools/vet-against.sh --branch feat/library-duplicates` builds one from this PR. `ENDPOINT=` overrides the path if it moves.

The four pairs are real catalogue entries and the corpus is re-verified against live metadata, so the series identifiers above are not something I typed in. The audio is generated silence.

Issue #675, opened 2026-06-10
[Bug] Audiobooks list view: oversized scroll area, overscroll past ends, and freeze on fast scrolling


s3ntin3l8:
_Suggested label: `bug`_

## Summary
On the **Audiobooks page in list view** (`?group=books`), the custom virtual scroller in `AudiobooksView.vue` misbehaves:

1. **Oversized scroll area** — the scrollbar/scrollable height is much larger than the actual number of rows, so there is a large empty region below the last item.
2. **Overscroll past both ends** — scrolling fast lets you fling *above* the first row and *below* the last row, beyond the real content.
3. **Freeze / "infinite loading"** — fast up/down scrolling makes the page unresponsive (the loading spinner appears to spin forever and the UI locks up).

Grid view is far less affected. The list view and its scroller exist on `canary` (this is not specific to the per-grouping list-view PR #585).

## Environment
- **Listenarr:** 1.0.11 (`canary`, HEAD `52fd2402`)
- **Frontend:** Vue 3 list view (`fe/src/views/library/AudiobooksView.vue`)
- Browser-independent; needs a library with at least a few hundred books to fling-scroll.

## Steps to reproduce
1. Open the Audiobooks page with a few hundred books and switch to **list view**.
2. Observe the scroll area is taller than the rows warrant (empty space below the last book).
3. Scroll up and down quickly (mouse-wheel flings or trackpad). You can scroll past the top and bottom of the list, and the page becomes unresponsive with a perpetual loading state.

## Expected behavior
The scrollable height equals the rendered content; scrolling clamps at the first and last row; fast scrolling stays smooth and never freezes.

## Root cause (analysis)
The list view uses a hand-rolled uniform-height virtual scroller. Three issues combine (line numbers on `canary`):

- **Single-sample uniform height for variable rows.** `syncMeasuredRowHeight()` (~L1730) measures only the **first** rendered `.audiobook-list-item` and applies that height to every row via `getRowHeight()` (L1705-1724) → `totalHeight = ceil(n / itemsPerRow) * getRowHeight()` (L1790). `measuredRowHeight` is a **single ref shared with grid view** (grid rows ≈ 220-240px). An inflated/stale sample (a tall row, or a leaked grid measurement) sticks and inflates `totalHeight` → oversized scroll area.
[...]

PR #731, opened 2026-07-04
List-view covers, pre-warm + scroller memory fix (requires #676)

Paelsmoessan:
## Requires #676 (prerequisite)

This PR builds on @s3ntin3l8's list-view virtual-scroller fix (**#676**) and needs it to work, #676 should land first. Our changes sit on top of Björn's scroller rather than replacing it.

## Summary

On top of #676:

- **Cover thumbnails + background pre-warm** — covers are generated/cached ahead of time, so scrolling doesn't fire a burst of on-demand cover work (lower memory churn, faster first paint).
- **Cacheable-placeholder for missing covers** — a missing cover returns a cacheable placeholder instead of an uncacheable 500 that grid views re-fire on every render pass.
- **Cover-metadata fallback fix** — read the metadata envelope via reflection instead of `dynamic` (the internal anonymous type across the assembly boundary threw `RuntimeBinderException` → uncacheable 500). This is the original #731 fix, retained.
- **Virtual-scroller memory fix** — drop the `visibleRange` watcher whose re-measure loop leaked DOM/RAM over a session on large libraries. Row height is measured on mount / resize / view-mode change / details toggle instead.

## Notes

- **Scroll-position preservation** (remembering scroll on list ↔ detail) was attempted and **dropped**, it was too flaky to ship reliably.
- This hand-rolled `.vue` scroller works well enough now, but has a ceiling on very large libraries. A virtualization rewrite (TanStack Virtual) has been prototyped and may be pursued separately in a fork; this PR keeps the existing scroller and just stops the freeze/leak.

## Testing

Verified locally on a large library: scrolling is smooth, RAM holds steady over a session, and covers pre-warm.


Paelsmoessan:
Heads up, I just found #676 (closing #675) by @s3ntin3l8, opened a few weeks before this PR, and it fixes the **same** `AudiobooksView.vue` virtual-scroller loop that the first half of this PR addresses. I hit the root cause independently and didn't see his PR at the time, credit for spotting and fixing it first goes to him.

His #676 is actually the more complete scroller fix: it also handles the oversized scroll area, overscroll past both ends, the unreachable last row, and header height in `totalHeight`, plus dedicated Vitest coverage. This PR's scroller change overlaps with it (both touch the shared `measuredRowHeight` / `visibleRange` measurement path, so they'd also conflict).

Where this PR doesn't overlap is the **second** fix: the cover-metadata fallback returning uncacheable 500s (an internal anonymous type read via `dynamic` throwing `RuntimeBinderException`, so those covers re-fetched on every render). #676 doesn't touch that.

Happy to consolidate whichever way you'd prefer:
- rebase this PR down to just the cover-500 fix and let #676 own the scroller, or
- fold both into one credited change if that's easier to review.

Mainly flagging so the effort isn't duplicated and #676 gets a look, it's been open and unreviewed for a few weeks.

therobbiedavis:
@Paelsmoessan Agreed, split these changes. #676 will own the virtual-scroller fix and #675. Remove AudiobooksView.vue from this PR and retain only the cover-metadata fallback fix and its tests.

PR #746, opened 2026-07-09
fix(library): sort audiobook files by natural filename order

kevinheneveld:
## Summary
- EF returns audiobook files in undefined (row-insertion/scan) order, so multi-part books rendered jumbled on the detail page ("Part 116, 100, 049, 065, 095, 016..." instead of ascending order)
- Add a shared `NaturalSort` key builder (treats embedded digit runs as numeric so "Part 2" sorts before "Part 10") and an `AudiobookFileOrdering.InNaturalOrder` helper
- Apply it to the two API projections that feed the frontend file list: the audiobook detail-page endpoint (`LibraryQueryWorkflow.MapDetails`) and the DTO builder used by scan/move broadcasts (`AudiobookDtoFactory.BuildFromEntity`)

## Test plan
- [x] `dotnet build` — clean, 0 warnings
- [x] `dotnet test` — 1189/1189 passing

Draft per PR-pacing convention (several PRs already active); will promote when a slot frees up.

### DIFF
--- a/listenarr.api/Features/Library/LibraryQueryWorkflow.cs
+++ b/listenarr.api/Features/Library/LibraryQueryWorkflow.cs
@@ -8,6 +8,7 @@
  * (at your option) any later version.
  */
 using Microsoft.AspNetCore.Mvc;
+using Listenarr.Domain.Common;
 
 namespace Listenarr.Api.Features.Library;
 
@@ -94,7 +95,7 @@ public sealed class LibraryQueryWorkflow(
             })
             .ToList(),
         tags = audiobook.Tags,
-        files = audiobook.Files?.Select(file => new
+        files = AudiobookFileOrdering.InNaturalOrder(audiobook.Files).Select(file => new
         {
             id = file.Id,
             path = file.Path,
--- a/listenarr.application/Mapping/AudiobookDtoFactory.cs
+++ b/listenarr.application/Mapping/AudiobookDtoFactory.cs
@@ -16,6 +16,8 @@
  * along with this program. If not, see <https://www.gnu.org/licenses/>.
  */
 
+using Listenarr.Domain.Common;
+
 namespace Listenarr.Application.Mapping
 {
     public static class AudiobookDtoFactory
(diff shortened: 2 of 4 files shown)


m4bard:
The diagnosis looks right and the shape of the fix does too. Sorting at the projection rather than in the query keeps it out of EF's hands, and a padded key compared ordinally is the usual way to do this.

One process thing and two edges I found while poking at it.

`NaturalSort` and `AudiobookFileOrdering` arrive without tests. The checklist above reports `dotnet test` at 1189 passing, which is the existing suite rather than new coverage, and CONTRIBUTING asks for tests where applicable. A pure string-to-key function is close to the ideal case for them: no fixtures, no host, and every interesting input fits on one line.

I built `4e7e7c1` and called `NaturalSort.ToKey` directly. Two results worth pinning down.

Zero-padded and bare numbers collapse to the same key:

```
ToKey("Part 01") == "PART 000000000001"
ToKey("Part 1")  == "PART 000000000001"   // equal
```

Two files whose names differ only by that padding therefore tie, and their relative order falls back to whatever EF handed over, which is the ordering this PR exists to remove. Mixed padding inside one book is not unusual in ripped libraries. Whether the tie matters is a judgement call, but right now the behaviour there is unspecified rather than chosen.

The other is a boundary. The key pads numeric runs to 12 digits but parses them with `int.TryParse`, which tops out at 10, so any run above `int.MaxValue` fails to parse, skips padding, and falls back to a plain string:

```
ToKey("Track 2147483647") == "TRACK 002147483647"   // padded
ToKey("Track 2147483648") == "TRACK 2147483648"     // not padded
```

Which then sorts lexicographically among its peers:

```
"Part 30000000000" sorts before "Part 9000000000"
```

No disc or track number reaches that, so this is latent rather than a live bug, and `long.TryParse` closes it. I mention it mostly because it is exactly the sort of thing a test would have pinned and review tends not to.

One I expected to find and did not: a long numeric prefix does not break the runs after it, so `9780141036144 Part 2` still sorts before `9780141036144 Part 10`.

Happy to send the test cases as a PR against your branch if that is useful, rather than leaving it as a review note.

PR #749, opened 2026-07-09
fix(search): unwrap bare-array response in searchAudibleByTitleAndAuthor

kevinheneveld:
## Summary
- `POST /search` in Advanced mode returns a bare JSON array, not `{results, totalResults}`
- `searchAudibleByTitleAndAuthor` only ever read `response.results`, which doesn't exist on a plain array, so it silently discarded every result and always reported empty — confirmed live against a real Audible title/author search where the backend genuinely found the match
- The other two frontend callers of this same endpoint (`getAuthorLookup`, `advancedSearch`) already defensively handle both shapes (`Array.isArray(resp) ? resp : (resp?.results ?? [])`); this one never did

## Test plan
- [x] `vue-tsc --build` — clean
- [x] `vite build` — clean
- [x] `vitest run` — 394/394 passing (2 new regression tests: bare-array shape + object-wrapped shape)

Draft per PR-pacing convention (several PRs already active); will promote when a slot frees up.

### DIFF
--- a/fe/src/__tests__/api.searchAudibleByTitleAndAuthor.spec.ts
+++ b/fe/src/__tests__/api.searchAudibleByTitleAndAuthor.spec.ts
@@ -1,0 +1,85 @@
+/*
+ * Listenarr - Audiobook Management System
+ * Copyright (C) 2024-2026 Listenarr Contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+import { describe, it, expect, vi, afterEach } from 'vitest'
+
+describe('ApiService searchAudibleByTitleAndAuthor', () => {
+  afterEach(() => {
+    vi.restoreAllMocks()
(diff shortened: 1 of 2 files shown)


m4bard:
Confirmed the shape claim at the backend rather than from the response, and it is broader than the description says. `StructuredSearchWorkflow` returns a bare array from both branches, not just the advanced one:

```
ExecuteSimpleSearchAsync    -> StructuredSearchWorkflowResult.Ok(mapped)       // Task.WhenAll(...) array
ExecuteAdvancedSearchAsync  -> StructuredSearchWorkflowResult.Ok(flatMapped)   // same
```

So `resp.results` was never going to be anything but `undefined` on that endpoint under any mode, and the `?? []` turned every search into a clean empty result. Your normalisation matches what `api.ts` already does at lines 299, 560 and 594, so this brings the last caller into line rather than inventing a convention.

Two things I ran into while checking, neither of them objections to the change.

I could not find a caller. Grepping `fe/` on canary for `searchAudibleByTitleAndAuthor` returns the definition in `api.ts` and a set of mocks in `AddNewView.spec.ts`, and nothing in views, components or stores. The mocks use optional chaining (`apiService.searchAudibleByTitleAndAuthor?.mockResolvedValue`), which reads like they are guarding against exactly that. You mention confirming it live against a real title and author search, so presumably there is a path I am not seeing, or it is wired up on another branch. Worth knowing either way: if nothing calls it yet, this is a latent bug rather than a user-visible one, which changes how urgent it looks without changing whether it is right.

The second is in the same function and predates this PR. The body sends `page` and `limit` flat:

```js
const body = { mode: 'Advanced', title, author, page, limit, region }
```

but `SearchRequest` carries them nested, as `Pagination { Page = 1, Limit = 50 }`, and the advanced path reads `req.Pagination.Limit` before clamping. A flat `limit` does not bind there, so the backend uses its default of 50 regardless of what the caller asked for, and the `page` argument goes nowhere at all. That also puts a ceiling on the new `totalResults: results.length`, which reports what came back rather than what exists. Nothing here makes that worse, and `results.length` is the only honest number available from a bare array, but it is a separate fix if anyone ever pages this.

Issue #737, opened 2026-07-05
[Bug] Download-imported audiobooks missing codec/bitrate/duration (probe empty at import time)


Paelsmoessan:
### Summary
Audiobooks brought in through the **download client** (`AudiobookFile.Source = "download"`) frequently get their file records written with **null codec, null bitrate, and duration 0.0** — only the filesystem size is captured. Files added by the regular **library scan** (`Source = "scan"`) are probed correctly.

### Evidence (my library)
Split is by `Source`, not by "imported":
- `scan` files: **30661 / 31508 have codec (~99% good)**
- `download` files: **5 / 661 have codec (~99% null)**

The null files have a real `Size` (from the filesystem) but `DurationSeconds = 0.0` and null `Codec`/`Bitrate`, i.e. they were never successfully probed. It's ongoing (most recent affected file is days old), not just legacy data.

### Not a "doesn't probe" bug
`DownloadImportService` **does** call the same `AudiobookFileService.EnsureAudiobookFileAsync(audiobook, destination, "download")` that the scan path uses, and that method ffprobes. So the path isn't skipping the probe, the probe **returns empty at download-import time**, and the row is persisted with nulls anyway. The same file scanned later populates correctly.

### Hypothesis (to confirm)
A timing/readiness issue at import: the file may not be fully settled, or ffprobe isn't ready yet (it auto-downloads at startup and "continues without ffprobe" on failure), so extraction yields nothing. Two fixes likely needed:
1. Make the download-import probe reliable (ensure ffprobe is ready + file is settled; or re-probe on failure instead of persisting nulls).
2. A one-time backfill/re-probe of already-affected `Source="download"` files.

### Relationship
Provides the quality data #582's upgrade-and-replace logic needs to compare existing vs candidate files. Related: #542.

---
*Corrected 2026-07-06 after DB + code investigation. Original report had the direction backwards (blamed library-import); the actual split is by download vs scan source, verified above.*

Paelsmoessan:
**Root cause confirmed (supersedes the timing/readiness hypothesis above)**

After deeper investigation (DB + service logs + manual ffprobe, with a fix verified live), this is **two pre-existing extraction bugs**, not an import-timing issue:

[...]

Issue #736, opened 2026-07-05
[Feature Request] Recycling Bin for safe deletes and upgrade-replacements


Paelsmoessan:
### Summary
Add a Sonarr/Radarr-style Recycling Bin: when Listenarr deletes or replaces files (removing a book, or replacing a lower-quality file on upgrade), **move** them to a configurable recycle-bin folder with an auto-cleanup age, instead of hard-deleting.

### Why
Deleting user files is irreversible. Any automated upgrade-and-replace strategy (#582) needs a safety net: if the quality decision is ever wrong, the file should be recoverable, not gone. This is exactly how Sonarr/Radarr de-risk the same feature, and it lets the upgrade logic be decisive without fear of data loss.

### Scope
- Setting: recycle-bin path + cleanup age (Media Management).
- A service that moves files to the bin (preserving relative structure + a timestamp) instead of `File.Delete` / `Directory.Delete`.
- A cleanup job that purges bin items older than the configured age.
- Route the existing delete sites through it (book removal, upgrade-replace). It would also be a safety net for the destructive folder move/delete path in #438.
- A bit of settings UI.

### Relationship
Prerequisite for a safe upgrade-and-replace strategy: #582.

Issue #542, opened 2026-04-24
Incorrect size for audiobook given


djjudas21:
I noticed the total size of the audiobook given in the summary is not the sum of all the MP3 files that are part of that audiobook.

I also noticed that not every audiobook gives the total size in the summary, and I haven't been able to figure out why some have it and some don't.

Happy to provide any extra info that would help with debugging.

<img width="2560" height="1293" alt="Image" src="https://github.com/user-attachments/assets/f496af88-1da4-4cbd-bff3-2107b7454f81" />

SwathiPulipati:
Hi! I'm a student in the CodePath AI301 program and I'd like to work on this issue.  Is it still available? It's my first time contributing to an open-source project, but I'm eager to learn about the repo and help bug fix.

therobbiedavis:
@SwathiPulipati This one is available. I assume you and @NourdotSiwar are in the same class or are working together somehow and need this for some type of credit?

NourdotSiwar:
Yep we are,this is the class we are part of if you are curious: https://www.codepath.org/courses/applied-ai-engineering

SwathiPulipati:
We are in the same class, yea! I'm excited to get started. Thanks!

NourdotSiwar:
Hey @SwathiPulipati, wanted to check on your progress for this issue. Let me know if you need help with it or if you want me to take over.

SwathiPulipati:
Hey, I'm almost done with it, should have an pr in the next couple of days. Thank you though!

m4bard:
Both halves of this come from the same thing: the size shown for a book is not a total. It is a single file's size, stored on the audiobook row, and nothing ever recomputes it from the files.

Reproduced the second half on canary with a generated 40-chapter book:

```
  on disk         : 223,520 bytes across 40 files
  linked by scan  : 223,520 bytes across 40 files
  reported size   : not set
```

The scan links all forty files and records every size correctly, so the per-file data is fine. `Audiobook.FileSize` is simply never populated by that path, which is why the summary shows nothing.

Grepping canary for assignments to it, there are four, and none of them sum anything:

```
RenameService.Helpers.cs:49        audiobook.FileSize = primaryFile.Size   // ONE file
ScanJobProcessor.cs:370            audiobook.FileSize = null
LibraryManualScanWorkflow.cs:348   audiobook.FileSize = null
[...]

PR #752, opened 2026-07-12
fix(downloads): make import quality comparison profile-aware (#582)

### LINKED ISSUE #582: Quality filters: Flaws in the current logic
## Context
When importing files from downloads for audiobooks with already imported files, we try to determine the following informations:
* Best quality of existing files
* Quality of the current file being processed
If the audio book has a quality profile set: We use those to skip file when the quality is not better than the one we already have

Effectively, I could not write tests for that behavior nor could I determine precisely which use case it covers.

## Suggestion
I suggest that, when we import a download, the following should be done:
* Determine the quality of the existing audiobook
* Determine the quality of the download
* Check the defined profile strategy
  * Skip the whole import if the quality is not better
  * Remove all existing files if the quality is better

I'm not sure what heuristic we should use to determine the quality of those though, a first approach could be to compare to take the best or worst (this is already what's implemented)

Regarding the quality determination algorithm, see #549 as it probably could use some improvements

jaredvincent414:
## Summary

When a completed download is imported for an audiobook that already has files, Listenarr is
supposed to consult the audiobook's quality profile and skip the incoming file when it isn't an
upgrade. That decision ran through `ImportQualityEvaluator.IsAcceptable`, which accepted a
`QualityProfile` but **never read it** — it compared a regex-extracted numeric bitrate instead:

```csharp
return !TryParseBitrate(candidate, out var c) || !TryParseBitrate(existing, out var e) || c >= e;
```

Any non-numeric quality (an existing **FLAC**) fails that parse and falls through to "acceptable",
so a lossy MP3 gets imported on top of a lossless file and the audiobook ends up with both.

A profile-aware, lossless-aware comparator already exists in the domain — `QualityMatcher` — and
the automatic-search path already uses it. Import was the only path still on the old regex
evaluator. This PR routes import onto it.

Fixes #582. Related: #549 (quality determination algorithm).

## Acceptance Criteria

- [x] Tests added — failing repro (`QualityGating_LosslessExisting_SkipsLossyDownload`) plus a
      4-case gating matrix (numeric upgrade, cross-codec upgrade, equal-quality, unknown/no-profile)
- [x] All tests passing — backend and frontend (real output pasted under Testing below)
- [x] Follows style guide — pre-commit/pre-push hooks (format, layering, no `async void`,
      type-check) pass
- [x] No breaking changes to any public contract — one intentional *behavior* change is called out
      in Notes below (lossy-over-lossless is now blocked even with no quality profile configured)

### Added
- `QualityMatcher.IsLossless(AudioQualityInput)` — public wrapper over the existing private lossless check, so callers outside the profile path can still honour lossless-vs-lossy.
- `ImportQualityEvaluator.FromFile(...)` / `FromMetadata(...)` — project an `AudiobookFile` and an incoming download's `AudioMetadata` into the shared `AudioQualityInput` vocabulary.
- `ImportQualityEvaluator.Describe(...)` — readable quality label for the import skip log.
- Four quality-gating tests: numeric upgrade, cross-codec (lossless-over-lossy) upgrade, equal-quality reimport, and unknown-quality/no-profile.

### Changed
- `ImportQualityEvaluator.IsAcceptable` now takes `AudioQualityInput` and ranks both sides through the profile's rung priority. When there is no profile, an empty profile, or a side the profile cannot rank, it falls back to codec/bitrate facts — lossy never lands on top of lossless, and genuinely unknown quality still imports rather than being silently dropped.
- `DownloadImportService` picks the hardest-to-beat existing file via the same comparator, and gates the candidate through it.

### Fixed
- A lossy download no longer imports over an existing lossless file (#582).
- The "best existing quality" loop no longer hardcodes MP3 bitrate buckets (`"MP3 320kbps"` etc.), which mislabelled every non-MP3 file as MP3.

### Removed
- The regex `TryParseBitrate` decision helper.

## Testing

```
$ dotnet test
Passed!  - Failed:     0, Passed:  1020, Skipped:     0, Total:  1020, Duration: 50 s - Listenarr.Tests.dll (net10.0)

$ cd fe && npm run test:unit
 Test Files  74 passed | 1 skipped (75)
      Tests  389 passed (389)

$ cd fe && npm run type-check
(clean — no errors)

$ dotnet build
0 Warning(s)
0 Error(s)
```

Pre-commit (format, layering, `async void`) and pre-push (format, type-check, FE tests) hooks pass.

The reproduction is `QualityGating_LosslessExisting_SkipsLossyDownload`: an audiobook with an existing FLAC and a profile ranking FLAC above MP3 receives a lossy MP3 download. Before this change the MP3 was imported and the audiobook held two files; now it is skipped.

Worth being straight about the other four tests: they would also pass against the old code (its regex fell through to "acceptable" for any non-numeric label, so it imported in all four cases). They exist to guard against **over-correction**, not to prove the fix. The load-bearing one is `QualityGating_EqualQuality_IsImported` — it fails if the gate is ever tightened to the strictly-better rule search uses, which would skip every part of a multi-file audiobook whose quality merely matches what is already on disk.

## Notes

**Import compares "not worse"; search compares "strictly better".** That asymmetry is deliberate and now documented in the code. Import must accept equal quality, or multi-file audiobooks would have every part skipped.

**The delete/replace half of #582 is deliberately not here.** The issue also proposes removing existing files when the incoming quality is better. That is a separate, riskier change and — per @Paelsmoessan's comment on the issue — is blocked on #542 (multi-file size aggregation), #736 (recycling bin, so a wrong quality call is not data loss), and #737 (imported files have no stored codec/bitrate). Deferring it keeps this PR to one bug fix.

**Behaviour change for users with no quality profile.** The fallback means lossy-over-lossless is now blocked even when no profile is configured. This is intended — previously nothing gated those users — but it is a real behaviour change worth flagging.

### DIFF
--- a/listenarr.application/Downloads/Import/DownloadImportService.cs
+++ b/listenarr.application/Downloads/Import/DownloadImportService.cs
@@ -90,31 +90,20 @@ namespace Listenarr.Application.Downloads.Import
                 try
                 {
                     // Precompute audiobook and best existing quality to avoid import-order races
-                    string? bestExisting = null;
-                    QualityProfile? abProfile = null;
-
-                    abProfile = audiobook.QualityProfile;
+                    AudioQualityInput? bestExisting = null;
+                    var abProfile = audiobook.QualityProfile;
 
                     if (audiobook.Files != null && audiobook.Files.Count != 0)
                     {
-                        foreach (var f in audiobook.Files)
+                        foreach (var existingFile in audiobook.Files)
                         {
-                            string q = string.Empty;
-                            if (!string.IsNullOrEmpty(f.Format)) q = f.Format;
-                            if (f.Bitrate.HasValue)
-                            {
-                                var kb = f.Bitrate.Value / 1000;
-                                if (kb >= 320) q = "MP3 320kbps";
-                                else if (kb >= 256) q = "MP3 256kbps";
-                                else if (kb >= 192) q = "MP3 192kbps";
-                                else if (kb >= 128) q = "MP3 128kbps";
-                            }
-                            if (string.IsNullOrEmpty(q) && !string.IsNullOrEmpty(f.Path)) q = ImportQualityEvaluator.Determine(null, f.Path);
+                            var existingInput = ImportQualityEvaluator.FromFile(existingFile);
 
-                            if (string.IsNullOrEmpty(bestExisting)) bestExisting = q;
-                            else if (!string.IsNullOrEmpty(q) && !string.IsNullOrEmpty(bestExisting) && abProfile != null && ImportQualityEvaluator.IsAcceptable(q, bestExisting, abProfile))
+                            // "Not worse than the current best" climbs the list to the hardest quality to beat.
+                            if (bestExisting is null
+                                || ImportQualityEvaluator.IsAcceptable(existingInput, bestExisting.Value, abProfile))
                             {
-                                bestExisting = q;
+                                bestExisting = existingInput;
                             }
                         }
                     }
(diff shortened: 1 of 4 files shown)


jaredvincent414:
@Paelsmoessan — this addresses the lossy-over-lossless import bug in #582. Per your comment on the issue, I've left the delete/replace-on-upgrade half out of scope for now since it's blocked on #542, #736, and #737. Would appreciate a look when you have a chance.

PR #754, opened 2026-07-12
Fix Discord/NTFY notifications double-posting via generic sender fall-through

### LINKED ISSUE #753: Discord/NTFY notifications are double-posted; Discord duplicate rejected with 400 {"embeds": ["0"]}
### Describe the bug

Every Discord notification is sent to the webhook **twice**. The Discord-specific handler in `NotificationService.Webhooks.cs` sends the notification (multipart with cover attachment, or plain JSON), but does not `return` afterward — execution falls through to the generic webhook sender at the bottom of `SendNotificationAsync`, which posts a second copy of the same notification.

Depending on configuration, this shows up two different ways:

1. **Duplicate messages** — when the generic payload is valid, the Discord channel receives the same notification twice.
2. **Spurious 400 errors in the log** — when the configured base URL is not an absolute `http(s)` URL (e.g. the default `UrlBase` of `"/"` in `config.json`), the generic payload's embed thumbnail is a *relative* path (`/api/v1/images/<asin>`), which the Discord API rejects:

```
[INF] Sending Generic POST to https://discord.com/api/webhooks/... with body: {"content":"Dracula by Bram Stoker has been added", ... "thumbnail":{"url":"/api/v1/images/B0078PA1OA"} ...}
[WRN] Failed to send notification to https://discord.com/api/webhooks/...: BadRequest - {"embeds": ["0"]}
```

The log makes delivery look broken even though the first (Discord-specific) send already succeeded — this cost me a while when setting up Discord notifications, since I saw a 400 on every `book-added` event.

The **NTFY handler has the same fall-through**, so NTFY targets also receive the notification twice (once as a proper NTFY message with Title/Priority/Tags headers, once as generic JSON).

The Pushover and Telegram handlers already `return` after handling, so they are unaffected.

### To Reproduce

1. Configure a Discord webhook URL as the notification target (default `UrlBase` of `"/"`).
2. Trigger any notification with an image, e.g. add a book to the library (`book-added`).
3. Observe the log: the Discord-specific send succeeds (message appears in the channel with cover attachment), followed by a `Sending Generic POST` to the same webhook that fails with `BadRequest - {"embeds": ["0"]}`.

### Expected behavior

One notification per event per target: the service-specific handler should be terminal for URLs it handles, with no second post through the generic sender.

### Environment

- Version: 1.2.2 (canary)
- OS: Windows 11 (dev build from source), also applies to current `canary` sources

### Additional context

I have a fix ready (two `return` statements plus regression tests asserting exactly one HTTP request per Discord/NTFY notification, verified failing before the fix) — PR incoming.

dny238:
Fixes #753

### Problem

The Discord and NTFY handlers in `SendNotificationAsync` (`NotificationService.Webhooks.cs`) do not `return` after handling the notification, so execution falls through to the generic webhook sender and every Discord/NTFY notification is posted a **second time**.

For Discord, the duplicate is built by `CreateDiscordPayload`; when the configured base URL is not an absolute `http(s)` URL (e.g. the default `UrlBase` of `"/"`), its embed thumbnail is a relative path, which the Discord API rejects with `400 {"embeds": ["0"]}` — so the log shows a delivery failure on every notification even though the first (Discord-specific) send succeeded. When the payload is valid, the channel simply receives two copies.

The Pushover and Telegram handlers already `return` and are unaffected.

### Fix

Add the missing `return;` at the end of the Discord and NTFY branches, making each service-specific handler terminal for URLs it handles — consistent with the existing Pushover/Telegram behavior.

### Tests

Adds `NotificationServiceSingleDispatchTests` with two regression tests asserting exactly **one** HTTP request is made per Discord/NTFY notification (using the same mocked-`HttpMessageHandler` pattern as the existing `NotificationServiceTests`).

Verified the tests fail before the fix and pass after:

```
WITHOUT fix: Failed! - Failed: 2, Passed: 0  (two HTTP requests observed per notification)
WITH fix:    Passed! - Failed: 0, Passed: 2
```

Full notification test group (`--filter FullyQualifiedName~NotificationService`, 13 tests) passes.

Also verified end-to-end against a live Discord webhook on a dev instance: one message delivered (with cover attachment), no `Sending Generic POST` follow-up, no 400.

### DIFF
--- a/listenarr.infrastructure/Notifications/Delivery/NotificationService.Webhooks.cs
+++ b/listenarr.infrastructure/Notifications/Delivery/NotificationService.Webhooks.cs
@@ -81,6 +81,8 @@ namespace Listenarr.Infrastructure.Notifications.Delivery
                 }
 #pragma warning restore CA1031
 
+                // Discord handled; do not fall through to the generic sender (would double-post).
+                return;
             }
 
             // NTFY-specific handling (https://docs.ntfy.sh/publish/)
@@ -139,6 +141,8 @@ namespace Listenarr.Infrastructure.Notifications.Delivery
                     _logger.LogError(ex, "Invalid operation while sending NTFY notification to {WebhookUrl}", LogRedaction.RedactText(webhookUrl, LogRedaction.GetSensitiveValuesFromEnvironment()));
                 }
 
+                // NTFY handled; do not fall through to the generic sender (would double-post).
+                return;
             }
 
             // Pushover (https://pushover.net/api)
--- a/tests/Features/Infrastructure/Notifications/Delivery/NotificationServiceSingleDispatchTests.cs
+++ b/tests/Features/Infrastructure/Notifications/Delivery/NotificationServiceSingleDispatchTests.cs
@@ -1,0 +1,111 @@
+/*
+ * Listenarr - Audiobook Management System
+ * Copyright (C) 2024-2026 Listenarr Contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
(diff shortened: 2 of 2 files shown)


dny238:
This one's been open a while — it's now rebased onto current canary and mergeable, so it's ready for a CI run whenever convenient. It's a small, self-contained fix (Discord/NTFY notifications double-posting via generic-sender fall-through) with a regression test that asserts exactly one HTTP request per notification. Happy to adjust anything if needed. Thanks for taking a look!

PR #756, opened 2026-07-13
fix(search): zero configured indexers returned five fake mock releases into the real grab pipeline

t3chnaztea:
## The bug

With zero indexers configured/enabled, `IndexerSearchWorkflow.SearchIndexersAsync` logged a warning and then returned `GenerateMockIndexerResults(query)`: five synthetic releases with fabricated titles, sizes, and URLs. Because this method also serves the automatic-search path, those fake releases flowed into real scoring and the grab decision, where one could be selected and handed to a download client.

This is not only reachable by a manual search click: `AutomaticSearchService` (a `BackgroundService` on a five-minute initial delay and recurring interval) also calls `SearchAsync(isAutomaticSearch: true)`, so on an install with no indexers configured yet the fake releases enter scoring unattended, on a timer. The fabricated entries carry realistic qualities (MP3 64kbps through FLAC), sizes, and non-zero seeders; the only tell is the `Mock Indexer` source string, which nothing in the pipeline gates on.

## The fix

Zero indexers now returns an empty result set with a clear warning ("No indexers configured or enabled; search returned no results"). Both `GenerateMockIndexerResults` overloads are deleted — a repo-wide search (backend, tests, `fe/`) found no other consumers, and no dev/debug flag gated them.

## Tests

- `SearchIndexers_NoIndexersConfigured_ReturnsEmptyWithoutSyntheticResults`

Full suite green: 1190 passed / 0 failed; `dotnet format --verify-no-changes` clean.

### DIFF
--- a/listenarr.application/Search/Indexers/Common/IndexerSearchWorkflow.cs
+++ b/listenarr.application/Search/Indexers/Common/IndexerSearchWorkflow.cs
@@ -63,8 +63,8 @@ public class IndexerSearchWorkflow
 
         if (!indexers.Any())
         {
-            _logger.LogWarning("No indexers configured, returning mock results for query: {Query}", query);
-            return GenerateMockIndexerResults(query);
+            _logger.LogWarning("No indexers configured or enabled; search returned no results for query: {Query}", query);
+            return results;
         }
 
         var searchTasks = indexers.Select(async indexer =>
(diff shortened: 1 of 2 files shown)


m4bard:
Traced this one on `canary` because the severity depends entirely on where those five results end up, and that is worth confirming rather than assuming. Everything you describe checks out, and there is one more caller than the description mentions.

The path from an empty indexer list to a grab decision:

```
IndexerSearchWorkflow.SearchIndexersAsync   line 64: if (!indexers.Any()) -> GenerateMockIndexerResults(query)
SearchService.SearchAsync                   line 111: isAutomaticSearch branch, converts them to SearchResult
DownloadService.cs                          line 150: SearchAsync(isAutomaticSearch: true)
                                            line 162: qualityProfileService.ScoreSearchResults(...)
                                                      then "Only consider non-rejected, score > 0 results"
```

So they reach scoring and selection rather than stopping at the API boundary. Whether one is actually grabbed depends on the profile, which I have not exercised, but nothing structurally excludes them: the fabricated entries carry realistic `Quality` values from `MP3 64kbps` up to `FLAC`, sizes between 200MB and 1.5GB, and non-zero seeders. `Source` is set to `Mock Indexer`, which is the only tell, and it is a display string rather than anything the pipeline gates on.

The addition to your description: `SearchAsync(isAutomaticSearch: true)` also has a second caller in `AutomaticSearchService`, which is a `BackgroundService` on a five minute initial delay and a recurring interval. So on an install with no indexers configured yet, this is not only reachable by someone clicking search. It runs unattended on a timer.

Your "no other consumers" claim also holds. Grepping `canary` for `GenerateMockIndexerResults` across the backend, tests and `fe/` returns the definition, the internal single-argument overload, and the one call at line 67. Nothing else references it, and nothing gates it behind a debug flag, so deleting both overloads takes nothing else with it.

Returning empty with a warning is the right shape. Mock data that is indistinguishable from real data at the point it is consumed is worse than an empty result, because an empty result is honest about knowing nothing.

t3chnaztea:
Good catch on `AutomaticSearchService`. That's worse than the manual path since it needs no user action at all; I've folded it into the PR description. Thanks for tracing it on canary.

PR #757, opened 2026-07-13
fix(search): one indexer timeout no longer discards every other indexer's results

t3chnaztea:
## The bug

The per-indexer catch in `IndexerSearchWorkflow.SearchIndexersAsync` excluded `OperationCanceledException` from its filter. An `HttpClient` per-request timeout throws `TaskCanceledException` — a subclass of `OperationCanceledException` — so a single slow or timing-out indexer escaped the per-indexer handling, propagated out of `Task.WhenAll`, and discarded every healthy indexer's results for the whole search. On the automatic path that zeroes out the entire cycle for that book until the next sweep.

## The fix

Catch `OperationCanceledException` per-indexer, log a warning naming the indexer, and return an empty list for that indexer only — matching how other per-indexer failures already behave. No workflow-level `CancellationToken` flows into this search (verified across the call chain, incl. `IIndexerSearchProvider.SearchAsync`), so an OCE here can only be a client timeout; there is no shutdown cancellation to preserve. `OutOfMemoryException`/`StackOverflowException` remain excluded.

## Tests

- `SearchIndexersAsync_OneIndexerTimesOut_ReturnsHealthyIndexerResults`
- `SearchIndexersAsync_AllIndexersTimeOut_ReturnsEmptyWithoutThrowing`

Full suite green: 1191 passed / 0 failed; `dotnet format --verify-no-changes` clean.

### DIFF
--- a/listenarr.application/Search/Indexers/Common/IndexerSearchWorkflow.cs
+++ b/listenarr.application/Search/Indexers/Common/IndexerSearchWorkflow.cs
@@ -78,7 +78,16 @@ public class IndexerSearchWorkflow
                 _logger.LogInformation("Found {Count} results from indexer {Name}", indexerResults.Count, indexer.Name);
                 return indexerResults;
             }
-            catch (Exception ex) when (ex is not OperationCanceledException && ex is not OutOfMemoryException && ex is not StackOverflowException)
+            catch (OperationCanceledException ex)
+            {
+                // No workflow-level cancellation token flows into this search, so an
+                // OperationCanceledException here is an HttpClient per-request timeout
+                // (TaskCanceledException derives from OperationCanceledException). Contain it
+                // to this indexer so a single slow indexer can't abort every other one's results.
+                _logger.LogWarning(ex, "Timed out searching indexer {Name} for query: {Query}", indexer.Name, query);
+                return new List<IndexerSearchResult>();
+            }
+            catch (Exception ex) when (ex is not OutOfMemoryException && ex is not StackOverflowException)
             {
                 _logger.LogError(ex, "Error searching indexer {Name} for query: {Query}", indexer.Name, query);
                 return new List<IndexerSearchResult>();
--- a/tests/Features/Application/Search/Indexers/Common/IndexerSearchWorkflowTests.cs
+++ b/tests/Features/Application/Search/Indexers/Common/IndexerSearchWorkflowTests.cs
@@ -1,0 +1,114 @@
+using Listenarr.Tests.Builders;
+using Listenarr.Tests.Common;
+
+namespace Listenarr.Tests.Features.Application.Search.Indexers.Common
+{
+    [Trait("Name", "IndexerSearchWorkflowTests")]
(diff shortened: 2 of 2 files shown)


m4bard:
While reading through this fix, I think the same exception-shape issue recurs at four more sites in the automatic search loop that this PR does not reach.

One layer up from the indexer search, `SearchResultScorer.cs:136` scores each result in its own `Task.WhenAll` fan-out. Same pattern, but its catch still excludes `OperationCanceledException`, so one slow lookup could fault the batch and discard every other result's score.

Further along, in `AutomaticSearchService.cs`, three more catches (lines 205, 242 and 314: the two debug broadcasts and the download-client submission) exclude `OperationCanceledException` too. None of them fan out, but a timeout there reaches the per-book catch that logs "canceled/timed out" and discards the whole book's already-done search rather than that one step.

I have this PR's branch on my local build. Would you rather widen it to cover these, or keep it scoped and have them filed as a follow-up that references it? I am not planning a competing PR either way.

t3chnaztea:
Thanks for tracing these. I checked all four on the branch; the catch shape matches, but the blast radius differs enough that I'd rather keep this PR to the fan-out it fixes and file the rest separately. Filed as #949 with the per-site breakdown.

Short version:

- `AutomaticSearchService.cs:314` is the one that can actually fire (`StartDownloadAsync` is an HTTP call to the download client). But it escapes to the per-book `catch (OperationCanceledException)` at line 118, which logs and moves on to the next book, and nothing after the submission step exists to lose. Since the per-book `try` never reaches the `LastSearchTime` update, the book is retried next cycle instead of after the 6-hour cooldown. That's a log-attribution problem, not lost work; widening the catch as-is would push the retry out 6 hours.
- The two broadcast catches (205, 242) have no realistic cancellation path today (`SendCoreAsync`, default token, in-process hub), but if one ever threw it would abort the book before scoring or submission over a debug message. Cheap to widen and worth doing.
- `SearchResultScorer.cs:136` has no cancellation path at all: `GetByIdAsync` is EF `FindAsync` on SQLite with no token, so no `OperationCanceledException` can come out of it.

Keeping #757 scoped since its test pins the specific fan-out behavior and the follow-up doesn't need to block on it.

m4bard:
Both of those corrections are right, and I have re-read them against canary `a630572e9`.

`GetByIdAsync` at `SearchResultScorer.cs:125` takes no token, so nothing under the catch at `:136` can raise an `OperationCanceledException`. The other one matters more. A timeout out of `StartDownloadAsync` at `AutomaticSearchService.cs:308` escapes the catch at `:314` into the per-book `catch (OperationCanceledException)` at `:118`, which sits above the `LastSearchTime` write at `:108`. So the book keeps a null or stale `LastSearchTime` and stays eligible under the predicate at `AudiobookRepository.cs:465`, which is the fast retry you described. Widening `:314` as it stands would stamp it and give that up. I had not followed the escape path through when I wrote my comment.

Keeping this scoped to the fan-out its test pins makes sense to me. I went through the rest on #949 the same day, so this is only here to put the correction on the thread that carried the claim.

PR #761, opened 2026-07-14
fix(moves): organize moves into a subfolder of the source no longer fail as overlapping

kevinheneveld:
## The bug

Naming templates that append subfolders (narrator, edition, …) make organize request moves like:

```
/audiobooks/Stephen King/Elevation  →  /audiobooks/Stephen King/Elevation/Stephen King
```

— a target **inside its own source**. The copy-walk mover can't execute that shape (the recursive copy would descend into its own output, and the post-copy `Directory.Delete(source, true)` would delete the result), so the overlap guard failed the whole class with **"Source and target paths overlap"**.

On my live library this hit **97 consecutive organize jobs** — every book whose folder didn't yet contain the template's narrator subfolder.

## The fix

Descend-shaped moves don't need to copy anything: two same-volume atomic renames do it safely —

1. rename the source aside (same parent, atomic): `Elevation` → `Elevation.moving-<jobId>`
2. recreate the vacated path: `mkdir Elevation`
3. rename into place: `Elevation.moving-<jobId>` → `Elevation/Stephen King`

with best-effort rollback if step 2/3 fails (the recreated skeleton is empty by construction, so it's dropped and the aside-rename is reversed). The inverse shape — flattening a folder up into its populated ancestor — stays blocked exactly as before.

Implementation notes:
- The rename logic lives in a `MoveJobProcessor.DescendMove` partial; the processor itself gains only the branch points (it was 1 line under the 500-line architecture rule).
- The existing completion path (history entry, notifications, scan enqueue, status) is shared unchanged.

## Testing

- New integration tests drive `ProcessJobAsync` directly: the descend shape completes with files physically renamed into the child folder and a `Moved` history entry; the upward-overlap shape still fails with the original error and leaves files untouched.
- Full suite: 1191 passed.
- Live-verified on my instance before porting here: all 97 previously-failed jobs were requeued after this fix — 90 completed (the other 7 were stale requests whose source folders no longer exist).

Heads-up: #717 also works in the moves area — happy to rebase this on top of it if that lands first; the footprint here is deliberately small.

### DIFF
--- a/listenarr.infrastructure/Library/Moving/MoveJobProcessor.DescendMove.cs
+++ b/listenarr.infrastructure/Library/Moving/MoveJobProcessor.DescendMove.cs
@@ -1,0 +1,101 @@
+/*
+ * Listenarr - Audiobook Management System
+ * Copyright (C) 2024-2026 Listenarr Contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+using Listenarr.Domain.Common;
+using Microsoft.Extensions.Logging;
+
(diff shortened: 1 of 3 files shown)


therobbiedavis:
Closing as superseded by #717. The final #717 branch supports target-inside-source moves through the durable AudiobookContentMoveService workflow, with persisted manifests, ownership/recovery evidence, restart safety, and regressions at both the move-service and processor levels. This covers the organize/narrator-subfolder scenario without relying on the older direct two-rename path. #717 references this PR.

kevinheneveld:
Agreed — a durable manifest/ownership workflow is the right systemic answer; the two-rename path was always a targeted patch for the organize/narrator-subfolder shape. Once #717 lands I'll re-run my target-inside-source scenarios (`{Author}/{Series}/{Title}/{Narrator}` templates) against it and report back if anything regresses.

PR #762, opened 2026-07-14
fix(metadata): treat Audible's empty product stub as no answer so the provider chain can continue

kevinheneveld:
## The bug

An ASIN that's missing from Audible's catalog doesn't 404 — the catalog API returns a product node with **every attribute null**, which `GetBookMetadataAsync` maps into a non-null but completely empty `AudibleBookResponse`. `AudiobookMetadataService.GetMetadataAsync`'s provider loop only checks `result != null`, so the empty shell counted as a win:

1. the caller receives junk (`source: Audible`, all fields null) — "fill missing from online" reports success while filling nothing;
2. every later provider in the chain (Audnexus today, anything added later) is silently skipped, even when it could have answered.

Live case that surfaced it: an Amazon-exclusive audiobook edition whose ASIN returns the empty stub from **all ten** regional Audible catalog APIs.

## The fix

A titleless `AudibleBookResponse` is treated as no answer (logged) and the loop falls through to the next configured source. One conditional; no behavior change for any ASIN Audible actually carries.

## Testing

Full suite: 1189 passed. Verified live on my instance: the previously-junk ASIN now falls through the chain instead of short-circuiting with an empty payload.

### DIFF
--- a/listenarr.application/Metadata/Core/AudiobookMetadataService.cs
+++ b/listenarr.application/Metadata/Core/AudiobookMetadataService.cs
@@ -115,6 +115,18 @@ namespace Listenarr.Application.Metadata.Core
                         continue;
                     }
 
+                    // An ASIN missing from Audible's catalog comes back as a
+                    // non-null but EMPTY product stub (no title, no fields).
+                    // Treating that as an answer both returns junk to the
+                    // caller and blocks every later source in the chain.
+                    if (result is AudibleBookResponse shell && string.IsNullOrWhiteSpace(shell.Title))
+                    {
+                        _logger.LogInformation(
+                            "{SourceName} returned an empty product stub for ASIN {Asin}; treating as no answer",
+                            source.Name, asin);
+                        result = null;
+                    }
+
                     if (result != null)
                     {
                         _logger.LogInformation("Successfully fetched metadata from {SourceName} for ASIN: {Asin}", source.Name, asin);


m4bard:
Checked the premise against the live catalog API rather than taking it on trust, because "returns a stub instead of a 404" is the kind of thing that quietly stops being true. It holds. A well-formed ASIN that is not in the catalog answers `HTTP 200` with a product node and almost nothing in it:

```
$ curl "https://api.audible.com/1.0/catalog/products/B000000XYZ?response_groups=product_desc,product_attrs"
{"product":{"asin":"B000000XYZ","asset_details":[],"is_vvab":false},"response_groups":["always-returned"]}
HTTP 200
```

A real ASIN from the same endpoint comes back with the full product. So the shell is exactly as you describe, and `result != null` cannot tell the two apart.

It also settles the choice of discriminator: the stub carries `asin`, `asset_details` and `is_vvab`, and no title. Testing `Title` catches it, and testing the ASIN would not have.

On pinning it down, `AudiobookMetadataServiceTests` already contains the neighbouring case:

```csharp
GetMetadataAsync_UsesAudnexus_WhenAudibleMetadataReturnsNull()
    audibleMock.Setup(...).ReturnsAsync((AudibleBookResponse?)null);
```

That asserts the chain continues when Audible answers null. The case this PR fixes is the same test with the setup returning `new AudibleBookResponse()` instead, and the same assertion should still hold. It is close to a one-line variant of a test that already exists, which is a cheap way to stop the behaviour regressing the next time that loop is touched.

One thought rather than an objection. The loop iterates sources generically, and the new check pattern-matches one concrete response type, so a different provider returning its own empty shell would still count as an answer. Having `AudibleService.GetBookMetadataAsync` return null for a titleless product would keep that knowledge with the provider that has it, and the orchestrator would need no special case at all. Where it sits now is the smaller diff, and both are defensible, so treat it as something to weigh rather than something to change.

The fix itself reads correctly to me, and the logging on the fall-through is the right call: a silent null here would be indistinguishable from the provider simply having nothing.

PR #766, opened 2026-07-14
fix(scan): don't attribute every book by an author to whichever one is scanned

### LINKED ISSUE #765: Library scan attributes every book by an author to whichever one is scanned, and sets the author folder as its BasePath
## Summary

When an audiobook has no `BasePath`, a library scan attributes **every audio file by that author** to it, and then stores the **author folder** as the book's `BasePath`.

Because `BasePath` is what rename and move operate on, a book in this state has an entire author's shelf as its folder.

## Cause

`ScanFileDiscovery.Matches` accepts "the path contains the author" as sufficient to attribute a file to a specific audiobook:

```csharp
var filePathMatchesAuthor = !string.IsNullOrEmpty(authorToken)
    && file.Contains(authorToken, StringComparison.OrdinalIgnoreCase);

return fileNameMatchesTitle || filePathMatchesAuthor || directoryMatchesTitle;
```

In an `{Author}/...` layout — which is most of them — *every* file beneath an author's folder contains that author's name. So the author clause is true for every one of them.

`FindMatchingAudioFiles` groups candidates by directory and takes **all** files in any group where **any** file matches, so each of the author's book folders is claimed whole. `ScanPathPlanner.CalculateBasePath` then reduces the result to its common parent — the author folder.

The author identifies a shelf, not a book.

## How it is reached

`ScanJobProcessor` uses `audiobook.BasePath` as the scan root when it is set. When it is empty, the root falls back to the configured `OutputPath` — the whole library — and every audio file under it becomes a candidate.

A `BasePath` can be empty in normal use: it is only assigned on add when a destination path is supplied, and `ScanJobProcessor` explicitly sets `audiobook.BasePath = null` when the stored folder no longer exists. The next scan of that book then runs against the whole library.

## Reproduction

A conventional layout, three public-domain books by one author:

```
{root}/H. Rider Haggard/H. Rider Haggard - She/She.m4b
{root}/H. Rider Haggard/H. Rider Haggard - Allan Quatermain/Allan Quatermain.m4b
{root}/H. Rider Haggard/H. Rider Haggard - King Solomon's Mines/King Solomon's Mines.m4b
```

Scanning the audiobook **She** (no `BasePath`, so the scan root is the library root):

```
expected: ["She.m4b"]
actual:   ["Allan Quatermain.m4b", "King Solomon's Mines.m4b", "She.m4b"]

BasePath expected: .../H. Rider Haggard - She
BasePath actual:   .../H. Rider Haggard          <-- the author folder
```

`ScanFileDiscovery.FindMatchingAudioFiles` is a static taking a root path and an `Audiobook`, so this reproduces in a unit test against a temp directory — no database or container needed.

## Suggested fix

Only the title can attribute a file to a specific book. Author-in-path should not produce a match on its own.

A path carrying neither the title nor anything else identifying is then simply **not attributable by path**, and is left unmatched rather than attached to an arbitrary book by the same author. That is the right outcome: a miss is recoverable, a wrong link is not — and the user has no reason to go looking for it.

This does mean layouts that encode neither the title nor the author in the path (AudioBookShelf-style series folders with numbered episode files) go from *wrongly matched* to *unmatched*. Those are exactly what an embedded-tag pass should claim, which is what #688 proposes — the two changes complement each other.

## Notes

No existing test depends on author-only matching; the full suite passes unchanged with the author clause removed.

Books cited are public domain with real Audible editions, so the case is reproducible by anyone.

m4bard:
## Summary

Fixes #765.

`ScanFileDiscovery.Matches` accepted "the path contains the author" as sufficient to attribute a file to a *specific* audiobook:

```csharp
var filePathMatchesAuthor = !string.IsNullOrEmpty(authorToken)
    && file.Contains(authorToken, StringComparison.OrdinalIgnoreCase);

return fileNameMatchesTitle || filePathMatchesAuthor || directoryMatchesTitle;
```

In the usual `{Author}/...` layout, *every* file beneath an author's folder contains that author's name — so the clause is true for all of them. `FindMatchingAudioFiles` takes **all** files in any directory group where **any** file matches, so each of the author's book folders is claimed whole, and `ScanPathPlanner.CalculateBasePath` reduces the result to its common parent: the **author folder**. That is then stored as the audiobook's `BasePath` — which is what rename and move operate on.

Reached whenever an audiobook has no `BasePath`, because the scan root then falls back to the configured `OutputPath` and the whole library becomes candidates. A `BasePath` is empty in normal use: it's only assigned on add when a destination path is supplied, and `ScanJobProcessor` sets it to `null` when the stored folder no longer exists.

### Reproduction

Three public-domain books by one author, conventional layout:

```
{root}/H. Rider Haggard/H. Rider Haggard - She/She.m4b
{root}/H. Rider Haggard/H. Rider Haggard - Allan Quatermain/Allan Quatermain.m4b
{root}/H. Rider Haggard/H. Rider Haggard - King Solomon's Mines/King Solomon's Mines.m4b
```

Scanning the audiobook **She**:

```
expected: ["She.m4b"]
actual:   ["Allan Quatermain.m4b", "King Solomon's Mines.m4b", "She.m4b"]

BasePath expected: .../H. Rider Haggard - She
BasePath actual:   .../H. Rider Haggard          <-- the author folder
```

### Changed
- `ScanFileDiscovery.Matches` now attributes a file to an audiobook only when the **title** appears — in the filename or in the containing directory. The author is no longer a match on its own.

### Fixed
- A scan no longer claims every book by an author for whichever one is being scanned.
- An audiobook's `BasePath` is no longer collapsed to its author's folder.

## Testing

`ScanFileDiscoveryReproTests` — 7 new cases against real temp directory trees (`FindMatchingAudioFiles` is a static taking a root path and an `Audiobook`, so this needs no database or container):

- scanning one book does not claim the other books by the same author
- `BasePath` resolves to the book folder, not the author folder
- title in the folder name still matches (the common `{Author}/{Author} - {Series} - {Title}` shape)
- title in the filename still matches when the folder does not carry it
- every file in a matched book folder is still claimed (multi-part books whose part filenames don't repeat the title)
- a sibling book in the same series folder is not claimed
- a path carrying neither title nor anything identifying is left unmatched

`dotnet test`: **1196 passing, 0 failing** (1189 existing + 7 new). No existing test changed — nothing in the suite depended on author-only matching.

## Notes

**The trade-off, stated plainly:** layouts that encode neither the title nor the author in the path — AudioBookShelf-style series folders with numbered episode files — go from *wrongly matched* to *unmatched*.

That is the right direction. A miss is recoverable and visible to the user; a wrong link is neither, because the file is silently attached to a book the user has no reason to go and check. And those layouts are exactly what an embedded-tag pass should claim, which is what #688 proposes — the two changes complement each other rather than overlap.

Books cited are public domain with real Audible editions, so the case is reproducible by anyone.

### DIFF
--- a/listenarr.infrastructure/Library/Scanning/ScanFileDiscovery.cs
+++ b/listenarr.infrastructure/Library/Scanning/ScanFileDiscovery.cs
@@ -33,8 +33,7 @@ internal static class ScanFileDiscovery
         foreach (var group in candidates.GroupBy(file => Path.GetDirectoryName(file) ?? string.Empty))
         {
             var directoryName = Path.GetFileName(group.Key) ?? string.Empty;
-            var groupHasMatch = group.Any(file =>
-                Matches(file, directoryName, titleToken, authorToken));
+            var groupHasMatch = group.Any(file => Matches(file, directoryName, titleToken));
             if (groupHasMatch)
             {
                 foundFiles.AddRange(group.Where(unique.Add));
@@ -42,7 +41,7 @@ internal static class ScanFileDiscovery
             }
 
             foreach (var file in group.Where(file =>
-                         Matches(file, directoryName: string.Empty, titleToken, authorToken)))
+                         Matches(file, directoryName: string.Empty, titleToken)))
             {
                 if (unique.Add(file))
                 {
@@ -103,20 +102,36 @@ internal static class ScanFileDiscovery
         return candidates;
     }
 
+    /// <summary>
+    /// Decides whether a file can be attributed to a specific audiobook from its path alone.
+    /// <para>
+    /// Only the TITLE can do this. The author names a shelf, not a book: in the usual
+    /// <c>{Author}/...</c> layout every file beneath an author's folder contains that author's
+    /// name, so accepting "the path contains the author" as a match attributes every book by
+    /// that author to whichever one is being scanned -- and the resulting common parent is the
+    /// author folder, which then becomes the audiobook's BasePath.
+    /// </para>
+    /// <para>
+    /// A path that carries neither the title nor anything else identifying is simply not
+    /// attributable by path, and is left for the embedded-tag pass to claim. Leaving a file
+    /// unmatched is recoverable; attaching it to the wrong book is not.
+    /// </para>
+    /// </summary>
     private static bool Matches(
         string file,
         string directoryName,
-        string titleToken,
-        string authorToken)
+        string titleToken)
     {
-        var fileNameMatchesTitle = !string.IsNullOrEmpty(titleToken)
-            && Path.GetFileNameWithoutExtension(file)
-                .Contains(titleToken, StringComparison.OrdinalIgnoreCase);
-        var filePathMatchesAuthor = !string.IsNullOrEmpty(authorToken)
-            && file.Contains(authorToken, StringComparison.OrdinalIgnoreCase);
+        if (string.IsNullOrEmpty(titleToken))
+        {
+            return false;
+        }
+
+        var fileNameMatchesTitle = Path.GetFileNameWithoutExtension(file)
+            .Contains(titleToken, StringComparison.OrdinalIgnoreCase);
         var directoryMatchesTitle = !string.IsNullOrEmpty(directoryName)
-            && !string.IsNullOrEmpty(titleToken)
             && directoryName.Contains(titleToken, StringComparison.OrdinalIgnoreCase);
-        return fileNameMatchesTitle || filePathMatchesAuthor || directoryMatchesTitle;
+
+        return fileNameMatchesTitle || directoryMatchesTitle;
     }
 }
--- a/tests/Features/Infrastructure/Library/Scanning/ScanFileDiscoveryReproTests.cs
+++ b/tests/Features/Infrastructure/Library/Scanning/ScanFileDiscoveryReproTests.cs
@@ -1,0 +1,166 @@
+/*
+ * Listenarr - Audiobook Management System
+ * Copyright (C) 2024-2026 Listenarr Contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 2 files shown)


m4bard:
Followed up with a runtime before/after against the published image, since the unit tests above only cover the matcher in isolation.

Setup: a synthetic library generated from public-domain metadata in the common `{Author}/{Author} - {Series} - {Title}` layout, with each book's files multiplied out (20 copies × 40 chapters = 800 files/book) to give the scan something to over-claim. Books are added with `BasePath` cleared — the state where the scan root falls back to the whole library — then scanned. Linked files are compared to a manifest of each file's true owner. Same library, two images: `ghcr.io/listenarrs/listenarr:canary` vs a build of this branch.

Each scanned book has a same-author sibling elsewhere in the library, so the author-in-path clause has something to wrongly grab:

| scan of | `:canary` | this branch |
|---|---|---|
| The Valley of Fear (Doyle) | 1,600 files across **2 books** — claimed a Doyle sibling | 800 files, **1 book** — only its own |
| Stalky and Co. (Kipling) | 1,554 files across **2 books** — claimed a Kipling sibling | 745 files, **1 book** — only its own |

So the over-attribution is present on `canary` and gone on the branch. One honest caveat: on the branch, *Stalky and Co.* links 745 of its own 800 files — a slight under-match on that period-titled book. That's a separate matching-completeness question (title-token matching on an unusual title), not something this change introduced — `canary` over-links it rather than under-linking, so there was no clean baseline to regress from. Worth a separate look, but orthogonal to the over-match this PR fixes.

One thing that surprised me: the stored `BasePath` string ends up as the author folder in *both* runs, so reading `BasePath` alone doesn't reveal the bug — it's only visible in *which files get linked*. That's why this is measured against true ownership rather than by inspecting `BasePath`.

The harness is a small generator + scan-diff script; happy to open-source it if that'd be useful for regression-testing scan behaviour more generally.

m4bard:
Closing this as superseded. The fix and the regression tests landed on #717 (`3509935f`, `b27a91f7`, `499b32e4`), and `ScanFileDiscovery` has since been reworked there well past what this diff does — `Matches` is gone entirely, replaced by title-required matching with the author demoted to corroboration. No reason to keep a PR open against a method that branch removes.

Thanks for taking it on and preserving the attribution. Happy to reopen if #717's direction changes and this becomes useful again.

therobbiedavis:
For the record now that #766 is closed: #717 incorporates and adapts this scan-attribution fix and its regression scenarios into the new ownership/stable-identifier/book-boundary scan architecture. The #717 description credits @m4bard explicitly. Thanks for finding and isolating the author-overmatch failure.

PR #768, opened 2026-07-14
fix(metadata): populate SeriesAsin from the Audnexus series record

### LINKED ISSUE #767: SeriesAsin is never populated: the Audnexus series ASIN is dropped when memberships are built
## Summary

`AudiobookSeriesMembership.SeriesAsin` is never populated from Audnexus. The column exists and the value is available, but the converter doesn't copy it, so the field is `NULL` for every audiobook.

## Cause

`MetadataConverters.ConvertAudnexusToMetadata` builds the memberships from `SeriesPrimary` / `SeriesSecondary`:

```csharp
audnexusSeriesMemberships.Add(new AudiobookSeriesMembership
{
    SeriesName = audnexusData.SeriesPrimary.Name,
    SeriesNumber = audnexusData.SeriesPrimary.Position,
    IsPrimary = true,
    SortOrder = 0
});
```

`AudnexusSeries` carries three fields:

```csharp
public class AudnexusSeries
{
    public string? Asin { get; set; }
    public string? Name { get; set; }
    public string? Position { get; set; }
}
```

`Name` and `Position` are mapped. `Asin` is not — even though `AudiobookSeriesMembership.SeriesAsin` exists for it (added in `AddAudiobookSeriesMemberships`, `maxLength: 64`), and the repository preserves it when it is set.

## Why it matters

The series ASIN is the stable identifier for a series; the series *name* is not.

- A **book** ASIN is per-marketplace and per-narrator. The same public-domain work carries different ASINs in different catalogues — Grimm's *Kinder- und Hausmärchen* is `B00B4FPO6A` in the German catalogue and `B00TPKFANI` in the US one, and each returns 404 from the other region. Re-releases mint new book ASINs too.
- A series **name** is free text and varies between editions and translations.
- The series ASIN is stable across both.

So `(SeriesAsin, SeriesNumber)` is the durable way to say "this is book N of that series", and it is currently unavailable because the ASIN is dropped at the point of conversion.

## Reproduction

H. Rider Haggard's *She and Allan* (`B00CQ5WAXW`) belongs to **two** series at once. Live from `api.audnex.us`:

```json
"seriesPrimary":   { "asin": "B01E633FQM", "name": "Ayesha",           "position": "0" },
"seriesSecondary": { "asin": "B01F5TL5K4", "name": "Allan Quatermain", "position": "7" }
```

After conversion, both memberships are created correctly with their names and positions — that part works — but `SeriesAsin` is `null` on both.

## Expected

`SeriesAsin` is populated from `AudnexusSeries.Asin` for both the primary and secondary membership.

## Notes

This looks like a small mapping omission rather than a design gap: the column, the DTO field, the dual-membership handling and the persistence merge are all already in place. Happy to put up a PR.

The book is public domain with a real Audible edition, so the case is reproducible by anyone.

m4bard:
## Summary

Fixes #767.

`AudnexusSeries` carries three fields:

```csharp
public class AudnexusSeries
{
    public string? Asin { get; set; }
    public string? Name { get; set; }
    public string? Position { get; set; }
}
```

`ConvertAudnexusToMetadata` mapped `Name` and `Position` but dropped `Asin`, so `AudiobookSeriesMembership.SeriesAsin` was `null` for every audiobook — even though the column exists for it (`AddAudiobookSeriesMemberships`, `maxLength: 64`) and the repository preserves it when it is set.

**Why the series ASIN matters:** it is the stable identifier for a series, and neither of the alternatives is.

- A **book** ASIN is per-marketplace and per-narrator. Grimm's *Kinder- und Hausmärchen* is `B00B4FPO6A` in the German catalogue and `B00TPKFANI` in the US one, and each returns 404 from the other region. Re-releases mint new book ASINs as well.
- A series **name** is free text and varies between editions and translations.
- The series ASIN is stable across both.

So `(SeriesAsin, SeriesNumber)` is the durable way to express "book N of that series", and it was unavailable because the ASIN was dropped at conversion.

### Fixed
- `MetadataConverters.ConvertAudnexusToMetadata` now copies `AudnexusSeries.Asin` into `AudiobookSeriesMembership.SeriesAsin`, for both the primary and the secondary membership.

## Testing

`MetadataConvertersSeriesAsinTests` — 3 new cases, built from the live Audnexus record for H. Rider Haggard's *She and Allan* (`B00CQ5WAXW`), which belongs to **two** series at once:

```json
"seriesPrimary":   { "asin": "B01E633FQM", "name": "Ayesha",           "position": "0" },
"seriesSecondary": { "asin": "B01F5TL5K4", "name": "Allan Quatermain", "position": "7" }
```

- the primary membership carries its series ASIN
- the secondary membership carries its series ASIN
- a book in two series keeps both memberships (this already worked; pinned so it stays working)

`dotnet test`: **1192 passing, 0 failing** (1189 existing + 3 new). No existing test changed.

## Notes

Two lines. The column, the DTO field, the dual-membership handling and the persistence merge were all already in place — only the mapping was missing.

The book cited is public domain with a real Audible edition, so the case is reproducible by anyone.

### DIFF
--- a/listenarr.application/Search/Metadata/MetadataConverters.cs
+++ b/listenarr.application/Search/Metadata/MetadataConverters.cs
@@ -143,6 +143,7 @@ public class MetadataConverters
         {
             audnexusSeriesMemberships.Add(new AudiobookSeriesMembership
             {
+                SeriesAsin = audnexusData.SeriesPrimary.Asin,
                 SeriesName = audnexusData.SeriesPrimary.Name,
                 SeriesNumber = audnexusData.SeriesPrimary.Position,
                 IsPrimary = true,
@@ -154,6 +155,7 @@ public class MetadataConverters
         {
             audnexusSeriesMemberships.Add(new AudiobookSeriesMembership
             {
+                SeriesAsin = audnexusData.SeriesSecondary.Asin,
                 SeriesName = audnexusData.SeriesSecondary.Name,
                 SeriesNumber = audnexusData.SeriesSecondary.Position,
                 IsPrimary = audnexusSeriesMemberships.Count == 0,
--- a/tests/Features/Application/Search/Metadata/MetadataConvertersSeriesAsinTests.cs
+++ b/tests/Features/Application/Search/Metadata/MetadataConvertersSeriesAsinTests.cs
@@ -1,0 +1,95 @@
+/*
+ * Listenarr - Audiobook Management System
+ * Copyright (C) 2024-2026 Listenarr Contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+using Listenarr.Tests.Common;
+using Microsoft.Extensions.Logging.Abstractions;
+
+namespace Listenarr.Tests.Features.Application.Search.Metadata
+{
+    /// <summary>
+    /// Audnexus returns an ASIN for each series a book belongs to, and
+    /// AudiobookSeriesMembership has a SeriesAsin column to hold it. The converter maps the
+    /// series name and position but drops the ASIN, so SeriesAsin is never populated.
+    ///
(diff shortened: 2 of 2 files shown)


m4bard:
Pushed a second commit, 618aff19, because the converter was only half of #767, and rebased the branch onto current canary while at it, since the lookup's return type changed underneath it. The same Audnexus series data is also dropped one step earlier: when the product lookup is answered by Audnexus, `AudiobookMetadataService` copies the response into the Audible-shaped one without its series list (`AudiobookMetadataService.cs:90-106` at canary `a630572e`), so the add modal and the metadata rescan, which both go through that lookup, see no series at all and fall back to the legacy name. The commit adds one small mapper that turns the Audnexus primary and secondary series into the Audible-shaped list with identifiers, uses it in that copy, and has the converter feed the same list through `BuildSeriesMemberships` rather than its own hand-written walk, so the mapping lives in one place. The three tests already here are unchanged; two are added for the lookup. One consequence worth stating, which I have not run yet: I think a metadata rescan on this build would repair existing rows when Audnexus answers, which it could not before.

PR #772, opened 2026-07-20
fix(qbittorrent): add URL Base support for reverse-proxied instances

### LINKED ISSUE #690: qBittorrent download client missing URL Base field
Sonarr and Radarr both expose a URL Base field on the qBittorrent download client configuration, which is required when qBittorrent is served behind a reverse proxy at a path prefix (e.g. https://{seedbox_host}/qbittorrent).

Listenarr's qBittorrent config only has Host and Port, so it always connects to the root path. This makes it incompatible with any reverse-proxied qBittorrent setup. The Transmission client already has this field. qBittorrent needs the same.

I can't find a workaround currently to connect to my client. I'm eager to try this out!

schmitzkr:
## Summary

Adds a "URL Base" field to the qBittorrent download client (closes #690), so Listenarr can reach qBittorrent instances served behind a path-prefixed reverse proxy (e.g. `example.com/qbittorrent`) — matching the field Transmission's client already has.

I deliberately didn't just copy Transmission's mechanism. Transmission's `urlBase` *replaces* its whole RPC path, because that mirrors Transmission's own daemon setting (`--rpc-url-base`) — Listenarr just matches whatever arbitrary path the user configured server-side. qBittorrent has no equivalent: I checked upstream, and a native WebUI base-path setting has been proposed twice (qbittorrent/qBittorrent#21471, closed unmerged; #23467, still open/unmerged) — neither has shipped. So every real qBittorrent-behind-a-proxy deployment today works by the *reverse proxy* stripping a path prefix before it reaches qBittorrent, which always serves at its own root regardless. That means the correct semantics here are a prefix prepended before the fixed `/api/v2/...` routes, not a full-path override.

### Added
- `QBittorrentHelpers.BuildBaseUrl()` — composes authority + normalized `urlBase` prefix from client settings; no-op when unset, so existing configs are unaffected
- "URL Base" field now renders for qBittorrent in the download client form, with qBittorrent-specific placeholder/help text (distinct from Transmission's field, since the semantics differ)
- Backend test asserting the actual HTTP request hits `/qbittorrent/api/v2/app/version` when configured
- Frontend test asserting the field renders for qBittorrent and flows into the save/test payload

### Changed
- The 8 qBittorrent workflow files that each independently called `DownloadClientUriBuilder.BuildAuthority(client)` now go through the new helper instead (mechanical, one line each — no other behavior change)

## Testing
- `dotnet build` / `dotnet format --verify-no-changes` clean
- Targeted backend tests (qBittorrent + Transmission): 71/71 pass
- Full backend suite: 1185/1190 pass (5 pre-existing failures unrelated to this change — diacritics-normalization tests that only fail in an ICU-less sandbox, not a normal dev setup)
- Frontend: Prettier / `vue-tsc` clean, full suite 393/393 pass
- **Manually tested against a real reverse-proxied qBittorrent instance** and confirmed the connection works with the new URL Base field set

## Notes

**AI assistance disclosure:** this PR was developed with Claude Code's assistance - the implementation (the helper, the 8-file sweep, the frontend field, and the tests) was written by Claude Code. My role was directing the approach and validating the result: I pushed back on an initial draft that assumed we should just copy Transmission's urlBase mechanism, which prompted the investigation that found qBittorrent needs different (prefix, not replace) semantics - that upstream research and the resulting design decision are reflected above. I also manually tested this against my own qBittorrent instance behind a real reverse proxy and confirmed it resolves the actual problem reported in #690, which is real, live validation beyond the automated test suite. I have not done a line-by-line manual read of the full diff myself - flagging that plainly since this is my first PR on a project I don't maintain, and I'd rather be upfront about the split between direction/testing and code-writing than imply a level of manual review that didn't happen.

One unrelated thing noticed while auditing every place qBittorrent builds a URL: there are 5 separate, independently-written login implementations across these files (pre-existing, not introduced by this change), and only one of them sets a `Referrer` header that qBittorrent's CSRF check can care about. It shouldn't interact with `urlBase` (qBittorrent's CSRF/Referer check is origin-based, not path-based), so I left it alone here, but it seemed worth mentioning in case it's the cause of a future report like "Test Connection works but adding a torrent fails."

### DIFF
--- a/fe/src/__tests__/DownloadClientFormModal.spec.ts
+++ b/fe/src/__tests__/DownloadClientFormModal.spec.ts
@@ -172,4 +172,47 @@ describe('DownloadClientFormModal', () => {
     expect(calledWith.password).toBe('')
     expect(calledWith.id).toBe('4')
   })
+
+  it('renders URL Base field for qbittorrent and includes it in the test payload when set', async () => {
+    const api = await import('@/services/api')
+    ;(api.testDownloadClient as unknown) = vi.fn(async (config: unknown) => ({
+      success: true,
+      message: 'ok',
+      client: config,
+    }))
+
+    const wrapper = mount(DownloadClientFormModal, {
+      global: { plugins: [createPinia()] },
+      props: { visible: true, editingClient: null },
+    })
+
+    await wrapper.setProps({
+      editingClient: {
+        id: '5',
+        name: 'qbt',
+        type: 'qbittorrent',
+        host: 'qbittorrent.local',
+        port: 8080,
+        isEnabled: true,
+        useSSL: false,
+        downloadPath: '',
+        username: '',
+        password: '',
+        settings: {},
+      },
+    })
+    await wrapper.vm.$nextTick()
+
+    const urlBaseInput = wrapper.find('input[id="urlBase"]')
+    expect(urlBaseInput.exists()).toBe(true)
+
+    await urlBaseInput.setValue('/qbittorrent')
+
+    const testButton = wrapper.find('button.btn-info')
+    await testButton.trigger('click')
+
+    expect(api.testDownloadClient as unknown).toHaveBeenCalled()
+    const calledWith = (api.testDownloadClient as unknown).mock.calls[0][0]
+    expect(calledWith.settings.urlBase).toBe('/qbittorrent')
+  })
 })
--- a/fe/src/components/domain/download/DownloadClientFormModal.vue
+++ b/fe/src/components/domain/download/DownloadClientFormModal.vue
@@ -116,15 +116,22 @@
               </Checkbox>
             </div>
 
-            <div class="form-group" v-if="formData.type === 'transmission'">
+            <div class="form-group" v-if="showUrlBase">
               <label for="urlBase">URL Base</label>
               <input
                 id="urlBase"
                 v-model="formData.urlBase"
                 type="text"
-                placeholder="/transmission/rpc"
+                :placeholder="getUrlBasePlaceholder()"
               />
-              <small
+              <small v-if="formData.type === 'qbittorrent'"
+                >Path prefix for qBittorrent instances behind a reverse proxy (e.g.
+                <code>/qbittorrent</code>). Must match the prefix your reverse proxy strips before
+                forwarding to qBittorrent's own root-relative API - qBittorrent itself has no
+                built-in base path setting. Leave blank if qBittorrent is reachable directly at the
+                host/port above.</small
+              >
+              <small v-else
                 >RPC path for the Transmission endpoint. Default is <code>/transmission/rpc</code>.
                 Some seedbox providers use a custom path (e.g. <code>/rpc</code>).</small
               >
@@ -496,6 +503,14 @@ const getPortHelpText = () => {
   return hints[formData.value.type] || 'Port the download client is listening on.'
 }
 
+const showUrlBase = computed(() => {
+  return formData.value.type === 'transmission' || formData.value.type === 'qbittorrent'
+})
+
+const getUrlBasePlaceholder = () => {
+  return formData.value.type === 'qbittorrent' ? '/qbittorrent' : '/transmission/rpc'
+}
+
 const getCategoryHelp = () => {
   if (isUsenet.value) {
     return 'Adding a category specific to Listenarr avoids conflicts with unrelated non-Listenarr downloads. Using a category is optional, but strongly recommended.'
@@ -593,9 +608,7 @@ const testConnection = async () => {
         ...(formData.value.type === 'sabnzbd' && formData.value.apiKey
           ? { apiKey: formData.value.apiKey }
           : {}),
-        ...(formData.value.type === 'transmission' && formData.value.urlBase
-          ? { urlBase: formData.value.urlBase }
-          : {}),
+        ...(showUrlBase.value && formData.value.urlBase ? { urlBase: formData.value.urlBase } : {}),
         ...(formData.value.category && { category: formData.value.category }),
         ...(formData.value.tags && { tags: formData.value.tags }),
         recentPriority: formData.value.recentPriority,
@@ -653,9 +666,7 @@ const handleSubmit = async () => {
         ...(formData.value.type === 'sabnzbd' && formData.value.apiKey
           ? { apiKey: formData.value.apiKey }
           : {}),
-        ...(formData.value.type === 'transmission' && formData.value.urlBase
-          ? { urlBase: formData.value.urlBase }
-          : {}),
+        ...(showUrlBase.value && formData.value.urlBase ? { urlBase: formData.value.urlBase } : {}),
         ...(formData.value.category && { category: formData.value.category }),
         ...(formData.value.tags && { tags: formData.value.tags }),
         recentPriority: formData.value.recentPriority,
--- a/listenarr.infrastructure/DownloadClients/Qbittorrent/QBittorrentHelpers.cs
+++ b/listenarr.infrastructure/DownloadClients/Qbittorrent/QBittorrentHelpers.cs
@@ -69,5 +69,37 @@ namespace Listenarr.Infrastructure.DownloadClients.Qbittorrent
                 logger.LogInformation("Fetching qBittorrent queue filtered by category: {Category}", category);
             }
         }
+
+        /// <summary>
+        /// Builds the base URL used for every qBittorrent WebAPI call, including an optional
+        /// "urlBase" path prefix from client settings (e.g. "/qbittorrent") for instances that
+        /// sit behind a reverse proxy at a path prefix.
+        /// Unlike Transmission's urlBase (which replaces the whole RPC path to match Transmission's
+        /// own configurable --rpc-url-base setting), qBittorrent has no equivalent server-side base
+        /// path setting, so this is a plain prefix prepended before the fixed "/api/v2/..." routes -
+        /// it must match whatever prefix the reverse proxy strips before forwarding to qBittorrent.
+        /// </summary>
+        /// <param name="client">The download client configuration providing host/port and settings.</param>
+        /// <returns>The authority (scheme://host:port) with the normalized urlBase prefix appended, if configured.</returns>
+        public static string BuildBaseUrl(DownloadClientConfiguration client)
+        {
+            var authority = DownloadClientUriBuilder.BuildAuthority(client);
+            var prefix = ResolveUrlBasePrefix(client);
+            return prefix.Length == 0 ? authority : authority + prefix;
+        }
+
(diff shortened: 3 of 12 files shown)


schmitzkr:
I have actually stood up a Bindery instance to be able to have my downloads handled. But if this ever gets merged I think I might come back to Listenarr.

m4bard:
I came to #690 meaning to implement it, found this, and read it instead. It is the better design and I would rather say why than duplicate it.

Everything below is from checking out the branch and running it, not from reading the diff.

## The prefix-versus-replace decision is right, and it is checkable

You reasoned that qBittorrent has no server-side base path, so the prefix must be prepended before the fixed `/api/v2/...` routes rather than replacing them. qBittorrent's own source agrees, and the specific place is `src/webui/webapplication.cpp`: `isSameOrigin` compares port and host only, with the scheme comparison commented out and the path not considered at all. So a prefixed base cannot upset anything qBittorrent checks. That seems worth having in the record, since it is the one part of this that is hard to convince yourself of from the Listenarr side alone.

## What I verified

The 8-file sweep is genuinely mechanical. Exactly eight one-line substitutions, and diffing the changed hunks shows no other line altered in those files. Afterwards, the only remaining `DownloadClientUriBuilder.BuildAuthority` call anywhere in the qBittorrent tree is the one inside `BuildBaseUrl` itself, which is what you want. Nothing was missed.

`TestConnection_WithUrlBase_PrefixesApiPath` is a real control, not just a passing test. I made `BuildBaseUrl` return the bare authority and it fails; restored, it passes. Worth stating because a test that cannot fail looks exactly like one that passes.

Full unfiltered backend suite on your branch: 1190 passed, 0 failed. Your five diacritics failures do not reproduce here, which supports your reading that they are an ICU-less sandbox artefact rather than anything you touched.

It still merges cleanly into current canary, and it does not conflict with #906, which touches `QbittorrentAddWorkflow.cs` as well.

## I think your Referrer note is the right observation pointing the wrong way

You flagged that only one of the login implementations sets a `Referrer`, and guessed the future report would be "Test Connection works but adding a torrent fails". Reading `webapplication.cpp`, I think it is the other way round, and the reason is a bit reassuring.

When neither `Origin` nor `Referer` is present, qBittorrent falls back to `Sec-Fetch-Site`, and when that is absent too it returns early with the comment "owasp.org recommends to block this request, but doing so will inevitably lead Web API users to spoof headers so let's be permissive here (Web API clients don't send 'Sec-Fetch-Site' at all)". So the logins that send no `Referrer` are permitted deliberately rather than by luck.

`QbittorrentConnectionTester.cs:95` is the only one that sets `Referrer`, and setting it opts into the strict comparison the others never reach. If an origin mismatch ever bites, Test Connection is the call that can fail while the torrent add keeps working. I counted six login call sites rather than five, since the connection tester has two.

None of that is a problem with this PR. Your change moves that header from the bare authority to the prefixed one, and since only host and port are compared, it still passes.

## One thing I would fix, and one I probably would not

A full URL in the field is accepted and silently mangled. I ran the normalization over a few inputs:

```
/qbittorrent                              -> http://host:8080/qbittorrent
qbittorrent                               -> http://host:8080/qbittorrent
/qbittorrent/                             -> http://host:8080/qbittorrent
  /qbittorrent                            -> http://host:8080/qbittorrent
https://seedbox.example.com/qbittorrent   -> http://host:8080/https://seedbox.example.com/qbittorrent
```

The first four are exactly right. The last one worries me because it is the likely paste: #690 describes the setup as `https://{seedbox_host}/qbittorrent` and your own comment describes yours as `https://example.com/qbittorrent/`, so someone reading either is being shown a full URL and asked for a path. They get a config that fails with no indication why.

Sonarr, Radarr and Readarr reject anything beginning with a scheme in their `ValidUrlBase` rule. Would you want the same here? A guard in `ResolveUrlBasePrefix`, or a warning on the field, either would do. I am asking rather than suggesting a patch because it is your PR and you have the real proxied instance to try it against.

The other one I noticed is that `//qbittorrent` produces a double slash. That seems genuinely not worth code.

## On the disclosure

For what it is worth, the line-by-line read you said you had not done is the thing I have now done, and the code holds up. Flagging the split between direction and code-writing was the right call and it made this easy to review well.

Checked against canary `a630572e`, and against qBittorrent `master` for the WebUI behaviour.

schmitzkr:
Thanks for this — genuinely one of the more thorough reviews I've had on this repo. The `isSameOrigin` citation, the mutation-test on `TestConnection_WithUrlBase_PrefixesApiPath`, and catching that the Referrer risk points the opposite direction from what I guessed are all things I hadn't verified myself, and I checked each of them against the branch afterward — all correct as written.

I've pushed a fix for the one open item: a value like `https://seedbox.example.com/qbittorrent` pasted into URL Base is now rejected instead of being silently concatenated into a broken URL. `ResolveUrlBasePrefix` now rejects anything that parses as an absolute URI (`Uri.TryCreate(trimmed, UriKind.Absolute, ...)`) and throws a `QbittorrentException` telling the user to strip the scheme/host and enter just the path. `QbittorrentConnectionTester` now catches that specifically so the message actually reaches the UI via Test Connection, instead of getting swallowed into a generic "Connection failed." Added `TestConnection_WithFullUrlAsUrlBase_ReturnsClearError` alongside the existing urlBase test to cover it.

In the same spirit of disclosure you raised: this fix, and the verification of your review's claims against the branch beforehand, was implemented by Claude Sonnet 5 via Claude Code, at my direction — I reviewed the diff and pushed it myself.

m4bard:
Thanks for checking those against the branch rather than taking them on trust. That is the part that makes a review worth writing.

The rejection does what you intended for a pasted full URL. It also rejects the value the field is documented to take, but only on Linux, so I would not expect it to have shown up for you.

## What I ran

Checked out 82ced00c and ran the suite. `TestConnection_WithUrlBase_PrefixesApiPath` fails at `QbittorrentAdapterTests.cs:119`, on `Assert.True(success)`, with `urlBase` set to `/qbittorrent`. Full unfiltered suite: 1190 passed, 1 failed. Your new `TestConnection_WithFullUrlAsUrlBase_ReturnsClearError` passes.

## Why

`Uri.TryCreate(trimmed, UriKind.Absolute, out _)` treats a leading-slash path as an absolute `file:` URI on Unix. I ran it directly rather than reasoning about it:

```
Uri.TryCreate("/qbittorrent", Absolute)                       = True   scheme=file
Uri.TryCreate("qbittorrent",  Absolute)                       = False
Uri.TryCreate("https://seedbox.example.com/qbittorrent", ...) = True   scheme=https
```

So `/qbittorrent`, which is the placeholder in your own help text, now throws with the message asking the user to remove a scheme and host that are not there. `qbittorrent` without the leading slash still works, which will make it look intermittent to whoever hits it.

On Windows `/qbittorrent` does not parse as absolute, which I assume is why the suite looked green when you pushed. Listenarr ships as a container, so the deployed path is the one that fails.

## Options, and I do not think it matters much which

Checking the scheme after parsing keeps the shape you have:

```csharp
if (Uri.TryCreate(trimmed, UriKind.Absolute, out var parsed)
    && (parsed.Scheme == Uri.UriSchemeHttp || parsed.Scheme == Uri.UriSchemeHttps))
```

A plain `trimmed.Contains("://")` would also do it and reads more obviously, at the cost of accepting something like `httpx://host` that nobody is going to type.

Either way it seems worth a case in the test for `/qbittorrent` surviving, since that is the one the current guard eats.

## On the disclosure

Appreciated, and mine is the same: the review, the runs above and this comment were done with Claude Code at my direction. I read the diff and the failure myself before writing this.

schmitzkr:
Confirmed — you're right, and thanks for actually running it instead of trusting the reasoning (mine or otherwise). `Uri.TryCreate(trimmed, UriKind.Absolute, out _)` matching a leading-slash path as `file:` on Unix is not something I'd have caught by inspection.

Pushed a fix: swapped that check for `DownloadClientUriBuilder.TryParseHttpOrHttpsAbsoluteUri`, which is already used elsewhere in this codebase for exactly this "reject anything that isn't a genuine http/https absolute URI" case, so `/qbittorrent` no longer gets misread as an absolute URI at all while `https://seedbox.example.com/qbittorrent` still gets rejected correctly. I went with reusing that helper over either of your two suggested shapes since it keeps the scheme-checking logic in one place rather than duplicating it a third time in this file.

This time I actually ran the suite before pushing rather than reasoning about it from static analysis (which is exactly the gap that let this ship in the first place) — full `QbittorrentAdapterTests` run, 35/35 passed, including `TestConnection_WithUrlBase_PrefixesApiPath` and `TestConnection_WithFullUrlAsUrlBase_ReturnsClearError` both green. Didn't add a further isolated case for `/qbittorrent` surviving specifically, since `TestConnection_WithUrlBase_PrefixesApiPath` already asserts exactly that (asserts success with `urlBase = "/qbittorrent"`) — happy to add a narrower one if you think the coverage is too indirect.

One more disclosure note, since we're both doing this openly: my commits now carry a `Co-Authored-By: Claude Sonnet 5` trailer going forward, for the same reason as the rest of this — starting with the one above. The earlier one (82ced00c) predates the convention; I'm leaving it as-is rather than rewriting history that's already referenced by SHA in this thread.

m4bard:
Confirmed on Linux: `bebed536` builds clean and the qBittorrent tests are 48 of 48, including both `TestConnection_WithUrlBase_PrefixesApiPath` and `TestConnection_WithFullUrlAsUrlBase_ReturnsClearError`. Reusing `DownloadClientUriBuilder.TryParseHttpOrHttpsAbsoluteUri` is a better answer than either shape I suggested, for the reason you gave: the scheme check belongs in one place rather than a third copy.

## On whether the coverage is too indirect

I do not think a separate test earns its keep. `TestConnection_WithUrlBase_PrefixesApiPath` sets `urlBase` to `/qbittorrent` and asserts success, so if the guard ever rejects a plain path again, that test goes red. The protection is real, not incidental.

The one thing I would consider is the name. It says the test is about prefixing the API path, which was true when you wrote it and is now half of what it covers. Whoever sees it fail next will read the name, look at the assertion on `uri.AbsolutePath`, and spend a minute working out why a URL Base guard broke a test about path prefixing.

A sentence in the test body naming the second thing it now protects would fix that, and costs less than a second test that would duplicate the setup to assert something the first one already asserts. Your call, and it is a small thing either way.

## On the trailer

Sensible, and leaving `82ced00c` alone is the right instinct: it is referenced by SHA in this thread and rewriting it would break the citations. I have taken the same position on my own earlier comments, which predate my adopting a disclosure line and do not carry one.

Worth saying that the thing which made this catchable was you stating plainly that you had reasoned from static analysis rather than run it. That is what told me where to look.

Worked through with Claude Code at my direction. The 48 of 48 above is a run on this machine against your branch head, not a reading of the diff. I reviewed this before posting.

PR #774, opened 2026-07-20
chore: isolate PR 717 scan attribution diagnostics

therobbiedavis:
Temporary diagnostic PR to run only `ScanFileDiscoveryRegressionTests` on Linux and Windows. This is not intended for merge and will be closed after the isolated result is captured.


therobbiedavis:
Diagnostic result:

- The isolated `ScanFileDiscoveryRegressionTests` suite passed on both Ubuntu and Windows.
- The full Linux backend TRX contained one failure: `FileMoverFallbackTests.MoveDirectoryAsync_SymbolicLinkAlias_BlocksCopyDeleteFallback`, at `tests/Features/Api/Services/FileMoverFallbackTests.cs:167` (`Assert.False()` expected false, actual true).
- That failure is outside the scan-discovery implementation and regression tests changed for #717/#766, so no additional scan production change was made.

This temporary PR remains closed and will not be merged.

Issue #801, opened 2026-08-06
Feature Request: Add option to omit author from automated search queries (Title-only search)


schinguhl:
**Is your feature request related to a problem? Please describe.**
When performing automated searches via indexers (especially Usenet / Newznab providers), combined queries formatted as Title + Author often yield 0 results. Indexers frequently require exact title matches or formatted strings like Author - Title, causing standard Title Author queries to fail completely. However, searching strictly by Title yields successful hits every time.

**Describe the solution you'd like**
Add a setting to search using Title only by default for automated searches, completely omitting the author's name from the indexer query string.

**Describe alternatives you've considered**
- Fallback logic: Automatically retry with Title only if the initial Title + Author search returns zero results.
- Manual editing: Manually removing the author name from the search bar (only works for one-off manual searches, not automated backlog/monitored grabs).

Context: This issue is particularly prevalent when searching for German audiobooks on Usenet indexers.

m4bard:
Half of this is already written, in open PR #714, though it is the half you listed as an alternative rather than the one you asked for.

#714 adds `AutomaticSearchTitleOnlyFallback`, defaulting to true. In `AutomaticSearchService`, when the full `Title Author Series` query comes back with nothing, it retries once with a title-only query built by `AutomaticSearchResultClassifier.BuildTitleOnlySearchQuery`. The retry only fires on zero results. A code comment there notes it doubles indexer queries for books that are not found, which is why it is opt-out on the background sweep rather than always on.

For your case, where Title plus Author returns nothing and Title alone hits every time, that fallback would find the release. It would also spend a wasted query first, on every book, every sweep.

What #714 does not add is a setting to omit the author from automated queries in the first place, which is what you actually asked for. Nothing in it lets you skip straight to title-only.

#714 is still open and not merged, so none of this is in canary yet.

I am not involved in that PR. I noticed the overlap while reading through open work, and it seemed worth saying so you know which half of the request is already spoken for.

PR #824, opened 2026-08-19
Retry automatic search with title only when title and author returns nothing

matheusfillipe:
Automatic search builds one query from the title plus the author, and gives up if it comes back empty. Audible titles often carry a suffix like `(Part 1 of 2)` or `(Dramatized Adaptation)`, and the author usually does not appear in release names, so that combined query matches nothing at the indexers.

Refs #801.


matheusfillipe:
Wrong base branch, reopened against canary as #825.

Issue #846, opened 2026-08-20
Library Import keeps only the first series of a multi-series book


m4bard:
On canary `03958c15` (1.3.2), adding a book through Library Import stores one series membership even when the metadata carries several. Adding the same book through Add Library keeps them all. Nothing warns that anything was dropped.

The backend is not the constraint here, which is what makes this worth reporting as a bug rather than a missing feature.

## The mechanism

`/library/add` calls `AudiobookSeriesMembershipHelper.ApplyToAudiobook(audiobook, metadata.SeriesMemberships, metadata.Series, ...)`, and `Normalize` only synthesizes a membership from the legacy scalar `Series` when no memberships arrive.

`matchToMetadata` in `fe/src/stores/libraryImport.ts` collapses the search result's series array to its first entry:

```ts
const seriesItem = Array.isArray(seriesRaw)
  ? (seriesRaw as Array<{ name?: string; asin?: string; position?: string }>)[0]
  : null
```

and never sets `seriesMemberships`, even though `AudibleBookMetadata` already has that field. So the request arrives carrying scalars only, the backend takes the legacy path, and exactly one membership is stored.

Multi-series is representable the whole way through. `AudiobookSeriesMembership` is an ordered child table with name, number, asin, isPrimary and sortOrder. `MetadataConverters.BuildSeriesMemberships` builds one per Audible entry, the Audnexus converter builds primary and secondary, the read DTOs return the list, `formatSeriesMemberships` in `fe/src/utils/seriesUtils.ts` already renders all of them, and `AddLibraryModal.vue` normalizes and sends them. Library Import is the one write path that reduces the list.

A real example, so this is not hypothetical: `B00CQ5WAXW` returns `seriesPrimary` `B01E633FQM` "Ayesha" at position 0 and `seriesSecondary` `B01F5TL5K4` "Allan Quatermain" at position 7 from `api.audnex.us`. Added through Add Library it keeps both; added through Library Import it keeps "Ayesha" only.

## One thing to be careful of in any fix, which cost me a rewrite

`SearchResponseMapper` has a fallback branch, used when the ASIN re-fetch fails, that synthesizes a single series entry and sets its `asin` to the series **name**:

```csharp
if (!string.IsNullOrWhiteSpace(md?.Series))
    fallbackSeries.Add(new { asin = md.Series, name = md.Series, ... });
```

[...]

PR #847, opened 2026-08-20
fix(library-import): keep every series membership, not just the first

m4bard:
## Summary

Library Import collapses a multi-series match to its first series and never populates `seriesMemberships`, so the backend falls back to the legacy scalar path and stores one membership. This builds the membership list in the store instead.

Full write-up and the hazard I had to guard against are in #846.

### Fixed
- `matchToMetadata` in `fe/src/stores/libraryImport.ts` maps every entry of the search result's series array into an `AudiobookSeriesMembership`, with `isPrimary` and `sortOrder` following the incoming order, and sets `seriesMemberships` on the metadata when there is at least one.

The scalar `series`, `seriesNumber` and `seriesAsin` fields stay populated from the first entry, so nothing reading them changes behaviour. `AudiobookSeriesMembershipHelper.Normalize` only synthesizes from the scalar when no memberships arrive, so sending the list is enough to take the other path.

### Guarded
- A membership's `seriesAsin` is kept only when it matches `/^[A-Z0-9]{10}$/i`.

This is not cosmetic. `SearchResponseMapper`'s fallback branch, used when the ASIN re-fetch fails, sets a synthesized entry's `asin` to the series **name**. That is harmless today because the value has nowhere to land and is dropped on the way in. Sending memberships gives it somewhere to land, and it would be persisted into `AudiobookSeriesMembership.SeriesAsin`, the column #767 is working to make trustworthy. Making the mapper emit null would be the better fix and belongs in its own change; this keeps the bad value out in the meantime.

The second flattening at the add call site is deliberately untouched. It adapts to `SearchResult.Series`, which is `string?` on the backend, so it is a single-series contract rather than a second instance of the bug.

## Testing

Two cases added to `fe/src/__tests__/libraryImport.store.spec.ts`: that every membership of a multi-series match reaches the add request, and that a series asin which is really the series name is dropped.

Verified as real guards in both directions. With only the source change reverted and both tests kept, both fail and the other seven in the file pass. Restored, all nine pass.

Full frontend suite: 597 tests across 89 files, all passing. `vue-tsc` type-check clean, prettier and eslint clean on both changed files.

## Notes

No container reproduction for this one, and I would rather say so than imply parity with the other reports. This is browser-side logic on the path between a search result and the add request; no scan, import action or database state exercises it, so a harness check here would not be testing the thing that is broken.

The multi-series case is real rather than constructed: `B00CQ5WAXW` returns `seriesPrimary` "Ayesha" and `seriesSecondary` "Allan Quatermain" from `api.audnex.us`, verified live rather than taken from a quote.

### DIFF
--- a/fe/src/__tests__/libraryImport.store.spec.ts
+++ b/fe/src/__tests__/libraryImport.store.spec.ts
@@ -417,4 +417,103 @@ describe('library import store', () => {
       'Jack of Shadows',
     )
   })
+
+  it('carries every series membership of a multi-series match into the add request', async () => {
+    const { useLibraryImportStore } = await import('@/stores/libraryImport')
+    const store = useLibraryImportStore()
+
+    // A book can legitimately belong to more than one series. Audnexus returns these as
+    // seriesPrimary/seriesSecondary and the backend builds one membership per entry, so the
+    // import path has to send both rather than keeping only the first.
+    store.items = {
+      '/incoming/Two Series/Book.m4b': {
+        id: '/incoming/Two Series/Book.m4b',
+        fullPath: '/incoming/Two Series/Book.m4b',
+        sourceFiles: ['/incoming/Two Series/Book.m4b'],
+        folderPath: '/incoming/Two Series',
+        relativePath: 'Two Series',
+        folderName: 'Two Series',
+        format: 'M4B',
+        fileCount: 1,
+        selectedMatch: {
+          title: 'Two Series Book',
+          authors: [{ name: 'Author' }],
+          series: [
+            { asin: 'B01E633FQM', name: 'First Series', position: '0' },
+            { asin: 'B01F5TL5K4', name: 'Second Series', position: '7' },
+          ],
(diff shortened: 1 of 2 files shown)


m4bard:
Pushed a second commit, 10377e76, on the same defect from the other end, and rebased the branch onto current canary. The first commit built the membership list from the search match, but the wizard's enrichment step, which re-fetches the product before posting, lifted only authors and narrators and discarded the series list it had just fetched, so an enriched item still posted whatever the match knew and nothing more. The commit moves the membership building into one shared helper in `seriesUtils.ts` that also returns the legacy series fields from the primary entry, uses it from both the match and the enrichment so the posted payload cannot contradict itself, and applies the same helper to the unmatched-files bulk add, which was reading Audnexus field names off an Audible-shaped response (`UnmatchedFilesModal.vue:452-460` at canary `a630572e`) and therefore always fell back to the file tag. `mapToAudible` is exported so its test is a unit test rather than a drive of the whole modal. Existing tests unchanged; four added.

Issue #822, opened 2026-08-15
Scan reports `Completed` with no files indexed and empty diagnostics when metadata extraction fails


kevinroberts:
### Summary

When metadata extraction fails for every candidate file, the scan job still reports status `Completed`, returns success from the API, writes **no** `AudiobookFile` rows, and records **no scan issues** — the diagnostics list comes back empty, not merely unhelpful. From the UI the book simply appears to have missing files, permanently, with no indication that anything went wrong.

Split out of #818, where it was filed as a secondary problem. @m4bard and I both think it deserves its own issue, and both of us stopped short of proposing a fix because where to draw the line is a behavioural decision across every scan path — that's a maintainer's call, not ours.

### Why it matters more than the bug that exposed it

#818 is a Linux-specific fault on one code path, and it has a patch. This is the thing that made it invisible.

In my case a book sat in the library for two days looking like a normal missing-file situation. There was nothing in the UI, nothing in the scan job's result, and nothing in diagnostics. The only signal anywhere was a single `[WRN]` line in the container log that identified the file by a bare file descriptor number rather than a path. Working out *which* file had failed took bisecting the library by re-pointing one book at a time.

The same silence hid a second, unrelated mis-import in the same library — a multi-book torrent attached to a single record — which I only found by reading the database directly.

Any future failure on this path will be equally silent, whatever its cause.

### Reproduction

Any condition that makes `ExtractFileMetadataAsync` return null or throw for all candidates will do it. Two known ones:

1. The `/proc` fd path bug in #818, on Linux, for any file path attribution declines.
2. @m4bard's fallback fixture in [m4bard/listenarr-testdata](https://github.com/m4bard/listenarr-testdata) — `./tools/validate_metadata_fallback.sh --image ghcr.io/listenarrs/listenarr:canary`. Their run on `f27c7989` reports the job completing with no files and an empty diagnostics list.

Observed on my install:

```
[WRN] MetadataService] Unable to extract metadata using ffprobe; using public filename metadata
[INF] MetadataService] Extracted basic metadata from file: 399
[INF] ScanQueueService] Updated scan job "fec900b9-..." status to Completed
```

Scan job result: `Completed`. Files indexed: 0. Issues recorded: 0.

### What I think is worth deciding

Stated as questions rather than a proposed patch, since these are product decisions:

[...]

Issue #821, opened 2026-08-15
Every `AudiobookFiles` row records `Size = 64` — `FileInfo` stats the `/proc` fd path instead of the file


kevinroberts:
### Summary

Every row in `AudiobookFiles` records `Size = 64`, regardless of the real file size. 64 is the size of a `/proc` fd symlink — the row is stat'ing the registration lease's descriptor path instead of the actual file.

Split out of #818 at @m4bard's suggestion. Same underlying descriptor path, but a **different call site on a different code path**, so it needs its own fix. This one is not confined to the metadata fallback: it is on the row-creation path that every ingest route shares.

### Measurement

Post-#717 canary, 217 of 217 rows:

```sql
sqlite> SELECT COUNT(*) FROM AudiobookFiles;
217
sqlite> SELECT COUNT(*) FROM AudiobookFiles WHERE Size = 64;
217
```

That includes rows created through the *working* path-attribution route, which never touch the metadata fallback in #818 — including one I created deliberately as a control after that issue was diagnosed. So the two faults are demonstrably independent.

### Cause

`listenarr.application/Audiobooks/Files/AudiobookFileService.cs`:

```csharp
var metadataPath = registrationLease?.MetadataPath ?? filePath;   // :173
...
var meta = await ExtractMetadataAsync(
    metadataPath,
    cacheIdentity,
    filePath);                                                    // :325-328  ← real path passed here

var fi = new FileInfo(metadataPath);                              // :330  ← descriptor path
var fileRecord = AudiobookFile.CreateUnresolved(filePath);
fileRecord.AudiobookId = audiobook.Id;
fileRecord.Size = fi.Exists ? fi.Length : null;                   // :333
```

Five lines above, `:325` already hands `filePath` over as the public half and gets the metadata right. Same method, correct for the probe, wrong for the size, with the value it needs already in scope.

@m4bard found a second site with the same shape, for replacement rows — `AudiobookFileService.PhysicalGeneration.cs:271`, assigned at `:274`:

```csharp
var fileInfo = new FileInfo(registrationLease.MetadataPath);      // :271
var replacement = AudiobookFile.CreateUnresolved(currentFile.Path);
replacement.AudiobookId = currentFile.AudiobookId;
replacement.Size = fileInfo.Exists ? fileInfo.Length : currentFile.Size;   // :274
```

[...]

PR #849, opened 2026-08-20
fix(scan): stop the Linux descriptor path leaking into metadata and size

### LINKED ISSUE #818: Linux: metadata fallback passes a `/proc` fd path to ffprobe, so files not claimed by path attribution can never be indexed
> **Status — 2026-08-14: root cause identified, and one of my original findings below was wrong.**
>
> On Linux the file registration lease sets its metadata path to `/proc/{pid}/fd/{fd}`.
> `AudiobookScanService.Metadata.cs:59-60` passes that descriptor path into the *single-argument*
> `ExtractFileMetadataAsync`, which builds `MetadataFileSource(path, path)` — so the **public** half
> becomes the descriptor path too. The guard at `FfmpegService.Probing.cs:44` then calls
> `FileUtils.IsAudioFile("/proc/1/fd/399")`, finds no extension, and throws at `:46` before ffprobe
> is ever started. The bare integer in the log is literally the file descriptor, because
> `LogRedaction.SanitizeFilePath` is `Path.GetFileName`.
>
> Diagnosed by **@m4bard** in the thread below, with a reproduction in
> [m4bard/listenarr-testdata](https://github.com/m4bard/listenarr-testdata); independently confirmed
> on my install. Introduced by #717. Windows is unaffected — the lease uses the canonical path there.
>
> **This only fires for files that path attribution declines.** `Metadata.cs:38-44` is a fallback,
> gated on `!attributed.Contains(path)`. Files claimed by path go through
> `AudiobookFileService.MetadataExtraction.cs:23`, which builds the correct two-argument
> `MetadataFileSource` and passes the guard. So the practical effect is: **any file the path
> heuristics don't claim can never be indexed at all on Linux**, silently.
>
> **Correction to "What I ruled out" below:** I wrote that the failure follows the file and that
> renaming it changes nothing. That is wrong, and it points away from the real cause — please
> disregard it. See the retraction section at the end.
>
> **Two secondary problems that were originally filed here have been split out**, since they are on
> different code paths and are independently fixable:
> - Per-file `Size = 64` on every row → #821
> - Scan reports `Completed` with no files and empty diagnostics → #822
>
> @m4bard has a three-site patch on a branch, waiting on a maintainer to say whether they'd prefer
> a PR or a diff.

---

### Summary

A valid `.m4b` file is rejected by an internal pre-check in `FfmpegService.RunFfprobeAsync` with `ffprobe target is not a supported audio file`, **before ffprobe is ever invoked**. `ffprobe` itself reads the same file without error (exit 0, `probe_score=100`).

The scan then silently falls back to filename metadata, reports **status `Completed`**, and creates **no `AudiobookFile` row**. The book shows up in the UI as having missing files, with no error surfaced anywhere except the log.

### Environment

| | |
|---|---|
| Listenarr | `1.3.0` (`ghcr.io/listenarrs/listenarr:canary`, post-#717) |
| Runtime | .NET 10.0.11 |
| OS | Linux 6.18.38 (x64), Unraid 7.x |
| Deployment | Docker, container runs as uid 0 |
| ffprobe | bundled, `/app/config/ffmpeg/ffprobe` (ffmpeg 7.0.2 amd64 static) |
| Library storage | XFS (direct disk path, not the Unraid FUSE user share) |

### The file

`Harry Potter and the Half-Blood Prince.m4b` — 713,423,915 bytes

```
format_name=mov,mp4,m4a,3gp,3g2,mj2
duration=66812.839184
probe_score=100
ftyp brand=mp42  (compatible: mp42, isom)

stream 0: aac       audio   mp4a  44100Hz  2ch
stream 1: png       video           (cover art)
stream 2: bin_data  data            (chapters)
```

### Log output

```
[INF] FfmpegService] Found bundled ffprobe at /app/config/ffmpeg/ffprobe
[WRN] MetadataService] Unable to extract metadata using ffprobe; using public filename metadata
Listenarr.Domain.SystemDiagnostics.Exceptions.FfmpegException: Error running ffprobe for 399
 ---> Listenarr.Domain.SystemDiagnostics.Exceptions.FfmpegException: ffprobe target is not a supported audio file: 399
   at Listenarr.Infrastructure.Ffmpeg.Installation.FfmpegService.RunFfprobeAsync(MetadataFileSource fileSource) in listenarr.infrastructure/Ffmpeg/Installation/FfmpegService.Probing.cs:line 46
   --- End of inner exception stack trace ---
   at Listenarr.Infrastructure.Ffmpeg.Installation.FfmpegService.RunFfprobeAsync(MetadataFileSource fileSource) in listenarr.infrastructure/Ffmpeg/Installation/FfmpegService.Probing.cs:line 96
   at Listenarr.Application.Metadata.Extraction.MetadataService.ExtractFileMetadataAsync(MetadataFileSource fileSource) in listenarr.application/Metadata/Extraction/MetadataService.cs:line 191
[INF] MetadataService] Extracted basic metadata from file: 399
[INF] ScanQueueService] Updated scan job "fec900b9-..." status to Completed
```

The inner exception comes from `Probing.cs:46`, i.e. a guard that runs *before* the ffprobe process is started.

### ffprobe disagrees

Running the bundled binary directly, inside the same container, on the same path:

```
$ /app/config/ffmpeg/ffprobe -v error \
    -show_entries format=format_name,duration,probe_score -of default=nw=1 \
    "/data/media/audiobooks/.../Harry Potter and the Half-Blood Prince.m4b"
format_name=mov,mp4,m4a,3gp,3g2,mj2
duration=66812.839184
probe_score=100
exit=0
```

### What I ruled out

Measured against every file Listenarr *has* successfully indexed in this library (~52 MP4-family files, ~160 others):

- **Not file size** — `It.m4b` (1.2 GB) and `Dune.m4b` (1.2 GB) index fine; the failing file is 713 MB.
- **Not the `ftyp` brand** — 7 successfully indexed files also use brand `mp42` (43 use `M4A `, 2 use `isom`).
- **Not PNG cover art** — at least one successfully indexed file also carries a PNG cover stream (46 use mjpeg).
- **Not the extension** — other `.m4b` files index fine; `.m4b` is in `AllowedFileExtensions`.
- ~~**Not the path or filename** — the failure follows the file, and renaming/relocating it does not change the outcome.~~ **This is wrong. See the retraction below — it is entirely about the filename.**

Notably, **`Harry Potter and the Order of the Phoenix, Book 5.m4b` (757 MB, same series, same folder tree) indexes successfully.** At the time I attributed this to the `ftyp` brand and cover-art codec, but neither survived scrutiny — both appear independently among files that index fine. The actual difference turned out to be the filename; see below.

### Retraction, and what the rest of the library shows

I reported that renaming the file changed nothing. Whatever names I tried at the time evidently did not produce a *match* against the folder and record, which is the specific thing that decides whether path attribution claims a file. That bullet sent the investigation toward file-format differences, and it should not have.

Looking at the whole set rather than the one file — seven Harry Potter books, same narrator, same folder tree, six of them claimed without incident:

```
book   folder name suffix   filename suffix   embedded title suffix   claimed
1-5,7  ", Book N"           ", Book N"        ", Book N"              yes
6      ", Book 6"           (none)            (none)                  no
```

Book 6 is the only file in the set whose name disagrees with its folder, and the only one that has ever reached the metadata fallback. Six controls and one fallback case, in a library that was never built as an experiment.

I then renamed the file to match and attached it to the correct record. Same unpatched `f27c7989` container, same storage, no code change:

```
Id    AudiobookId   DurationSeconds   Codec   Source
262   129           66812.839184      aac     LibraryScan
```

It probes fine, and always could. The rejection was never a property of this file — only of the route it was forced down. The bug reproduces and un-reproduces purely by moving a file between the attributed and fallback paths.

To be clear about the limits of that: it corroborates the diagnosis, it does not test the fix, because it avoids the broken path rather than exercising it. @m4bard's control/fallback harness is the thing that validates a patch.

Worth noting that there appear to be at least two ways to defeat path attribution and land in the fallback: a title/filename mismatch (this report) and no author anywhere on the path (@m4bard's fixture). A fix probably wants a fixture for each, since the mismatch case seems likelier to happen by accident.

### Related

- #821 — per-file `Size = 64`, split out of this issue
- #822 — scan reports `Completed` with no files and empty diagnostics, split out of this issue
- #542 — book-level `Audiobook.FileSize` is never summed. Different field and different cause, but the two compound: a correct sum over incorrect per-file sizes is still wrong.
- #737 (closed) — same subsystem, opposite symptom: probe returns *empty* and a row is written with nulls. Here the probe is *refused* and no row is written at all.
- #820 — unrelated, but it makes the repo's test suite fail in bulk on some filesystems, which affects anyone trying to validate a patch here.

m4bard:
## Summary

On Linux a registration lease's metadata path is a `/proc/{pid}/fd/{fd}` descriptor link rather than a path to the file, and two separate readers treated it as though it were one. This fixes the first: the probe's audio-extension guard rejects a candidate whose public half has no extension, so the embedded-metadata pass never runs.

The second reader is `FileInfo` on the link, which reports the length of the link rather than of the file. That is #821 and it is not fixed here; see the note below.

Fixes #818. Rebased onto `03958c15` as promised on #818 once #819 landed.

This originally claimed #821 as well and no longer does. The size-site half was split out of this branch and is not in the current diff, so leaving the claim in place would have auto-closed that issue on merge with nothing here addressing it. The size-site half is covered by #901, which handles both sites rather than the one this branch used to touch.

The mechanism was worked out jointly with @kevinroberts on that thread rather than by me alone. He reproduced it independently on unpatched canary, withdrew his own suggested validation case once it turned out not to exercise the pass it was meant to test, and rewrote the issue around the root cause. The second of those is the reason the check below varies which agreements break instead of asserting one construction.

### Fixed
- `AudiobookScanService.Metadata.cs` passes `new MetadataFileSource(pinnedMetadataFile.MetadataPath, candidate)` instead of collapsing both halves onto the descriptor path. The read path stays the pinned generation; the public path keeps the real filename, so the guard sees an audio extension and ffprobe runs.

The pinning guarantee is unchanged. The read path stays the pinned generation, which is the property the lease exists to provide.

## Testing

Two existing tests needed to follow the read onto the two-part source. Both mocked only the single-path overload, so a strict mock saw no matching setup once the scan routed through `MetadataFileSource`, the extractor returned nothing, and every file read as unreadable.

`ScanAsync_MetadataReplacementAndRestore_ReadsPinnedFileGeneration` needed a decision rather than a swap. Its callback reads `ReadPath`, because the point of the test is that the scan sees the original generation while the visible file is swapped underneath it. It now also asserts `PublicPath` is still the candidate with its extension intact, and that assertion is load-bearing: collapsing both halves onto the descriptor path fails it with the real path expected and `/proc/<pid>/fd/<fd>` observed. That is this defect, in a unit test, where before it was only reachable through a running container.

Full suite: 3,029 passed, 0 failed, 125 skipped, which is the baseline on `03958c15` unchanged, since this modifies two existing tests rather than adding any.

Reproduced before and after against `ghcr.io/listenarrs/listenarr:canary` and a build of this branch, with a public check that varies which agreements are broken rather than asserting a single case:

| construction | canary | patched |
|---|---|---|
| claimed by path attribution | claimed | claimed |
| filename no longer matches the title | claimed | claimed |
| folder no longer matches the record's title | claimed | claimed |
| neither folder nor filename matches | **UNCLAIMED** | claimed |
| only the embedded tags identify it | **UNCLAIMED** | claimed |

Both agreements have to break before anything reaches the pass, which is why the two single-mismatch rows are controls rather than filler: each rescues the other, so a check that only broke the filename would never exercise the path it was written to test.

## What this does not fix

Worth being explicit, since the branch touches the neighbourhood of three other issues.

**#542 is untouched, and out of scope here now that this branch changes no size value.** The book-level total still reads "not set" on both builds. #901 corrects the per-file rows; whatever should sum them is a separate defect and I have not looked at it here.

**#822 is only partly addressed, and not by this.** #819 added an early return recording `MetadataEnrichmentSkippedLimitedStorage`, which covers the limited-storage case. A probe refusal on storage that passes the generation check still commits `Completed` with nothing recorded and no diagnostic. That is what the matrix rows above show as UNCLAIMED with no diagnostic, and it stays true after this change for any other reason a probe might refuse.

I verified #819's early return does not shadow this: it returns at `:25` and the patched call is at `:73`, so on storage with durable generation proof execution still reaches it.

## On the scope

This asked earlier whether you would rather have the two issues as two PRs. They are two PRs now: this one is the metadata-fallback path, and the size sites are #901's. So there is nothing to split and no answer needed.

### DIFF
--- a/listenarr.application/Audiobooks/Files/AudiobookFileService.MetadataExtraction.cs
+++ b/listenarr.application/Audiobooks/Files/AudiobookFileService.MetadataExtraction.cs
@@ -5,44 +5,6 @@ namespace Listenarr.Application.Audiobooks.Files;
 
 public partial class AudiobookFileService
 {
-    /// <summary>
-    /// The byte length of the file a registration lease holds open.
-    /// </summary>
-    /// <remarks>
-    /// A lease's metadata path is not always a path to the file. On Linux it is a
-    /// <c>/proc/{pid}/fd/{fd}</c> descriptor link, and stat'ing the link reports the length of
-    /// the link itself rather than of its target, which is a constant 64 bytes. Reading the
-    /// length from the pinned handle keeps the generation guarantee the lease exists to
-    /// provide, since it never consults the visible path, and reports the bytes the file
-    /// actually has.
-    ///
-    /// Leases that do not expose generation-bound reads fall back to the metadata path, which
-    /// is the public path for those callers.
-    /// </remarks>
-    private static long? ResolveRegisteredLength(
-        IAudiobookFileRegistrationLease? registrationLease,
-        string metadataPath)
-    {
-        if (registrationLease != null)
-        {
-            try
-            {
-                using var stream = registrationLease.OpenMetadataReadStream();
-                if (stream.CanSeek)
-                {
-                    return stream.Length;
-                }
-            }
-            catch (NotSupportedException)
-            {
-                // The lease does not expose generation-bound reads; fall through to the path.
-            }
-        }
-
-        var fileInfo = new FileInfo(metadataPath);
-        return fileInfo.Exists ? fileInfo.Length : null;
-    }
-
     private async Task<AudioMetadata?> ExtractMetadataAsync(
         string metadataPath,
         string cacheIdentity,
--- a/listenarr.application/Audiobooks/Files/AudiobookFileService.MetadataExtraction.cs
+++ b/listenarr.application/Audiobooks/Files/AudiobookFileService.MetadataExtraction.cs
@@ -5,6 +5,44 @@ namespace Listenarr.Application.Audiobooks.Files;
 
 public partial class AudiobookFileService
 {
+    /// <summary>
+    /// The byte length of the file a registration lease holds open.
+    /// </summary>
+    /// <remarks>
+    /// A lease's metadata path is not always a path to the file. On Linux it is a
+    /// <c>/proc/{pid}/fd/{fd}</c> descriptor link, and stat'ing the link reports the length of
+    /// the link itself rather than of its target, which is a constant 64 bytes. Reading the
+    /// length from the pinned handle keeps the generation guarantee the lease exists to
+    /// provide, since it never consults the visible path, and reports the bytes the file
+    /// actually has.
+    ///
+    /// Leases that do not expose generation-bound reads fall back to the metadata path, which
+    /// is the public path for those callers.
+    /// </remarks>
+    private static long? ResolveRegisteredLength(
+        IAudiobookFileRegistrationLease? registrationLease,
+        string metadataPath)
+    {
+        if (registrationLease != null)
+        {
+            try
+            {
+                using var stream = registrationLease.OpenMetadataReadStream();
+                if (stream.CanSeek)
+                {
+                    return stream.Length;
+                }
+            }
+            catch (NotSupportedException)
+            {
+                // The lease does not expose generation-bound reads; fall through to the path.
+            }
+        }
+
+        var fileInfo = new FileInfo(metadataPath);
+        return fileInfo.Exists ? fileInfo.Length : null;
+    }
+
     private async Task<AudioMetadata?> ExtractMetadataAsync(
         string metadataPath,
         string cacheIdentity,
--- a/listenarr.application/Audiobooks/Files/AudiobookFileService.PhysicalGeneration.cs
+++ b/listenarr.application/Audiobooks/Files/AudiobookFileService.PhysicalGeneration.cs
@@ -274,10 +274,11 @@ public partial class AudiobookFileService
         string? source,
         bool replaceMetadata)
     {
-        var fileInfo = new FileInfo(registrationLease.MetadataPath);
         var replacement = AudiobookFile.CreateUnresolved(currentFile.Path);
         replacement.AudiobookId = currentFile.AudiobookId;
-        replacement.Size = fileInfo.Exists ? fileInfo.Length : currentFile.Size;
+        replacement.Size = ResolveRegisteredLength(
+            registrationLease,
+            registrationLease.MetadataPath) ?? currentFile.Size;
         replacement.DurationSeconds = replaceMetadata
             ? metadata?.Duration.TotalSeconds
             : Math.Abs(metadata?.Duration.TotalSeconds ?? 0) > double.Epsilon
--- a/listenarr.application/Audiobooks/Files/AudiobookFileService.PhysicalGeneration.cs
+++ b/listenarr.application/Audiobooks/Files/AudiobookFileService.PhysicalGeneration.cs
@@ -274,11 +274,10 @@ public partial class AudiobookFileService
         string? source,
         bool replaceMetadata)
     {
+        var fileInfo = new FileInfo(registrationLease.MetadataPath);
         var replacement = AudiobookFile.CreateUnresolved(currentFile.Path);
         replacement.AudiobookId = currentFile.AudiobookId;
-        replacement.Size = ResolveRegisteredLength(
-            registrationLease,
-            registrationLease.MetadataPath) ?? currentFile.Size;
+        replacement.Size = fileInfo.Exists ? fileInfo.Length : currentFile.Size;
         replacement.DurationSeconds = replaceMetadata
             ? metadata?.Duration.TotalSeconds
             : Math.Abs(metadata?.Duration.TotalSeconds ?? 0) > double.Epsilon
--- a/listenarr.application/Audiobooks/Files/AudiobookFileService.cs
+++ b/listenarr.application/Audiobooks/Files/AudiobookFileService.cs
@@ -327,11 +327,10 @@ namespace Listenarr.Application.Audiobooks.Files
                     cacheIdentity,
                     filePath);
 
+                var fi = new FileInfo(metadataPath);
                 var fileRecord = AudiobookFile.CreateUnresolved(filePath);
                 fileRecord.AudiobookId = audiobook.Id;
-                fileRecord.Size = ResolveRegisteredLength(
-                    registrationLease,
-                    metadataPath);
+                fileRecord.Size = fi.Exists ? fi.Length : null;
                 fileRecord.Source = source;
                 fileRecord.CreatedAt = DateTime.UtcNow;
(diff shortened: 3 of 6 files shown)


m4bard:
I have cut this PR down. #901 covers half of what it was doing, and covers it better than I did.

This originally fixed two things behind the same `/proc/{pid}/fd/{fd}` link. The size, where `FileInfo` stats the link and reports its constant 64 bytes rather than the file's length. And the metadata read, where the probe's audio-extension guard rejects a candidate whose path has no extension, so ffprobe never runs at all.

#901 fixes the size. It opens the metadata path and takes the stream length unconditionally. Mine read the length through the lease's generation-bound stream and then fell back to `FileInfo` on the metadata path when the lease did not expose one, which on Linux is the original 64-byte bug again. That fallback was mine and it was a hole. #901 does not have it.

Worth adding that my size half had no test of its own here either. Both test files on this branch cover the scan boundary, not the size.

So the size change is gone and the two files it touched are back at canary. `AudiobookFileService.cs` and `AudiobookFileService.PhysicalGeneration.cs` are untouched by this PR now, which means it no longer shares a file with #901 or with #902. Whichever order they land in, neither has to resolve anything against the other.

## What is left

One production change, in `AudiobookScanService.Metadata.cs`.

The scan passed the lease's metadata path as both the byte source and the media identity. Those are deliberately separate things: the lease exists to give stable byte access while the public path stays the file's real identity, and the single-path overload collapses the two. On Linux the metadata path is an extensionless `/proc` link, so the collapse makes the extension guard reject the candidate before ffprobe is reached. Passing the candidate as the identity alongside the metadata path as the byte source keeps the guard looking at the real filename.

That is a different symptom from a wrong size. A file with this problem has no embedded metadata read at all, rather than metadata plus a bad length, so #901 landing does not make it go away.

## Checks

Full suite unfiltered: 3039 passed, 0 failed, 127 skipped, which is the canary baseline. Both builds clean at 0 warnings. The test changes here modify existing scan tests rather than adding new ones, so the count does not move.

Control: reverting `AudiobookScanService.Metadata.cs` to canary and keeping the tests fails 2 of the 29 scan tests. Restoring it goes green.

Test-merged against canary, against #901 and against #902: all three clean.

I have not compared my scan change against anything in #902 beyond the merge, so if it overlaps in intent rather than in text, say so and I will look again.

Issue #852, opened 2026-08-20
A cross-volume Move is refused deliberately, but the refusal is invisible, so the import just never happens


m4bard:
On canary `03958c15` (1.3.2), an import configured to **move** a completed file onto a root folder that sits on a different mount does nothing. No destination appears, the source is untouched, and nothing surfaces as an error.

I want to be clear up front that I do not think the refusal itself is a bug, and I am not asking for it to be reverted. Listenarr blocks this case on purpose and says why. What I think is wrong is that the block is indistinguishable from silence.

## What I measured

Two cases, identical except for whether the source and the root folder are on the same mount:

```
sameMount   destination written, source consumed  -> moved      OK
sepMounts   no destination after 45s, source untouched -> stalled  FAIL
```

The same-mount case is the control. It rules out "move is broken" and leaves only the mount boundary as the difference.

The server does record something, and it is a good message:

```
[INF] [FileMover] File mutation Blocked: Move The Valley of Fear.m4b -> The Valley of Fear.m4b.
      Reason: Unix cross-volume registration moves require source retirement that cannot be
      generation-fenced without a library-side namespace claim
```

That is at **Information**. The check waits 45 seconds and gives up; I have not established whether anything retries later.

Reproducing it:

```bash
git clone https://github.com/m4bard/listenarr-testdata && cd listenarr-testdata
python3 -m venv .venv && .venv/bin/pip install -e .
./tools/validate_import_action.sh ghcr.io/listenarrs/listenarr:canary --action move
```

The source assertion inverts for this action and it is worth saying why: hardlink, symlink and copy must all leave the source in place, because an import that removes it is data loss, but a move must remove it. So the tool reports three outcomes rather than two, and "nothing happened at all" is `stalled` rather than an ordinary failure.

## Where it comes from

The refusal is at two sites, `FileMover.MarkerlessMove.cs:67` and `FileMover.MarkerlessRegistration.cs:74`, both guarded by `!OperatingSystem.IsWindows() && (ForceCrossVolumeForTest || !initialSource.IsOnSameVolume(...))`. Both call `LogMutation(FileMutationOutcome.Blocked, ...)` and return.

The visibility problem is one method, `FileMover.Actions.cs:408`:

```csharp
private void LogMutation(FileMutationOutcome outcome, FileAction action, string source, string? destination, string? reason = null)
{
    var result = new FileMutationResult(outcome, action, source, destination, reason);
    _logger.LogInformation(
[...]

PR #853, opened 2026-08-20
fix(filemover): log a blocked or failed mutation above Information

m4bard:
## Summary

`LogMutation` logs every outcome at `Information`, so a refusal to perform a file mutation sits at the same level as a successful one. `Blocked` and `Failed` both mean the action did not happen.

This changes the level for those two and nothing else. Full write-up in #852.

### Fixed
- `FileMover.LogMutation` selects `Warning` for `Blocked` and `Failed`, `Information` otherwise.

## Why this and not more

The case that surfaced it is the Unix cross-volume `Move` refusal, which is deliberate and states its reason: the source retirement cannot be generation-fenced without a library-side namespace claim. **That refusal is not touched here and I am not proposing it should be.** What this changes is that an operator watching at the default level can now tell the difference between a move that was refused and a move that succeeded.

It applies to all outcomes rather than special-casing the cross-volume one because none of the seven `Blocked` sites in this class is routine: a non-publishable action, an empty operation ID, unavailable durable markerless state, linked aliases of the same file. Every one is a refusal to do what was asked.

Worth stating plainly: **this does not make cross-volume moves work.** It makes their refusal audible. Whether to implement the namespace claim described in that reason is a separate and much larger question.

## Testing

`FileMoverMarkerlessMoveTests` gains a case that drives the real refusal through the existing `ForceCrossVolumeForTest` hook and asserts the level rather than the message text. `CreateMover` gains an optional `ILogger<FileMover>`; it previously passed `NullLogger` unconditionally, which is why no existing test in that file could observe a level. There is already a test asserting this move is rejected, so the gap was never whether it refuses, only how loudly.

Verified as a real guard: with the level flattened back to `Information` the new test fails with `Expected: Warning, Actual: Information`, and passes with the change.

Reproduced end to end against `ghcr.io/listenarrs/listenarr:canary` with a public check that runs the same import twice, once with the source and root folder on one mount and once across two. The same-mount case is the control and completes; the cross-mount case produces no destination and leaves the source untouched. The check is in the test-data repo linked from the issue, and it reports "nothing happened at all" as its own verdict rather than as an ordinary failure, since a stall and a failure are different things to operate.

### DIFF
--- a/listenarr.infrastructure/FileSystem/FileMover.Actions.cs
+++ b/listenarr.infrastructure/FileSystem/FileMover.Actions.cs
@@ -408,7 +408,15 @@ public partial class FileMover
     private void LogMutation(FileMutationOutcome outcome, FileAction action, string source, string? destination, string? reason = null)
     {
         var result = new FileMutationResult(outcome, action, source, destination, reason);
-        _logger.LogInformation(
+        // Blocked and Failed mean the requested action did not happen, and the caller only gets
+        // back a bool, so this line is the sole record that anything went wrong. Logging it at
+        // Information put it at the same level as a successful move, which is how a refused
+        // cross-volume move reads as silence to anyone watching at the default level.
+        var level = outcome is FileMutationOutcome.Blocked or FileMutationOutcome.Failed
+            ? LogLevel.Warning
+            : LogLevel.Information;
+        _logger.Log(
+            level,
             "File mutation {Outcome}: {Action} {Source} -> {Destination}. Reason: {Reason}",
             result.Outcome,
             result.Action,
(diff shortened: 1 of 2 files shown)


m4bard:
Closing this. #864 landed the same fix, and did it better.

`LogMutation` on canary now switches on the outcome (`listenarr.infrastructure/FileSystem/FileMover.Actions.cs:433-446`):

```csharp
case FileMutationOutcome.Blocked:
    _logger.LogWarning(template, arguments);
    break;
case FileMutationOutcome.Failed:
    _logger.LogError(template, arguments);
    break;
case FileMutationOutcome.Skipped:
    _logger.LogDebug(template, arguments);
    break;
default:
    _logger.LogInformation(template, arguments);
    break;
```

This PR raised `Blocked` and `Failed` to Warning. Splitting them and putting `Failed` at Error is the right call, and `Skipped` at Debug is a distinction I did not make at all.

Nothing left here. #852 is closed and the refusal is no longer silent at the default level.

Issue #859, opened 2026-08-20
GET /api/v1/images/{identifier} returns 500: dynamic access to an anonymous type across an assembly boundary throws RuntimeBinderException


m4bard:
Observed on canary `03958c15` (1.3.2). All line numbers below are from that commit and were re-resolved against it rather than carried forward.

The ASINs in this report are illustrative substitutes taken from a public-domain test corpus, not the values observed. I have kept them distinct from each other because the second half of the report is about two identifiers differing.

## Symptom

Image lookup fails with a 500 and this in the log:

```
[ERR] [Listenarr.Api.Features.Images.ImagesController] Error retrieving image for identifier: B008DFUGCQ
Microsoft.CSharp.RuntimeBinder.RuntimeBinderException: 'object' does not contain a definition for 'metadata'
   at CallSite.Target(Closure, CallSite, Object)
   at Listenarr.Api.Features.Images.ImageCandidateLookupWorkflow.TryResolveAsync(String identifier, String relativePath, String requestedRegion)
```

## Root cause

`IAudiobookMetadataService.GetMetadataAsync` is declared as `Task<object?>`, and on the success path `AudiobookMetadataService` returns an anonymous type (`listenarr.application/Metadata/Core/AudiobookMetadataService.cs:121`):

```csharp
return new
{
    metadata = result,
    source = source.Name,
    sourceUrl = source.BaseUrl
};
```

The consumer unwraps that envelope with `dynamic` (`listenarr.api/Features/Images/ImageCandidateLookupWorkflow.cs:340`):

```csharp
// Try dynamic access
dynamic env = metadataEnvelope;
object? mdObj = env.metadata;
```

The C# compiler emits anonymous types as `internal`. That envelope is internal to `Listenarr.Application`, and the call site lives in `Listenarr.Api`. The runtime binder resolves member access against what is visible from the call site's assembly, so `metadata` is not a member it can bind to. That also explains why the message names `'object'`: the binder never saw the anonymous type, so it reported against the only type it could see.

`listenarr.application/Listenarr.Application.csproj:20` grants `InternalsVisibleTo` to `Listenarr.Tests` only, so `Listenarr.Api` gets no exemption.

I think this makes the throw deterministic rather than intermittent. Whenever the fallback runs and gets back anything non-null, line 342 throws.

## Why it escapes the local handler

Line 342 sits inside a `try` whose `catch` is filtered (`ImageCandidateLookupWorkflow.cs:381`):

```csharp
[...]

PR #860, opened 2026-08-20
fix(images): read the metadata envelope by reflection, not dynamic

m4bard:
## Summary

`GET /api/v1/images/{identifier}` returns 500 whenever the metadata fallback runs, because the envelope it gets back is read with `dynamic` across an assembly boundary. This reads it by reflection instead.

Full write-up in #859.

### Fixed
- `ImageCandidateLookupWorkflow` reads `metadata` off the envelope with `GetType().GetProperty(...)` rather than `dynamic`.

`GetMetadataAsync` is declared `Task<object?>` and returns an anonymous type. The compiler emits anonymous types as `internal`, so the envelope is internal to `Listenarr.Application` while this call site is in `Listenarr.Api`. The runtime binder resolves members against what is visible from the call site's assembly, so it cannot see `metadata` and throws `RuntimeBinderException` reported against `object`. Reflection is not subject to that.

`RuntimeBinderException` is not in `IsRecoverableImageLookupException`, so it escapes the filtered catch that was written to make a malformed envelope degrade quietly, and reaches the controller catch-all. The throw is deterministic on that path; what makes it rare is the guard above, since most requests resolve a URL earlier and never enter the fallback.

## Why reflection rather than something nicer

Because it is what the surrounding code already does. Ten lines below the change, the same block reads `ImageUrl` and `Isbn` off the inner object with `t.GetProperty(...)`. `LibraryMetadataRescanWorkflow.TryExtractMetadataLookupResult` unwraps this exact envelope the same way and works. Of the four callers of `GetMetadataAsync`, this was the only one using `dynamic`, and the only one that throws.

The better fix is a named public record for the envelope so `GetMetadataAsync` stops returning `object?`, which would remove the reason three consumers each invented their own unwrapping. I did not do that here because it changes an interface with four call sites, and whether that signature should stay `object?` is a design decision rather than a bug fix. This does not foreclose it. The issue lays out both.

## Testing

`ImagesController_MetadataDownloadFallbackTests` gains a case using an anonymous envelope, which is the shape production actually returns. Because the anonymous type is emitted internal to the test assembly and the workflow lives in `Listenarr.Api`, it reproduces the same accessibility relationship as production.

The existing test in that file covers this fallback but mocks a bare `AudibleBookResponse`, and its own comment says that is to avoid "anonymous envelope issues". That takes the `is AudibleBookResponse` branch and never reaches the unwrap, so the one shape that fails was the one shape not covered.

Verified as a real guard: with `dynamic` restored the new test fails, and it fails by never calling the downloader at all, which is the actual production symptom rather than a proxy for it.

## One note on the comment

`BackendArchitectureTests.ActiveProductionSourceFiles_RemainFocused` caps a production source file at 500 lines, and this file is at 487 on canary. My first version of the explanatory comment took it to 503 and failed that test. The comment is deliberately short as a result and the full reasoning lives in the issue. Flagging it because it is the kind of budget worth knowing about before writing prose into a file near the limit.

### DIFF
--- a/listenarr.api/Features/Images/ImageCandidateLookupWorkflow.cs
+++ b/listenarr.api/Features/Images/ImageCandidateLookupWorkflow.cs
@@ -8,6 +8,7 @@
  * (at your option) any later version.
  */
 
+using System.Reflection;
 
 namespace Listenarr.Api.Features.Images
 {
@@ -337,9 +338,15 @@ namespace Listenarr.Api.Features.Images
                                     }
                                     else
                                     {
-                                        // Try dynamic access
-                                        dynamic env = metadataEnvelope;
-                                        object? mdObj = env.metadata;
+                                        // Not `dynamic`: the envelope is an anonymous type, emitted
+                                        // internal to Listenarr.Application, so the binder cannot
+                                        // see `metadata` from this assembly and throws. Reflection
+                                        // is how the inner object is read below, and how
+                                        // LibraryMetadataRescanWorkflow reads this same envelope.
+                                        object? mdObj = metadataEnvelope.GetType().GetProperty(
+                                                "metadata",
+                                                BindingFlags.Public | BindingFlags.Instance | BindingFlags.IgnoreCase)
+                                            ?.GetValue(metadataEnvelope);
 
                                         // If it's already the Audible type, use it
                                         if (mdObj is AudibleBookResponse mdMeta)
--- a/tests/Features/Api/Features/Images/ImagesController_MetadataDownloadFallbackTests.cs
+++ b/tests/Features/Api/Features/Images/ImagesController_MetadataDownloadFallbackTests.cs
@@ -96,5 +96,107 @@ namespace Listenarr.Tests.Features.Api.Features.Images
                 // Best-effort test cleanup; ignore cleanup failures.
             }
         }
+
+        // The test above returns the AudibleBookResponse directly, and its own comment says that is
+        // to avoid "anonymous envelope issues". But an anonymous envelope is what the production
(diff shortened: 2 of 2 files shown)


m4bard:
Closing this. #878 fixed it, and fixed it better than this PR did.

This changed `ImageCandidateLookupWorkflow` to read the metadata envelope by reflection instead of `dynamic`, because the envelope is an anonymous type and anonymous types are `internal`, so the binder throws `RuntimeBinderException` across an assembly boundary. Reflection is not subject to binder accessibility, so it worked, but it was a workaround for the binding rather than a fix for the shape.

At `a630572e` the `dynamic` is simply gone. `metadataEnvelope.Metadata.ImageUrl` and `.Isbn` are read directly, and the file lost 49 lines in the process. There is nothing left for reflection to work around, so keeping this open would mean reapplying a workaround for a problem that no longer exists.

#859 is closed too, which is the right outcome.

## The tests

The only other thing here is 102 lines added to `ImagesController_MetadataDownloadFallbackTests.cs`, and I am not proposing to salvage them.

They assert `Assert.IsNotType<StatusCodeResult>(result)`, which says the endpoint did not return a 500 and nothing about whether the image was found. That was adequate as a guard for a crash, and it is weak as a regression test for the behaviour. Strongly-typed property access cannot raise the exception it was guarding, so what it protects against is now unreachable.

That file is also one of the nine I raised in #897 for accepting either outcome, so adding more to it would sit badly beside that issue.

If a regression test for the fallback path is wanted, it should assert the served path rather than the absence of a 500, and it belongs with whatever comes of #897 rather than bolted onto a superseded PR.

Nothing here needs carrying forward. Thanks for picking it up.

Issue #862, opened 2026-08-21
Search result scoring queries one scoped DbContext from every task in a parallel batch, and loses retention silently when it collides


m4bard:
> **Corrections since filing.** The body below is left exactly as written. Two claims in it are wrong, and both are corrected in the comments rather than edited away.
>
> - **"That leaves this one site" is wrong.** There are at least two, and the other one is larger: `DownloadClientGateway` issues a path-mapping query per queue item inside a `Task.WhenAll`, nested inside the poller's own fan-out across clients. Filed separately as #870 with PR #871. Detail in https://github.com/Listenarrs/Listenarr/issues/862#issuecomment-5375990418
> - **"not present in the class it names any more" is wrong.** `DownloadClientQueuePoller` never held a repository or a `DbContext`, including at `4555ad21`, the commit #783 reports against. Nothing was refactored away, so #783 is not stale in the way I implied. Detail in https://github.com/Listenarrs/Listenarr/issues/862#issuecomment-5375847555
>
> Neither affects the scoring defect described below or the fix in #863, both of which were verified separately.

---

On canary `03958c15` (1.3.2), `QualityProfileService.ScoreSearchResults` scores a batch with `Task.WhenAll`, and each task queries the same scoped `IIndexerRepository` for its result's indexer. A batch of twelve results issues twelve concurrent queries against one `DbContext`, which EF refuses.

The request does not fail. That is the part I think matters.

## What actually happens when it collides

`SearchResultScorer` wraps the lookup in a catch that logs at `Debug` (`listenarr.application/Search/Scoring/SearchResultScorer.cs:150-153`):

```csharp
catch (Exception ex) when (ex is not OperationCanceledException && ...)
{
    _logger.LogDebug(ex, "Failed to fetch indexer retention for IndexerId {Id}", searchResult.IndexerId.Value);
}
```

So a result that loses the race keeps `indexerRetention = 0` and never runs the branch that sets `isNzb` from the indexer's type. Age and retention checks then score it against assumptions that do not match its indexer, and the batch comes back ordered plausibly and quietly wrong. At the default log level nothing appears at all.

## Why it is a shared context

`ListenArrDbContext` is registered once, at `listenarr.infrastructure/DependencyInjection/Persistence/PersistenceRegistrationExtensions.cs:25`:

```csharp
services.AddDbContextFactory<ListenArrDbContext>(configureDb, ServiceLifetime.Singleton);
```

[...]

PR #863, opened 2026-08-21
fix(scoring): resolve indexers once per batch, not once per result in parallel

m4bard:
## Summary

`ScoreSearchResults` fans out with `Task.WhenAll` and each task queried the same scoped `IIndexerRepository`, so a batch of N results issued N concurrent queries against one scoped `DbContext`. This resolves each distinct indexer once before the fan-out and passes the results in.

Full write-up in #862.

### Fixed
- `QualityProfileService.ScoreSearchResults` resolves the batch's distinct indexers sequentially up front and hands them to the scorer.
- `SearchResultScorer` takes an optional pre-resolved lookup and prefers it over querying.

The single-result `ScoreSearchResult` is unchanged and still queries; there is nothing to batch there, and it stays on the interface as it was.

## Why it matters more than a caught exception suggests

The scorer already catches the failure and logs at `Debug`, so nothing fails. What it loses is `indexerRetention`, which stays 0, and the branch that sets `isNzb` from the indexer's type. Age and retention checks then score that result against assumptions that do not match its indexer. The batch comes back ordered plausibly and quietly wrong, and at the default log level there is nothing to see.

Three callers reach it: the quality profile controller, download submission, and the six-hour automatic search sweep.

## Fewer queries, not just safer ones

A batch commonly carries many results across a handful of indexers. This goes from one query per result to one per distinct indexer, so it removes redundant work as well as the overlap.

`IndexerSearchWorkflow` already fetches its indexers before its own fan-out rather than inside it, so this makes the two agree rather than introducing an approach.

## Testing

Asserting on the EF exception would mean racing it, which makes for a flaky test and a weak claim. The new test counts the overlap directly: a stub `IIndexerRepository` records the highest number of calls it ever had in flight at once, with a small delay so the overlap is observable rather than a scheduling accident.

It reports **12 concurrent without the change and 1 with it**, over twelve results across three indexers. It also pins the call count at 3, one per distinct indexer, so a future change that restores per-result querying fails even if it somehow avoids overlapping.

## Notes

The context's lifetime is the load-bearing fact here, so I verified it rather than inferring it from the registration. Reading the `ServiceDescriptor` back out of the built collection gives `ListenArrDbContext => Scoped`, with the repositories scoped as well. `AddDbContextFactory(..., ServiceLifetime.Singleton)` at `PersistenceRegistrationExtensions.cs:25` registers the factory as a singleton and the context as scoped.

I have not reproduced wrong scores end to end against a running instance, and the issue says so. A collision needs the queries to actually overlap in time, which is machine-dependent. What is deterministic is that the calls are issued concurrently against a context that permits one at a time.

### DIFF
--- a/listenarr.application/Audiobooks/Quality/QualityProfileService.cs
+++ b/listenarr.application/Audiobooks/Quality/QualityProfileService.cs
@@ -190,9 +190,15 @@ namespace Listenarr.Application.Audiobooks.Quality
             }
         }
 
-        public async Task<QualityScore> ScoreSearchResult(SearchResult searchResult, QualityProfile profile)
+        public Task<QualityScore> ScoreSearchResult(SearchResult searchResult, QualityProfile profile) =>
+            ScoreSearchResult(searchResult, profile, resolvedIndexers: null);
+
+        private async Task<QualityScore> ScoreSearchResult(
+            SearchResult searchResult,
+            QualityProfile profile,
+            IReadOnlyDictionary<int, Indexer>? resolvedIndexers)
         {
-            var scorer = new SearchResultScorer(_indexerRepository, _logger);
+            var scorer = new SearchResultScorer(_indexerRepository, _logger, resolvedIndexers);
             var score = await scorer.Score(searchResult, profile);
 
             // Also calculate the Prowlarr-style composite (Smart) score so the UI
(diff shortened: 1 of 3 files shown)


m4bard:
Closing in favor of #921. This PR's own indexer-batching fix is not abandoned; it moved. The batch-resolve-once change and its concurrency test are now commits 3fddf4680 and 713f152d6 on #921's branch. #921's own body already flagged the conflict with this PR in `SearchResultScorer.cs`, so this fix is folded in there instead of being rebased separately.

Issue #870, opened 2026-08-21
A queue poll issues one path-mapping query per item, in parallel, against a scoped DbContext


m4bard:
On canary `03958c15` (1.3.2), polling a download client's queue issues a database query for every item in the queue, all concurrently, against a `DbContext` that permits one operation at a time. A queue of thirty items issues up to sixty overlapping queries.

I think this is the origin of the stack trace in #783, which has been open since v1.2.2.

## The chain

Everything on it is scoped, so there is one instance of each per scope and one `DbContext` behind them all:

```
DownloadRegistrationExtensions.cs:84   AddScoped<IDownloadClientGateway, DownloadClientGateway>
DownloadRegistrationExtensions.cs:85   AddScoped<IRemotePathMappingService, RemotePathMappingService>
DownloadRegistrationExtensions.cs:94   AddScoped<IRemotePathMappingRepository, EfRemotePathMappingRepository>
```

`DownloadClientGateway.GetQueueAsync:96` fans out over the queue:

```csharp
var items = await adapter.GetQueueAsync(client, ct);
var tasks = items.Select(item => TranslateQueueItemPathsAsync(client, item));
return [.. await Task.WhenAll(tasks)];
```

`TranslateQueueItemPathsAsync` calls `TranslatePathAsync` once for `RemotePath` and once for `ContentPath`. `TranslatePathAsync:126` calls `GetPathMappingByClientAsync`, which is `remotePathMappingRepository.GetByClientIdAsync` at `RemotePathMappingService.cs:42`. There is no caching between calls, so each one is a query.

`DownloadClientQueuePoller` then runs that inside its own `Task.WhenAll` over every enabled client, so it is nested. There is a second identical fan-out at `DownloadClientGateway.cs:168`.

## On #783

That issue reports this exception class with a trace through `DownloadClientQueuePoller.FetchAsync`. I checked that class at `4555ad21`, the commit it reports against: its constructor is the same as today's and it has never held a repository or a context. The trace surfaces there because that is where the await unwinds. The query is further down.

I have not reproduced #783's original scenario, so I am not claiming this closes it. But it is a mechanism that produces that trace, at that line, from a class that owns no state of its own.

## Suggested fix

Split the lookup from the translation. A caller translating many paths for one client resolves the mappings once and translates from them.

[...]

Issue #783, opened 2026-07-28
DbContext concurrency exception in DownloadClientQueuePoller causes "Using cached queue data" and downstream "Import Blocked" failures


f1fson:
## Title
DbContext concurrency exception in DownloadClientQueuePoller causes "Using cached queue data" and downstream "Import Blocked" failures

## Summary
On v1.2.2-canary (commit 4555ad2, the #599 queue-polling refactor), the queue poller for a download client (observed with both SABnzbd and qBittorrent configured simultaneously) repeatedly throws an EF Core concurrency exception:

```
System.InvalidOperationException: A second operation was started on this context instance before a previous operation completed. This is usually caused by different threads concurrently using the same instance of DbContext.
   at Listenarr.Application.Downloads.Queue.DownloadClientQueuePoller.FetchAsync(DownloadClientConfiguration client, TimeSpan clientQueueTimeout, TimeSpan staleSnapshotMaxAge) in /home/runner/work/Listenarr/Listenarr/listenarr.application/Downloads/Queue/DownloadClientQueuePoller.cs:line 90
```

This is immediately followed by a secondary cascade error on the same connection:

```
[ERR] [Microsoft.EntityFrameworkCore.Database.Command] Failed executing DbCommand (0ms) [Parameters=[@downloadClientId='?' (Size = 23)], CommandType='Text', CommandTimeout='30']
System.InvalidOperationException: ExecuteReader can only be called when the connection is open.
```

The UI surfaces this as:
> Using cached queue data — SABnzbd using cached data 5s/10s old after a client error

This repeats on every poll cycle (every ~5s), so the client is treated as permanently degraded even though the actual SABnzbd/qBittorrent API calls succeed (visible a few lines earlier in the log: `Retrieved 1 total items from SABnzbd (queue + history)`).

## Suspected root cause
`DownloadClientQueuePoller.FetchAsync` appears to reuse a single injected `DbContext` instance across concurrent polling tasks (one per configured download client, e.g. SABnzbd + qBittorrent polled in parallel). EF Core's `DbContext` is not thread-safe; concurrent use throws `ConcurrentMethodInvocation`. This looks like a regression from the v1.2.2-canary "split queue display and monitor polling contracts" refactor (PR #599), since it introduces separate/parallel polling paths per client.

## Possible downstream impact
[...]

PR #871, opened 2026-08-21
fix(downloads): resolve a client's path mappings once per batch, not per item

m4bard:
## Summary

`DownloadClientGateway.GetQueueAsync` fans out over every queue item, and each item translated its paths by calling `TranslatePathAsync`, which queries the repository for that client's mappings on every call. The service, the repository and the `ListenArrDbContext` behind them are all scoped, so a queue of N items issued up to 2N concurrent queries against a context that permits one at a time. `DownloadClientQueuePoller` runs that inside its own `Task.WhenAll` across every client, so the fan-out is nested.

Full write-up in #870.

### Fixed
- `IRemotePathMappingService` gains a `TranslatePath` overload taking mappings the caller has already resolved. It does no I/O. `TranslatePathAsync` keeps its behaviour by fetching and delegating to it, so existing callers are unaffected.
- `DownloadClientGateway` resolves a client's mappings once before each of its two item fan-outs and passes them down.
- The single-item import path resolves once as well, where it is one query either way.

Fewer queries as well as non-overlapping ones: a queue resolves the client's mappings once per batch rather than once per translated path. **Corrected 2026-08-24:** this line previously said a thirty-item queue goes from up to sixty queries to one. That held only for items carrying no source files, because the first commit left the `SourceFiles` loop querying per file. The follow-up commit fixes that; see the comment below.

## On #783

That issue reports this exception class with a trace through `DownloadClientQueuePoller.FetchAsync`. That class holds no repository and no context, and did not at `4555ad21` either, which is the commit it reports against. The trace surfaces there because that is where the await unwinds; the query is further down this chain.

I have not reproduced #783's original scenario and this does not claim to close it.

## Testing

`DownloadClientGatewayPathMappingConcurrencyTests` counts the overlap directly rather than racing the EF exception, which would make for a flaky test and a weak claim. A stub records the highest number of lookups in flight, with a small delay so the overlap is observable rather than a scheduling accident.

Ten items carrying two translatable paths each report **10 concurrent lookups without the change and 1 with it**. The lookup count is also pinned at 1, so a change that restores per-item querying fails even if it somehow avoids overlapping.

Full suite: 3,031 passed, 0 failed, 125 skipped, against a 3,029 baseline on `03958c15`.

## Note

This is the second instance of the same pattern, after #862 and its PR #863, which fixes a smaller one in search-result scoring. They are separate files and independently revertable, so they are separate PRs. I originally reported that scoring site as the only instance; that was wrong and is corrected on that issue.

### DIFF
--- a/listenarr.application/Downloads/Common/DownloadClientGateway.cs
+++ b/listenarr.application/Downloads/Common/DownloadClientGateway.cs
@@ -93,7 +93,11 @@ namespace Listenarr.Application.Downloads.Common
         {
             var adapter = ResolveAdapter(client);
             var items = await adapter.GetQueueAsync(client, ct);
-            var tasks = items.Select(item => TranslateQueueItemPathsAsync(client, item));
+            // Resolved once for the batch. Translating each item used to query for the client's
+            // mappings itself, inside this fan-out, against a scoped repository shared by everything
+            // else in the scope.
+            var mappings = await remotePathMappingService.GetPathMappingByClientAsync(client);
+            var tasks = items.Select(item => TranslateQueueItemPathsAsync(mappings, client, item));
             return [.. await Task.WhenAll(tasks)];
         }
 
@@ -137,7 +141,9 @@ namespace Listenarr.Application.Downloads.Common
             var adapter = ResolveAdapter(client);
             var item = await adapter.GetImportItemAsync(client, download, queueItem, null, ct);
 
-            return await TranslateQueueItemPathsAsync(client, item);
+            // Single item, so the lookup here is one query either way.
+            var mappings = await remotePathMappingService.GetPathMappingByClientAsync(client);
+            return await TranslateQueueItemPathsAsync(mappings, client, item);
         }
 
         public async Task<List<Download>> FetchDownloadsAsync(DownloadClientConfiguration client, List<Download> downloads, CancellationToken ct = default)
@@ -165,7 +171,8 @@ namespace Listenarr.Application.Downloads.Common
                     ex);
             }
 
-            var tasks = items.Select(item => TranslateQueueItemPathsAsync(client, item));
+            var mappings = await remotePathMappingService.GetPathMappingByClientAsync(client);
+            var tasks = items.Select(item => TranslateQueueItemPathsAsync(mappings, client, item));
             items = [.. await Task.WhenAll(tasks)];
 
             foreach (QueueItem item in items)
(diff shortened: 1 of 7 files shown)


m4bard:
Pushed a follow-up commit. The batch lookup in the first commit does not cover the whole fan-out, and I would rather say that here than have it reviewed as though it did.

`TranslateQueueItemPathsAsync` translates three things: `RemotePath`, `ContentPath`, and every entry in `SourceFiles`. The first commit moved the first two onto the pre-resolved mappings and left the `SourceFiles` loop calling `TranslatePathAsync`, which still queries per file. That loop is sequential within one item, but it runs inside `GetQueueAsync`'s `Task.WhenAll` over items, so the overlap I was trying to remove is still there whenever items carry source files.

On a torrent client they usually do. `QbittorrentResponseMapper.MapQueueItem` sets `SourceFiles` from the torrent's file list, and Transmission's resolver does the same, so most queue items from those clients arrive with the list already populated.

My own test missed it because the items it builds carry no source files. The new one builds ten items with three files each. It reports 10 concurrent lookups without the follow-up commit and 1 with it, and I ran it with the production line reverted to confirm it fails rather than assuming it would.

One test change came with it. `DownloadClientGatewayTests`'s mapping mock only stubbed `TranslatePathAsync`. Once the source-file loop moved onto `TranslatePath` the mock returned null for every path, the two case-only filenames deduped to one, and the case-sensitive row of `GetQueueItemAsync_DedupesCaseOnlySourceFilesUsingResolvedSemantics` failed. Stubbing both restores it. The same staleness was already there after the first commit, for `ContentPath`. It just did not happen to move that assertion.

Having missed one inside a method I had already read, I swept the rest rather than leave it at that. #870 says I had not checked the other gateway methods, and I can close that now.

Three places resolve mappings, and all three are right: `GetQueueAsync` at :99 and `FetchDownloadsAsync` at :174 each resolve once and then fan out, and `GetQueueItemAsync` at :145 resolves once for a single item. Both fan-outs go through `TranslateQueueItemPathsAsync`, so the one line above is the whole of it rather than the first of several.

Outside the gateway there are only two callers of either method, both in `RemotePathMappingsController`, both serving one request with no fan-out.

Full suite: 3,031 passed, 0 failed, 125 skipped.

The line in the description saying a thirty-item queue goes from up to sixty queries to one held only for items with no source files. I have corrected it to say the lookups per batch, which is the claim the tests actually support.

PR #920, opened 2026-09-01
fix(quality-profiles): stop hiding Maximum Age, and load a profile's qualities

### LINKED ISSUE #895: Unticking Prefer newer releases hides the Maximum Age input while the stored value keeps rejecting releases
Checked against canary `d25b3e11` (1.3.3). Every line number below is from that commit.

## Summary

In a quality profile, unticking "Prefer newer releases" hides the Maximum Age (Days) input. It does not clear the stored value, and nothing in the scorer consults the checkbox, so a profile with Maximum Age set to 30 keeps rejecting releases with "Too old (N days > profile maximum age 30 days)" after the input that sets it has disappeared from the form. To get the input back and change the number, you have to tick a checkbox whose description says it gives bonus points to recent releases, which is not what you want if you were trying to stop age from mattering.

The checkbox was wired once. It gated a real recency bonus that was removed in a scoring rewrite in October 2025, and its description still describes that removed code. So this is a leftover from a rewrite rather than a control that never meant anything.

I have not reproduced this against a running instance. The chain is short and all of it is visible in two files.

## The form

`fe/src/components/settings/QualityProfileFormModal.vue:471-490`:

```html
<CheckboxCard
  v-model="formData.preferNewerReleases"
  title="Prefer newer releases"
  description="Give bonus points to more recent releases (torrent upload date)"
/>

<FormRow
  v-if="formData.preferNewerReleases"
  label="Maximum Age (Days)"
  labelFor="maximumAge"
  help="Reject releases older than this many days (0 = no limit)"
>
  <input id="maximumAge" v-model.number="formData.maximumAge" type="number" min="0" />
</FormRow>
```

The `v-if` is the whole mechanism. `maximumAge` stays on `formData` when the input is not rendered, and the save handler emits the entire object (`:1215`, `emit('save', formData.value)`), so the stored number goes back to the server unchanged.

Two defaults are worth putting next to each other. The form initialises `preferNewerReleases: false` at `:803`, and resets to `false` at `:879`. The domain initialises it to `true` at `listenarr.domain/Audiobooks/QualityProfile.cs:100`. `QualityProfileController.Create` binds `[FromBody] QualityProfile` directly (`listenarr.api/Features/Library/QualityProfileController.cs:107`), so a profile created through the form is stored `false` and a profile created from JSON that omits the key is stored `true`. When you open an existing profile the form takes whatever was stored (`:842-843`).

## The scorer

`listenarr.application/Search/Scoring/SearchResultScorer.cs` never mentions `PreferNewerReleases`. It applies `profile.MaximumAge` in both branches of the NZB check, at `:153-157` and `:171-175`, with no condition other than the value being greater than zero:

```csharp
if (profile.MaximumAge > 0 && ageDays > profile.MaximumAge)
{
    score.RejectionReasons.Add($"Too old ({(int)ageDays} days > profile maximum age {profile.MaximumAge} days)");
    score.TotalScore = -1;
    return score;
}
```

It also applies an unconditional age penalty at `:334-344`, up to minus 60 over ten years. There is no recency bonus anywhere in the file.

Across the whole backend, `PreferNewerReleases` appears twice: the declaration at `QualityProfile.cs:100` and a field copy in `QualityProfileRepository.cs:74`. Neither is a read that changes behaviour.

So the description is wrong twice over. There are no bonus points for recent releases, and there is a penalty for old ones that applies whether the box is ticked or not.

## Where the gate went

`491d9d03`, "Add QualityProfileService and refactor polling methods", 2025-10-07, added the read:

```csharp
// Score age (newer is better if preference is set)
if (profile.PreferNewerReleases && searchResult.PublishedDate != default(DateTime))
{
    var age = (DateTime.UtcNow - searchResult.PublishedDate).TotalDays;
    var ageScore = Math.Max(0, 30 - (int)(age / 7)); // Lose 1 point per week
    if (ageScore > 0)
    {
        score.TotalScore += ageScore;
        score.ScoreBreakdown["Age"] = ageScore;
    }
}
```

That is exactly what the checkbox description promises. `bd20454d`, "Add quality profile selection to library addition", 2025-10-08, removed that block along with the seeder scoring next to it as part of rewriting the scoring method. Nothing replaced the gate, and the unconditional penalty at `:334-344` is what the age handling turned into.

The frontend text was accurate when it was written and has outlived the code it described.

## The part I think matters most

The dead toggle on its own would be cosmetic. The interaction with `v-if` is not, because it takes a working setting away from the user while leaving it in force.

The sequence is ordinary. Set Maximum Age to 30. Later decide age should not influence results and untick Prefer newer releases, since that is the only age-related control on the page. The Maximum Age input disappears, which reads like confirmation that it no longer applies. Searches keep rejecting releases with "Too old (N days > profile maximum age 30 days)", and the profile editor no longer contains the number 30 anywhere.

One thing that softens this, and I want to be accurate about it: the value is not completely invisible. The profile card on the Quality Profiles list shows it under Limits and Requirements at `fe/src/views/settings/QualityProfilesTab.vue:177-181`, and that block is gated only on `profile.maximumAge > 0`, not on the checkbox. So a user who goes back to the list can see "Max Age: 30 days". What they cannot do is change it without first ticking a box that says it does something else. The rejection message names the profile maximum age, so somebody reading logs closely has a thread to pull.

## Control

The other numeric limits in the same profile section behave the way you would expect. `MinimumSize` and `MaximumSize` are read at `SearchResultScorer.cs:95-106`, `MinimumSeeders` at `:110`, and none of their inputs are gated behind a checkbox. `Indexer.Retention` is read a few lines above the `MaximumAge` check, at `:128`. So the scorer does consult profile and indexer configuration generally, which is what makes this one property's absence look like a gap rather than a search that missed something.

## Reproducing

Run at the repo root on `d25b3e11`:

```sh
# every reference to the toggle: declaration, one field copy, and the frontend
git grep -in "preferNewerReleases" -- ':!fe/eslint-report.json' ':!**/Migrations/**'

# the input it hides, and the code that keeps using the value
sed -n '471,490p' fe/src/components/settings/QualityProfileFormModal.vue
sed -n '150,176p' listenarr.application/Search/Scoring/SearchResultScorer.cs

# the gate that used to exist, and the commit that removed it
git log -S "PreferNewerReleases" --oneline d25b3e11 -- \
  'listenarr.domain' 'listenarr.application' 'listenarr.infrastructure' 'listenarr.api' ':!**/Migrations/**'
git show 491d9d03 | grep -n -A10 "PreferNewerReleases && searchResult.PublishedDate"
git show bd20454d | grep -n -B4 -A12 "PreferNewerReleases && searchResult.PublishedDate"
```

Exclude `fe/eslint-report.json`; it is checked in and repeats every frontend symbol. Use `git grep` rather than `grep -rn`, which on a built tree also walks `bin/` and `obj/`. `git log -S` needs `Migrations` excluded or the EF model snapshot buries the result.

## Options

The visibility rule and the dead toggle can be fixed separately, and the first one is small.

Dropping the `v-if` at `:479` shows Maximum Age unconditionally, which matches what the scorer does today and means the control that has an effect is always reachable. On its own that is a one-line change and leaves the toggle inert but harmless. I am happy to send it if you want it that way, though I would rather not, because it decides half of the second question by default.

The second question is what the checkbox is for. Restoring something like the removed bonus would make the description true again, but the scoring model has changed a lot since October and I would be guessing at how a bonus should interact with the current penalty. Making it gate the age handling that exists now, so that unticking it disables both the `MaximumAge` rejection and the age penalty, is closer to what a user would infer from the form but is a different behaviour from what it originally did. Removing the checkbox and always applying age handling is the third answer, and given that `MaximumAge` already works and is the control people actually reach for, it may be the honest one.

That is a product call rather than a bug fix, so I have not written any of them.

m4bard:
Fixes #895, and fixes a second problem in the same file that I found while writing the test for the first.

## Maximum Age was hidden while it was still rejecting

`Maximum Age (Days)` rendered behind `v-if="formData.preferNewerReleases"`. It is not part of that checkbox. `SearchResultScorer` applies `profile.MaximumAge` as a hard reject whenever it is above zero, at `:153` and `:171`, and never reads `PreferNewerReleases` at all.

So unticking the box hid an input while the filter it sets stayed on, and the only route back to the number was to tick a box whose description offers a bonus for recent releases, which is the opposite of what someone turning off an age limit wants.

The input now always renders.

## Opening a profile lost its qualities

The watch that copies the profile into the form runs with `{ immediate: true }`, so it calls `initializeQualitiesFromProfile` during setup, and that was a `const` arrow declared below it. Every mount carrying a profile threw `ReferenceError` there. Vue catches it, so the only symptom was an empty quality list, and saving was then refused by the "select at least one quality" guard until the whole ladder was picked again by hand.

`parseQualityString` had the same shape one level down and took over as soon as the first was fixed. Both are function declarations now, so they hoist.

I have written that up as its own issue, since it is a bigger problem than #895 and deserves to be findable on its own. It is in this PR because the two are in the same file and I could not test #895 without it. Say the word if you would rather review them separately and I will split it.

## Left alone deliberately

`PreferNewerReleases` still has no reader anywhere in the backend. Its description promises a recency bonus that no longer exists. Removing the checkbox and implementing the bonus are both reasonable and the choice is yours, so I have not guessed. Say which and I will follow up.

Also unchanged, and worth knowing: the form initialises `preferNewerReleases` to `false` while `QualityProfile.cs:100` initialises it to `true`, so a profile made in the UI and one made from JSON that omits the key disagree. Nothing reads the field, so it has no effect today, and I did not want to bury a default change in this.

## Tests

Four, in a new spec. There was no test that mounted this modal at all, which is how a component that threw on every mount with a profile went unnoticed.

Re-hiding Maximum Age fails one. Putting either helper back to a `const` arrow fails all four.

---

Worked through with Claude Code at my direction. The claims above were checked by running them rather than by reading, and I reviewed this before posting.


m4bard:
The quality-initialisation half of this is now filed on its own as #923, since it is the more serious of the two and worth being findable without reading a PR about a hidden input.

PR #985, opened 2026-09-14
fix(search): compute the profile size gates in long, not int

### LINKED ISSUE #977: A quality profile size limit of 2 GB or more rejects every release
## What happens
Set Maximum Size on a quality profile to 2048 MB or higher and every search result is rejected with "File too large", whatever its actual size. Set Minimum Size to 2048 MB or higher and the minimum silently stops applying: nothing is ever rejected as too small.
## Why
`listenarr.application/Search/Scoring/SearchResultScorer.cs`, in the size block that runs for torrents:
```csharp
if (profile.MinimumSize > 0 && searchResult.Size < profile.MinimumSize * 1024 * 1024)
...
if (profile.MaximumSize > 0 && searchResult.Size > profile.MaximumSize * 1024 * 1024)
```
`QualityProfile.MinimumSize` and `QualityProfile.MaximumSize` are both `int` megabytes, so `profile.MaximumSize * 1024 * 1024` is evaluated in `int`. 2048 MB is one past `int.MaxValue` once multiplied and wraps to -2147483648. `searchResult.Size` is a `long`, so the comparison promotes the wrapped negative value rather than the intended one, and the two gates fail in opposite directions:
- Maximum: every size is greater than a negative number, so everything is rejected.
- Minimum: no size is smaller than a negative number, so nothing is rejected and the setting does nothing.
The indexer size gate a few lines above already writes `(long)indexerMaximumSizeMb * 1024 * 1024`, which is what makes the two neighbouring lines look deliberate.
There is no upper bound on the input. `fe/src/components/settings/QualityProfileFormModal.vue` renders both as `type="number"` with `min="0"` and no `max`, and I did not find server-side validation of either. A 2 GB ceiling on an audiobook profile does not seem exotic.
## Reproduction
Either through the UI:
1. Settings > Quality Profiles, edit a profile.
2. Set Maximum Size to 2048 and save.
3. Run a search that previously returned results, with a torrent indexer configured.
4. Every result is rejected as too large.
Or as a test, which is what I would rather hand you:
```csharp
var profile = new QualityProfile { MinimumSeeders = 0, MaximumAge = 3650, MaximumSize = 2048 };
var result = new SearchResult
{
    Title = "Ordinary Torrent",
    DownloadType = "torrent",
    Size = 1536L * 1024 * 1024,
    Seeders = 10,
    PublishedDate = DateTime.UtcNow.AddDays(-1).ToString("o")
};
var score = await service.ScoreSearchResult(result, profile);
// Fails: the 1.5 GB result is rejected against a 2 GB ceiling.
Assert.DoesNotContain(score.RejectionReasons, r => r.Contains("too large ("));
```
## Fix
Cast before the multiply, in both lines, the way the indexer gate already does.
I have a branch ready with the two casts and two theories covering it, one per gate, each carrying a control row below the overflow so a fix that simply stopped rejecting, or simply started rejecting, would fail. Without the casts the four overflow rows fail and the two control rows pass. The full backend suite is green with them.
I found this while wiring `Indexer.MaximumSize` in #921 and kept it out of that PR deliberately: it is a different bug, and that file already has a rebase ahead of it against #863.
---

m4bard:
Fixes #977.

`SearchResultScorer`'s size gate multiplies `QualityProfile.MinimumSize` and `MaximumSize`, both `int` megabytes, by `1024 * 1024` before comparing against a result's `long` byte size. At 2048 MB the multiply overflows `int` and wraps to -2147483648, so the two gates fail in opposite directions: Maximum rejects every size, since everything is greater than a negative number, and Minimum stops rejecting anything, since nothing is smaller than one. The indexer size gate a few lines above already casts to `long` first, which is what makes the two profile lines next to it look like an oversight rather than a design choice.

The fix is the same cast the indexer gate already uses, applied to both lines.

Two theories, each with a control row below the overflow threshold so a change that simply stopped rejecting, or simply started rejecting, would fail the control. Without the casts, the four overflow rows fail and the two controls pass.

Suite run against current canary (`upstream/canary` at `a630572e9`, which this branch is built on): targeted filter `QualityProfileScoringTests`, 23 passed, 0 failed.

Found while wiring `Indexer.MaximumSize` in #921 and kept out of that PR as a separate bug. Full mechanism and the repro test are in the issue.

Disclosure: drafted with Claude Code at my direction; I read the cited code at commit a630572e9 and reviewed this before posting.


m4bard:
Closing in favor of #921. This fix is not abandoned; it moved. The `(long)` cast for the profile size gates is now commit 201220829 on #921's branch, the same PR where this bug was found in the first place.

